PR #6, opened 2025-05-15
Add how-to guide for multi-cloud

alaye-ms:


api/preview/documentdb_types.go
```diff
@@ -40,11 +40,13 @@ type Resource struct {
...
 }
 
-type ClusterReplication struct {
-	// Primary is the name of the primary cluster for replication.
-	Primary string `json:"primary"`
-	// ClusterList is the list of clusters participating in replication.
-	ClusterList []string `json:"clusterList"`
+type PhysicalReplication struct {
+  // FleetEnabled determines whether to use KubeFleet mechanics for the replication
+	FleetEnabled bool     `json:"fleetEnabled,omitempty"`
```
hossain-rayhan, line 45: Should we say `enableFleetForCrossCloud`? as we dont need for replication in same cloud environment.

api/preview/documentdb_types.go
```diff
@@ -40,11 +40,13 @@ type Resource struct {
...
+  // Primary is the name of the primary cluster for replication.
+	Primary      string   `json:"primary"`
+  // ClusterList is the list of clusters participating in replication.
+	ClusterList  []string `json:"clusterList"`
```
hossain-rayhan, line 49: Are they name of the DocumentDB app in the deployment manifest? Wondering how to make it easier if the names need to be same.

docs/multi-cloud-deployment-guide.md
```diff
@@ -0,0 +1,489 @@
...
+This multi-cloud deployment uses KubeFleet to manage DocumentDB instances across different cloud providers:
+
+- **Hub Cluster**: Central control plane for managing all member clusters
+- **Member Clusters**: Clusters in different cloud environments (Azure, On-prem)
+- **DocumentDB Operator**: Custom operator for DocumentDB deployments
+- **Fleet Networking**: Enables communication between clusters
+
+## Setting Up the Hub Cluster
+
+The hub cluster serves as the central controller for managing the member clusters, find setup instructions here:
```
hossain-rayhan, line 42: should we add some commands alongside mentioning the link?

docs/multi-cloud-deployment-guide.md
```diff
@@ -0,0 +1,489 @@
...
+https://learn.microsoft.com/en-us/azure/kubernetes-fleet/overview
+
+## Adding Clusters to Fleet
+
+### Adding AKS cluster to the fleet
+
+Adding an AKS cluster to the fleet is very simple with the Azure portal: 
+https://learn.microsoft.com/en-us/azure/kubernetes-fleet/quickstart-create-fleet-and-members-portal
```
hossain-rayhan, line 50: Same. Should we render some commands here? If it's a lot, we can keep the link only.

docs/multi-cloud-deployment-guide.md
```diff
@@ -0,0 +1,489 @@
...
+git clone https://github.com/kubefleet-dev/kubefleet.git
+cd kubefleet
+./hack/membership/joinMC.sh v0.14.8 hub $clusterName
+cd ..
+
+# Wait until the cluster shows the correct number of nodes, usually about a minute
+
+git clone https://github.com/Azure/fleet-networking
+cd fleet-networking
+./hack/membership/joinMC.sh v0.14.8 v0.3.8 hub $clusterName
```
hossain-rayhan, line 73: why two version `v0.14.8 v0.3.8`? Also, is hardcoded good for now?

docs/multi-cloud-deployment-guide.md
```diff
@@ -0,0 +1,489 @@
...
+konprem apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.17.1/cert-manager.yaml
+```
+
+3. Install sidecar injector on the hub:
+```bash
+cat <<EOF > sidecar-injector.yaml
+apiVersion: v1
+kind: Service
+metadata:
+  annotations:
```
hossain-rayhan, line 108: Replace with our latest setup guide from README.

docs/multi-cloud-deployment-guide.md
```diff
@@ -0,0 +1,489 @@
...
+kubectl apply -f sidecar-injector.yaml
+```
+
+4. Install the DocumentDB operator on the hub:
+
+```bash
+VERSION=36  # Replace with your current version
+helm package documentdb-chart --version 0.0.${VERSION}
+helm install documentdb-operator ./documentdb-operator-0.0.${VERSION}.tgz --namespace documentdb-operator-ns 
+```
```
hossain-rayhan, line 180: Don't need this separately anymore as mentioned above. Replace with setup section from README.

docs/multi-cloud-deployment-guide.md
```diff
@@ -0,0 +1,489 @@
...
+
+5. Deploy certificates on each cluster:
```
hossain-rayhan, line 182: Maybe we need to mention how many certificates we need and where they are used. And then mention we are using self-signed certs here?

docs/multi-cloud-deployment-guide.md
```diff
@@ -0,0 +1,489 @@
...
+apiVersion: placement.kubernetes-fleet.io/v1beta1
+kind: ClusterResourcePlacement
+metadata:
+  name: documentdb-base
+spec:
+  resourceSelectors:
+    - group: ""
+      version: v1
+      kind: Namespace
+      name: documentdb-operator-ns
```
hossain-rayhan, line 257: Our operator by default gets installed in `documentdb-operator` namespace. Is there way to configure it. Or use the default one?

docs/multi-cloud-deployment-guide.md
```diff
@@ -0,0 +1,489 @@
...
+
+## Setting Up Replication
+
+Physical replication provides high availability and disaster recovery capabilities across clusters.
+
+1. Create configuration maps to identify clusters:
+
+```bash
+konprem create configmap cluster-name -n kube-system --from-literal=name=replica-cluster
+kazure create configmap cluster-name -n kube-system --from-literal=name=primary-cluster
```
hossain-rayhan, line 360: `konprem` and `kazure` are ours alias?

docs/multi-cloud-deployment-guide.md
```diff
@@ -0,0 +1,489 @@
...
+
+apiVersion: db.microsoft.com/preview
+kind: DocumentDB
+metadata:
+  name: documentdb-preview
+  namespace: documentdb-preview-ns
+spec:
+  nodeCount: 1
+  instancesPerNode: 1
+  documentDBImage: <IMAGE LOCATION>
```
hossain-rayhan, line 410: Update with our default image from GitHub.

docs/multi-cloud-deployment-guide.md
```diff
@@ -0,0 +1,489 @@
...
+spec:
+  resourceSelectors:
+    - group: ""
+      version: v1
+      kind: Namespace
+      name: documentdb-preview-ns
+  policy:
+    placementType: PickAll
+  strategy:
+    type: RollingUpdate
```
hossain-rayhan, line 435: Why this ClusterResourcePlacement short while the first one is massive?

docs/multi-cloud-deployment-guide.md
```diff
@@ -0,0 +1,489 @@
...
+
+4. To perform a failover:
+
+```bash
+kubectl patch documentdb documentdb-preview -n documentdb-preview-ns \
+  --type='json' -p='[
+  {"op": "replace", "path": "/spec/physicalReplication/primary", "value":"replica-cluster"},
+  {"op": "replace", "path": "/spec/physicalReplication/clusterList", "value":["replica-cluster"]}
+  ]'
+```
```
hossain-rayhan, line 450: where do we run this?

docs/multi-cloud-deployment-guide.md
```diff
@@ -0,0 +1,489 @@
...
+
+## Testing and Verification
+
+1. Test connection to DocumentDB:
+
+```bash
+# Get the service IP
```
hossain-rayhan, line 457: On which cluster? Mention `primary` or azure?

internal/controller/physical_replication.go
```diff
@@ -92,15 +97,20 @@ func (r *DocumentDBReconciler) AddPhysicalReplicationToClusterSpec(
...
+		clusterNameConfigMap := &corev1.ConfigMap{}
+		err := r.Get(ctx, types.NamespacedName{Name: clusterMapName, Namespace: "kube-system"}, clusterNameConfigMap)
+		if err != nil {
+			return "", "", err
+		}
+
+		self = clusterNameConfigMap.Data["name"]
+		if self == "" {
+			return "", "", fmt.Errorf("name key not found in kube-system:cluster-name configmap")
+		}
```
hossain-rayhan, line 113: I think we also need to test and update guidance for replication without fleet like both clusters on Azure. Then this is the advanced case. 
This we can ensure our normal replication works as well without many complexities.

api/preview/documentdb_types.go
alaye-ms: In the multi-cloud scenario, they need to be the names set in their configmap, in the single-cloud scenario it's the documentdb CRD names

docs/multi-cloud-deployment-guide.md
alaye-ms: Like the `az` commands? I don't think that's needed if there's already a guide

docs/multi-cloud-deployment-guide.md
alaye-ms: One is the KubeFleet version, one is the Fleet networking version. They're separate projects. I think hardcoded is okay for a guide.

docs/multi-cloud-deployment-guide.md
```diff
@@ -0,0 +1,489 @@
...
+
+```bash
+cat <<EOF > primary-name.yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: cluster-name
+  namespace: kube-system
+data:
+  name: "primary-cluster"
```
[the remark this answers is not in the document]
alaye-ms: I don't think that's likely. Any other information would be a part of the documentdb object, not needed on a per-cluster level.

internal/controller/physical_replication.go
alaye-ms: That's reasonable.

docs/multi-cloud-deployment-guide.md
hossain-rayhan: I mean which kubectl profile? The Azure one or the OnPrem one? Should we mention that?

alaye-ms: Oh it uses the default, which is the hub, I can make that clear if it isn't obvious. Any time I don't use a profile in this doc it's a hub request.

docs/multi-cloud-deployment-guide.md
```diff
@@ -0,0 +1,402 @@
...
+9. [Failover Procedures](#failover-procedures)
+
+## Prerequisites
+
+- Azure account
+- Azure CLI installed and configured with appropriate permissions
+- kubectl installed
+- helm installed
+- Git client
+- Two kubernetes clusters that are network connected to each other
```
xgerman, line 28: say to use express route or https://learn.microsoft.com/en-us/azure/vpn-gateway/vpn-gateway-about-vpngateways or similar

xgerman (commented):
Can we add after each step a few commands to verify that all worked, e.g. after you add acluster check that it shows up in kube-fleet?

PR #17, opened 2025-05-16
Create codeql.yml

xgerman:


xgerman:
This is no good: see https://github.com/documentdb/documentdb/blob/main/.github/workflows/codeql.yml for the documentdb project. We need to add the go stuff for a reasonable build @guanzhousongmicrosoft wonder if you could help

PR #29, opened 2025-05-21
Default CNPG Postgres StopDelay to 30s instead of 1800s and make configurable

hossain-rayhan:
Pods are stuck in TERMINATING state for 1800 seconds. CNPG defaults pods to 1800s `stopDelay`. For DocumentDB we are defaulting to 30s and making it configurable through DocumentDB spec.

From CNPG:
```
When a Pod running Postgres is deleted, either manually or by Kubernetes
following a node drain operation, the kubelet will send a termination signal to the
instance manager, and the instance manager will take care of shutting down
PostgreSQL in an appropriate way.
The `.spec.smartShutdownTimeout` and `.spec.stopDelay` options, expressed in seconds,
control the amount of time given to PostgreSQL to shut down. The values default
to 180 and 1800 seconds, respectively.
```

Now customer can change it in the spec like:
```yaml
apiVersion: db.microsoft.com/preview
kind: DocumentDB
metadata:
  name: documentdb-preview
  namespace: documentdb-preview-ns
spec:
  nodeCount: 1
  instancesPerNode: 1
  documentDBImage: pgcosmoscontroller.azurecr.io/documentdb-pg:16
  resource:
    pvcSize: 10Gi
  publicLoadBalancer:
    enabled: true
  timeouts:
    stopDelay: 60
```


internal/cnpg/cnpg_cluster.go
```diff
@@ -23,36 +23,40 @@ func GetCnpgClusterSpec(req ctrl.Request, documentdb dbpreview.DocumentDB, docum
...
+				PostgresUID: 105,
+				PostgresGID: 108,
+				PostgresConfiguration: cnpgv1.PostgresConfiguration{
+					AdditionalLibraries: []string{"pg_cron", "pg_documentdb_core", "pg_documentdb"},
+					Parameters: map[string]string{
+						"cron.database_name":    "postgres",
+						"max_replication_slots": "10",
+						"max_wal_senders":       "10",
+					},
+					PgHBA: []string{
```
xgerman, line 49: we will need to scope that down in a later commit

xgerman (approved):

PR #39, opened 2025-06-19
Publish DocumentDB CoonectionString in Status fields

hossain-rayhan:
If public load balancer is enabled, use the load balancer ip to generate the DocumentDB connection string and publish it in the status fields.

This is the start to make things easier when load balancer is enabled. For without loadbalancer, we need extra thinking which will come in separate PRs.

**Test:**
- Verified by deploying in AKS cluster with Azure public load balancer.

### DIFF
--- a/internal/controller/documentdb_controller.go
+++ b/internal/controller/documentdb_controller.go
@@ -62,6 +62,7 @@ func (r *DocumentDBReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 		return ctrl.Result{}, err
 	}
 
+	var loadBalancerIp string
 	// Only create/manage the public LoadBalancer if enabled in the CRD spec
 	if documentdb.Spec.PublicLoadBalancer.Enabled {
 		// Define the Public LoadBalancer Service for this DocumentDB instance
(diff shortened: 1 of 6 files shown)


internal/utils/util.go
```diff
@@ -231,3 +234,19 @@ func DeleteRoleBinding(ctx context.Context, c client.Client, name, namespace str
 	}
 	return nil
 }
+
+// GenerateConnectionString returns a MongoDB connection string for the DocumentDB instance
+func GenerateConnectionString(documentdb *dbpreview.DocumentDB, loadBalancerIP string) string {
+	if loadBalancerIP == "" {
+		return ""
```
xgerman, line 241: Is this an error condiction? People will ask why the conenction string is ""

internal/controller/documentdb_controller.go
```diff
@@ -75,7 +76,8 @@ func (r *DocumentDBReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		}
 
 		// Ensure LoadBalancer has an IP assigned
-		if err := util.EnsureLoadBalancerIP(ctx, foundService); err != nil {
+		loadBalancerIp, err = util.EnsureLoadBalancerIP(ctx, foundService)
+		if err != nil {
 			log.Info("DocumentDB LoadBalancer External IP not assigned, Requeuing.")
```
xgerman, line 81: does this only work for external IP? Ehat about internal?

hossain-rayhan: Now supporting both public loadbalancer and Cluster IP service.

xgerman (approved):

xgerman (approved):

xgerman (approved):

PR #61, opened 2025-07-28
TESTS - Refactor Tests to Use Local Build, Add Multi-Arch Build and Tests

guanzhousongmicrosoft:


guanzhousongmicrosoft:
Use local build, the Pull Requests can build local image, use it to deploy k8s clusters, and use python or mongosh to test connectivity.

PR #129, opened 2025-10-23
Adds a dev container setup

xgerman:
- the dev container will install automaticlaly the documentdb extension
- each component has their own makefile
- scripts support arbitrary image registries
- it has a scrit to start a kind cluster with registry
- a new setup script will create a kind  clsuter, deploy/install document db operator, and deploy a documentdb cluster for convenient testing

### DIFF
--- a/scripts/development/deploy.sh
+++ b/scripts/development/deploy.sh
@@ -11,6 +11,7 @@ TAG=${TAG:-"0.1.1"}
 SKIP_KIND_SETUP=${SKIP_KIND_SETUP:-"false"}
 SKIP_CERT_MANAGER=${SKIP_CERT_MANAGER:-"false"}
 DEPLOY_CLUSTER=${DEPLOY_CLUSTER:-"false"}
+DEPLOY=${DEPLOY:-"false"}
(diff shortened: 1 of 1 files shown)


scripts/development/deploy.sh
```diff
@@ -0,0 +1,470 @@
...
+        echo "Skipping Kind cluster setup (SKIP_KIND_SETUP=true)"
+        check_registry
+    fi
+    
+    # Build and push both images
+    build_push_operator
+    build_push_plugin
+    
+    # Optionally deploy to cluster
+    if [ "${DEPLOY:-false}" = "true" ]; then
```
alaye-ms: Can we include this ENV variable at the top as well

scripts/development/deploy.sh
```diff
@@ -0,0 +1,470 @@
...
+        deploy_documentdb_cluster
+        
+        # Deploy mongosh client for easy testing
+        deploy_mongosh_client
+    fi
+    
+    # Show deployment status
+    echo ""
+    echo "Checking deployment status..."
+documentdb-operator    kubectl get pods -n  || echo "No pods found in documentdb-operator namespace yet"
```
alaye-ms: It looks like this is a typo

PR #133, opened 2025-10-28
Allows use of Istio and adds multi-cloud global strong scripts

alaye-ms:


config/crd/bases/db.microsoft.com_documentdbs.yaml
```diff
@@ -56,12 +56,32 @@ spec:
...
+                      - name
+                      type: object
                     type: array
-                  enableFleetForCrossCloud:
-                    description: EnableFleetForCrossCloud determines whether to use
-                      KubeFleet mechanics for the replication
-                    type: boolean
+                  crossCloudNetworkingStrategy:
+                    description: CrossCloudNetworking determines which type of networking
+                      mechanics for the replication
```
xgerman: shouldn't that be an enum?

scripts/multi-cloud-deployment/failover.sh
```diff
@@ -0,0 +1,55 @@
+#/bin/bash
```
xgerman: isn't that just the DNS failover? Maybe rename the script name

operator/src/internal/cnpg/cnpg_cluster.go
```diff
@@ -31,9 +31,9 @@ func GetCnpgClusterSpec(req ctrl.Request, documentdb *dbpreview.DocumentDB, docu
 	}
 
 	// Configure storage class - use specified storage class or nil for default
-	var storageClass *string
-	if documentdb.Spec.Resource.Storage.StorageClass != "" {
```
hossain-rayhan: We are removing this line means for single cluster setup in AWS or GKE it will fail unless we are passing it right. From the controller I see we are only passing `replicationContext.StorageClass` but not from spec.

operator/src/scripts/multi-cloud-deployment/deploy-gke.sh
```diff
@@ -0,0 +1,100 @@
+#!/bin/bash
```
hossain-rayhan: We should keep this script folder clean. Only the scripts which are used/run the the operator itself or we need inside the host machine. 

All other setup scripts for different environment should go under `documentdb-playground/` specific folder.

PR #141, opened 2025-11-04
Multi-Region Backup and Restore

WentingWu666666:


operator/src/internal/controller/backup_controller.go
```diff
@@ -82,7 +83,18 @@ func (r *BackupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	}
 	if err := r.Get(ctx, cnpgBackupKey, cnpgBackup); err != nil {
 		if apierrors.IsNotFound(err) {
-			return r.createCNPGBackup(ctx, backup, cluster.Spec.Backup)
+
+			// Skip backup if the cluster is not primary
```
xgerman: did we test this during a failkover? Probably fine but we can't be backing up while the server is catching up.... or at least think about that case.

alaye-ms (approved):

alaye-ms (approved):

PR #143, opened 2025-11-05
Update and cleanup README

abhirockzz:


hossain-rayhan (changes requested):
I still prefer to keep the Quick Start section separately.

xgerman (approved):

hossain-rayhan (approved):
we can add that later. Lets merge this one first to fix broken link.

PR #203, opened 2025-12-12
Add the 0.1.3 bump to changelog.md

alaye-ms:


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## [0.1.3] - 2025-12-12
+
+### Major Features
+- **Change the CRD API version to match documentdb.io**
```
hossain-rayhan: I feel like we introduced a breaking change. We removed the old CRD API group name and files. Old clusters/pods won't come up once we update the operator. Let's discuss more in meeting.

alaye-ms: This PR is just to put it into the changelog. The change itself has already gone through

PR #237, opened 2026-02-12
Appinsights implementation

Ritvik-Jayaswal:
## Summary

This PR adds Application Insights telemetry integration to the DocumentDB Kubernetes Operator using OpenTelemetry (OTel) SDK with an OTel Collector sidecar architecture.

## Architecture

Operator (OTel Go SDK) → OTLP gRPC → OTel Collector Sidecar → Azure Monitor Exporter → App Insights

The operator emits telemetry via the standard OpenTelemetry Go SDK (`go.opentelemetry.io/otel`). An OTel Collector contrib sidecar receives OTLP data and exports to Application Insights via the Azure Monitor exporter.

### Telemetry Package (`internal/telemetry/`)
- **OTel SDK integration**: Replaced archived `microsoft/ApplicationInsights-Go` with actively maintained OpenTelemetry Go SDK
- **Adapter interfaces**: Decoupled telemetry from reconcilers using `DocumentDBTelemetry`, `BackupTelemetry`, `ScheduledBackupTelemetry` interfaces with noop defaults for tests
- **InstrumentedReconciler**: Decorator pattern for cross-cutting reconciliation duration/error tracking
- **Resource ID via `metadata.uid`**: Replaced UUID annotation persistence (GUIDManager) with Kubernetes `metadata.uid` directly — zero API writes, no race conditions
- **Option 3 Cluster ID**: Cloud-native K8s cluster identification (AKS/EKS/GKE detection + kube-system UID fallback), attached as `kubernetes_cluster_id` common property
- **Buffered metrics**: Collect gauge metrics every 15 min, flush hourly (500 item cap with auto-flush)
- **Cloud provider detection** from node labels/providerID

### Controller Integration
- **DocumentDB Controller**: Track ClusterCreated, ClusterUpdated, ClusterDeleted events
- **Backup Controller**: Track BackupCreated, BackupDeleted, BackupExpired events
- **ScheduledBackup Controller**: Track ScheduledBackupCreated and triggered events
- **Periodic metrics**: operator.health.status, clusters.active.count, cluster.configuration, service_exposure, TLS, plugin usage, backup counts

### Helm Chart Updates
- **OTel Collector sidecar** added to operator deployment (`11_otel_collector_config.yaml`)
- OTLP receiver + Azure Monitor exporter configured via `telemetry` values
- Added `telemetry.collector.image` config for collector image override
- Support for connection string, instrumentation key, or existing secret
- Added `POD_NAMESPACE` and `OTEL_EXPORTER_OTLP_ENDPOINT` environment variables
- Added node read RBAC permissions for cloud provider detection

### Release Workflow
- Added App Insights connection string injection to `release_operator.yml` (bakes secret into published chart)

### Design Documents
- `docs/designs/appinsights-metrics.md` — Telemetry events and metrics specification
- `docs/designs/telemetry-guid-strategy.md` — GUID strategy analysis (UUID vs SHA256 vs metadata.uid vs cloud-native)
- `docs/designs/telemetry-configuration.md` — Configuration guide with opt-out instructions
- `AGENTS.md` — Telemetry testing section for AI agents
- `documentdb-playground/telemetry/sample-appinsights-cluster.yaml` — Sample cluster for E2E testing

## Testing

- All unit tests pass (including telemetry package tests at ~55% coverage)
- E2E tested on AKS (K8s v1.35.0) with OTel Collector sidecar
- Verified events (OperatorStartup, ClusterCreated, ClusterDeleted) flowing to Application Insights via `traces` table
- Confirmed `kubernetes_cluster_id` (Option 3) attached as common property
- Hourly metric flush verified (28 items buffered → transmitted, HTTP 200)
- Instrumentation key validated with direct App Insights API test

## Related

- Spec document: docs/designs/appinsights-metrics.md
- GUID strategy: docs/designs/telemetry-guid-strategy.md
- Configuration: docs/designs/telemetry-configuration.md


WentingWu666666 (commented):
## Suggestion: Decouple telemetry from reconcilers

The tracking logic is currently embedded directly in each reconciler — every controller carries a `TelemetryMgr` field, repeats `if r.TelemetryMgr == nil || !r.TelemetryMgr.IsEnabled()` guards, and mixes data-gathering (GUID lookups, retention calculation, replication context) with the core reconcile flow. This violates Single Responsibility and makes both reconciliation and telemetry harder to test independently.

### Recommended approach

**1. Decorator reconciler for cross-cutting metrics**

Reconciliation duration and error tracking are pure cross-cutting concerns. Wrap each reconciler instead of embedding `defer` / error-tracking calls:

```go
type InstrumentedReconciler struct {
    inner        reconcile.Reconciler
    telemetry    *telemetry.Manager
    resourceType string
}

func (r *InstrumentedReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
    start := time.Now()
    result, err := r.inner.Reconcile(ctx, req)
    r.telemetry.Metrics.TrackReconciliationDuration(telemetry.ReconciliationDurationMetric{
        ResourceType:    r.resourceType,
        Operation:       "reconcile",
        DurationSeconds: time.Since(start).Seconds(),
        Status:          statusStr(err),
    })
    if err != nil {
        r.telemetry.Events.TrackReconciliationError(...)
    }
    return result, err
}
```

This removes ~40 lines of repeated boilerplate across the three controllers.

**2. Telemetry adapter interface for domain events**

For domain-specific events (`ClusterCreated`, `BackupDeleted`, etc.), extract a per-resource adapter behind an interface. The reconciler makes a single call; the adapter owns all data-gathering and nil-safety:

```go
// Interface — easy to mock in tests
type BackupTelemetry interface {
    BackupCreated(ctx context.Context, backup *Backup, cluster *DocumentDB, backupType string)
    BackupDeleted(ctx context.Context, backup *Backup, reason string)
}

// No-op default (eliminates nil checks everywhere)
type noopBackupTelemetry struct{}
func (noopBackupTelemetry) BackupCreated(...) {}
func (noopBackupTelemetry) BackupDeleted(...) {}

// Real implementation lives in the telemetry package
type backupTelemetryImpl struct { mgr *Manager }
func (t *backupTelemetryImpl) BackupCreated(ctx context.Context, backup *Backup, cluster *DocumentDB, backupType string) {
    backupID, err := t.mgr.GUIDs.GetOrCreateBackupID(ctx, backup)
    // ... all the data-gathering that currently lives in backup_controller.go
    t.mgr.Events.TrackBackupCreated(BackupCreatedEvent{...})
}
```

The reconciler becomes:
```go
type BackupReconciler struct {
    client.Client
    Scheme    *runtime.Scheme
    Recorder  record.EventRecorder
    Telemetry BackupTelemetry  // interface, never nil
}

// In createCNPGBackup — one clean line replaces 30 lines:
r.Telemetry.BackupCreated(ctx, backup, cluster, backupType)
```

### Benefits
- **SRP**: Reconcilers focus on reconciliation; telemetry logic lives in the telemetry package
- **Testability**: Tests inject `noopBackupTelemetry{}` — no nil-check gymnastics
- **DRY**: The `if r.TelemetryMgr == nil || !r.TelemetryMgr.IsEnabled()` guard (repeated 6+ times across 3 controllers) is eliminated
- **Incremental**: No new event-bus infrastructure required — just interfaces and one wrapper

WentingWu666666 (commented):
## Recommendation: Use `metadata.uid` as Cluster ID

> As discussed in our last meeting, I recommended using `documentdb.metadata.uid` as the cluster telemetry ID instead of generating and persisting a separate UUID.

### Current approach (UUID annotation in `guid.go`)

The PR generates a random `uuid.New()`, persists it as a `telemetry.documentdb.io/cluster-id` annotation via `client.Update()`, and reads it back on subsequent reconciliations. This requires ~113 lines in `guid.go` plus a `GUIDManager` struct.

**Known weaknesses:**
- **Race condition**: If the operator crashes after generating the UUID but *before* `client.Update()` succeeds, a new UUID is generated on restart — telemetry correlation is silently broken.
- **Extra API write**: Every new resource triggers a `client.Update()` solely to persist the annotation, which also **re-triggers the reconcile loop** (annotation change fires the watch).
- **Complexity**: `GUIDManager` with `GetOrCreate*` / `Get*` / `Set*` methods across 3 resource types is a lot of machinery for what is fundamentally "give me a stable ID for this object."

### Recommended approach: `metadata.uid` directly

Every Kubernetes resource already has a UID assigned atomically by the API server at creation time. It is:
- **Always present** — available the instant the reconciler sees the object
- **Immutable** — never changes for the lifetime of the resource
- **Unique** — guaranteed unique across the cluster (UUID v4 format)
- **Zero-cost** — no extra writes, no race conditions, no re-triggering reconcile loops
- **Not sensitive** — it's an opaque random UUID, contains no PII, no namespace name, no cluster name. Hashing it would add no real privacy benefit.

The entire `guid.go` file collapses to:

```go
func GetResourceTelemetryID(obj client.Object) string {
    return string(obj.GetUID())
}
```

Apply the same pattern for backups and scheduled backups — every K8s resource has its own `metadata.uid`.

### Comparison with the options in `telemetry-guid-strategy.md`

| Criteria | Current (UUID annotation) | `metadata.uid` | Hash(ns+name) (2A) | Hash(ns+name+uid) (2B) |
|---|---|---|---|---|
| Survives operator restart | ⚠️ Only if write succeeded | ✅ Always | ✅ Always | ✅ Always |
| Extra API writes | 1 per resource | 0 | 0 | 0 |
| Re-triggers reconcile | Yes (annotation change) | No | No | No |
| Recreated cluster = new ID | Yes (new UUID) | Yes (new UID) | ❌ No (same hash) | Yes (new hash) |
| Implementation complexity | ~113 lines + GUIDManager | ~3 lines | ~10 lines | ~10 lines |

### Why `metadata.uid` is better than Options 2A/2B

- **vs 2A** (hash without UID): `metadata.uid` correctly distinguishes a deleted-and-recreated cluster from the original. 2A merges them into one telemetry identity, which hides lifecycle events like "user had to recreate the cluster."
- **vs 2B** (hash with UID): Functionally equivalent behavior, but `metadata.uid` is simpler — there is no point in hashing three inputs to derive a unique value when the UID alone is already globally unique.

### Summary

Using `metadata.uid` directly eliminates `GUIDManager`, the `telemetry.documentdb.io/*` annotations, and all `GetOrCreate*` / `Get*` / `Set*` methods — roughly 110 lines of code — while fixing the restart race condition. No hashing needed since the UID is already an opaque random identifier.

docs/designs/telemetry-configuration.md
```diff
@@ -0,0 +1,134 @@
...
+## Configuration
+
+### Environment Variables
+
+Configure telemetry by setting these environment variables in the operator deployment:
+
+| Variable | Description | Required |
+|----------|-------------|----------|
+| `APPINSIGHTS_INSTRUMENTATIONKEY` | Application Insights instrumentation key | Yes (or connection string) |
+| `APPLICATIONINSIGHTS_CONNECTION_STRING` | Application Insights connection string (alternative to instrumentation key) | Yes (or instrumentation key) |
```
hossain-rayhan, line 18: APPINSIGHTS vs APPLICATIONINSHIGHTS. We might stick to one name.

docs/designs/telemetry-configuration.md
```diff
@@ -0,0 +1,134 @@
...
+
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: documentdb-operator-telemetry
+  namespace: documentdb-system
+type: Opaque
+stringData:
+  APPINSIGHTS_INSTRUMENTATIONKEY: "YOUR-INSTRUMENTATION-KEY-HERE"
```
hossain-rayhan, line 52: Another underscore would be good separation: APPINSIGHTS_INSTRUMENTATION _KEY

docs/designs/telemetry-configuration.md
```diff
@@ -0,0 +1,134 @@
...
+   env:
+     - name: DOCUMENTDB_TELEMETRY_ENABLED
+       value: "false"
+   ```
+
+2. **Via Helm**:
+   ```yaml
+   telemetry:
+     enabled: false
+   ```
```
hossain-rayhan, line 108: Can they turn off telemetry on an already provisioned/running cluster?

docs/designs/telemetry-configuration.md
Ritvik-Jayaswal: These are actually official Microsoft names that the AppInsights Go SDK expects. I included both because a user could use either one to pass in their own string. I added a section in the design doc that explains this a bit more.

APPINSIGHTS_INSTRUMENTATIONKEY — Takes a bare instrumentation key (just the GUID, e.g., f5614a64-9358-44db-b19b-18a2eb54f623). This is the legacy method from the original Application Insights SDK. The Go SDK we use ([microsoft/ApplicationInsights-Go](https://github.com/microsoft/ApplicationInsights-Go)) reads this variable.

APPLICATIONINSIGHTS_CONNECTION_STRING — Takes a full connection string (e.g., InstrumentationKey=f5614a64-...;IngestionEndpoint=https://westus2.in.applicationinsights.azure.com/). Based on modern connection string from https://learn.microsoft.com/en-us/azure/azure-monitor/app/connection-strings

docs/designs/telemetry-configuration.md
Ritvik-Jayaswal: see above comment

docs/designs/telemetry-configuration.md
Ritvik-Jayaswal: Yes, users can do helm upgrade --set telemetry.enabled=false. I can add this to more documentation if needed.

docs/designs/telemetry-configuration.md
WentingWu666666: ⚠️ microsoft/ApplicationInsights-Go is archived and unsupported

The Go SDK this PR depends on (github.com/microsoft/ApplicationInsights-Go) is archived and explicitly unsupported:

 "This SDK is NOT currently maintained or supported by Microsoft. Azure Monitor only provides support when using our supported SDKs, and this SDK does not yet meet that
 standard."

Known gaps listed by Microsoft themselves:

 - No operation correlation
 - No sampling support
 - No automatic collection
 - No offline storage

Building a new telemetry feature on an archived, unsupported dependency is risky — any bugs or security issues won't get upstream fixes, and it may break with future App
Insights backend changes.

Recommended alternative: OpenTelemetry Go SDK + OTel Collector sidecar

Microsoft's recommended path for Go is the standard OpenTelemetry Go SDK (go.opentelemetry.io/otel) + an OTel Collector with the Azure Monitor exporter. This is:

 - ✅ Actively maintained (CNCF project)
 - ✅ Microsoft's official recommendation for languages without a supported App Insights SDK
 - ✅ Vendor-neutral — just swap the exporter if we ever need to export elsewhere
 - ✅ Supports connection strings, regional endpoints, Entra ID auth

We already have an OTel Collector sidecar architecture in progress — see [PR #286](https://github.com/documentdb/documentdb-kubernetes-operator/pull/286) by Uri, where I've left a review comment recommending this approach. Your telemetry PR would build on top of that foundation: the operator emits custom events/metrics via the OTel SDK, and the collector routes them to App Insights. This unifies the telemetry story — one SDK (OTel) for both user-facing monitoring and product telemetry.

I'd suggest coordinating with Uri to get [PR #286](https://github.com/documentdb/documentdb-kubernetes-operator/pull/286) merged first. If needed, I can help create a simpler PR to add the collector sidecar to unblock your work here.

PR #286, opened 2026-03-05
feat(telemetry): local telemetry playground using injected OTel sidecar (#342)

udsmicrosoft:
## Summary

Adds a self-contained local telemetry playground (Kind + Prometheus + Grafana + dashboards + traffic generators) and aligns it with the **injected OTel Collector sidecar** introduced by #342.

## Architecture (re-aligned on #342)

Every DocumentDB pod runs a **per-pod OTel Collector sidecar** injected by the CNPG sidecar-injector plugin (`spec.monitoring.enabled=true`). There is **no central OTel Collector Deployment** and **no per-node DaemonSet** in this stack:

- **Gateway metrics** — gateway pushes OTLP/gRPC → sidecar's `otlp` receiver on `127.0.0.1:4317` → re-exported via the `prometheus` exporter on `:9188`.
- **Container/node metrics** — Prometheus scrapes kubelet/cAdvisor directly (no OTel involvement).
- **CNPG metrics** — Prometheus scrapes the CNPG instance manager on `:9187` directly.

```
documentdb pod
├─ postgres            (CNPG, /metrics :9187)
├─ documentdb-gateway  (OTLP push → 127.0.0.1:4317)
└─ otel-collector      (OTLP in :4317 / Prom out :9188)
```

## Operator changes (under `operator/`)

- **`internal/otel/base_config.yaml`** — adds an `otlp` receiver bound to `127.0.0.1:4317` (loopback only; gateway and collector share pod netns) and wires it into the `metrics` pipeline alongside the existing `sqlquery` receiver.
- **`cnpg-plugins/sidecar-injector/internal/lifecycle/lifecycle.go`** — when injecting the sidecar, also adds OTel env vars to the gateway container:
  - `POD_NAME` (downward API)
  - `OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4317`
  - `OTEL_METRICS_ENABLED=true` ← required: the gateway gates OTel init off by default and checks this env var
  - `OTEL_RESOURCE_ATTRIBUTES=service.instance.id=$(POD_NAME)`

  Injection is **idempotent across CREATE and PATCH** via name-based dedup — without this, repeated reconciles double-append entries and CNPG fails with `Pod is invalid: spec: Forbidden: pod updates may not change fields other than ...`.
- **New** `lifecycle_test.go` covers: all-appended, metrics-enabled-present, preserves-existing, idempotent re-invocation, no-gateway-container.
- `internal/otel/config_test.go` extended for the new `otlp` receiver and pipeline wiring.

## Playground (`documentdb-playground/telemetry/local/`)

One-command deploy: `./scripts/deploy.sh` — Kind cluster + cert-manager + operator (from this branch) + observability stack + DocumentDB HA + traffic generators.

| Component | Purpose |
|---|---|
| 3-instance DocumentDB HA cluster | 1 primary + 2 streaming replicas |
| Per-pod OTel Collector sidecar | Injected by the operator's CNPG plugin |
| Prometheus | Scrapes sidecars (annotation discovery), CNPG, kubelet/cAdvisor |
| Grafana | Two dashboards: Gateway, Internals |
| Traffic generators | RW to primary, RO to replicas, with deliberate error injection |

Dashboards' Instance variable pivots on the `pod` label (set by the collector's `resource` processor) — Prometheus's scrape `instance` label clobbers OTel's `service.instance.id`.

Sidecar Prometheus port is **9188** (not the conventional 9187, which collides with the CNPG instance-manager in the same pod netns).

`validate.sh` polls up to 120s for `db_client_operations_total` and **fails red** if absent — surfaces gateway-image issues clearly instead of warning.

## ⚠️ Gateway image requirement

The default upstream gateway image does **not** yet emit OTLP `db_client_*` metrics — that instrumentation lives in pgmongo and has not been published in an upstream release. The playground's `cluster.yaml` therefore pins:

```yaml
gatewayImage: "udsmiley/documentdb-gateway-otel:k8s-pgmongo-main-latest"
```

Once an upstream release ships with OTel metrics, drop the pin to fall back to the operator default. README §Gateway image documents the build recipe (pgmongo `oss` → emulator-shape image → re-wrap with this repo's `Dockerfile_gateway_public_image` for the K8s slim shape).

## Documentation (`docs/.../monitoring/`)

- **Architecture overview** — sidecar injection flow, operator's `base_config.yaml`, ServiceMonitor examples
- **Metrics reference** — gateway metrics, container resources, CNPG/PostgreSQL metrics, OTel naming conventions
- Updated `mkdocs.yml` navigation

### DIFF
--- a/docs/operator-public-documentation/preview/monitoring/overview.md
+++ b/docs/operator-public-documentation/preview/monitoring/overview.md
@@ -1,0 +1,232 @@
+---
+title: Monitoring Overview
+description: How to monitor DocumentDB clusters using the in-pod OpenTelemetry Collector sidecar, Prometheus, and Grafana.
+tags:
+  - monitoring
+  - observability
+  - metrics
+  - opentelemetry
+---
+
+# Monitoring Overview
+
+This guide describes how to monitor DocumentDB clusters running on Kubernetes using the operator's built-in OpenTelemetry Collector sidecar, Prometheus, and Grafana.
+
+## Prerequisites
+
+- A running Kubernetes cluster with the DocumentDB operator installed
+- [Helm 3](https://helm.sh/docs/intro/install/) for deploying Prometheus and Grafana
+- [kubectl](https://kubernetes.io/docs/tasks/tools/) configured for your cluster
+- [`jq`](https://jqlang.github.io/jq/) for processing JSON in verification commands
+
+## Architecture
+
+When `spec.monitoring.enabled: true` is set on a `DocumentDB` resource, the operator instructs the CNPG sidecar-injector plugin to add an **`otel-collector` container to every cluster pod**. The collector runs alongside the `postgres` and `documentdb-gateway` containers and exposes a Prometheus `/metrics` endpoint that any scraper (Prometheus, Datadog Agent, Grafana Alloy, …) can consume.
+
+```mermaid
+graph TB
+    subgraph pod["DocumentDB Pod (one per cluster instance)"]
+        pg["postgres :5432"]
+        gw["documentdb-gateway :10260<br/>OTLP push → localhost:4317"]
+        otel["otel-collector sidecar<br/>otlp :4317 in / prometheus :9187 out"]
+        gw -. OTLP gRPC .-> otel
+    end
+
+    prom["Prometheus<br/>(annotation-based pod discovery)"]
+    cadv["kubelet / cAdvisor<br/>(per-node)"]
+    grafana["Grafana"]
+
+    prom -- "scrape :9187" --> otel
+    prom -- "/metrics/cadvisor" --> cadv
+    prom --> grafana
+```
+
+Key points:
+
+- **One collector per pod** — no central Deployment, no ExternalName bridge, no per-instance Service is required.
+- **Gateway → sidecar push** — the sidecar-injector sets `OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4317` and `OTEL_RESOURCE_ATTRIBUTES=service.instance.id=$(POD_NAME)` on the gateway container so each pod's metrics are tagged with the originating pod name.
+- **Prometheus discovery via pod annotations** — the injector adds `prometheus.io/scrape=true`, `prometheus.io/port=<port>`, and `prometheus.io/path=/metrics` annotations so a standard pod-annotation scrape config works out of the box.
+- **Container & node metrics come from kubelet/cAdvisor directly** — Prometheus scrapes them natively; no DaemonSet collector is needed.
+
+### Enabling the sidecar
+
+```yaml
+apiVersion: db.microsoft.com/preview
+kind: DocumentDB
+metadata:
+  name: my-cluster
+spec:
+  monitoring:
+    enabled: true
+    exporter:
+      prometheus:
+        port: 9187      # the port the sidecar exposes /metrics on
+```
+
+Once applied, the operator generates an `OpenTelemetryConfig` ConfigMap and the sidecar-injector adds the `otel-collector` container to every CNPG instance pod. Pods will be 3/3 (`postgres` + `documentdb-gateway` + `otel-collector`).
+
+### Collector pipeline (current)
+
+The collector ships with a minimal pipeline:
+
+| Stage | Components |
+|-------|------------|
+| Receivers | `otlp` (gRPC `0.0.0.0:4317`) and `sqlquery` (a stub `documentdb.postgres.up` query) |
+| Processors | `batch`, `resource` (adds `service.instance.id`, namespace, cluster, pod metadata) |
+| Exporters | `prometheus` on the configured port (default `8888`; the playground uses `9187`) |
+
+The pipeline is deep-merged from an embedded static config (`base_config.yaml`) and a dynamic config rendered by the operator. Changes to either trigger a content-hash update on the ConfigMap; the sidecar-injector compares hashes and rolls pods only when the config actually changes.
+
+### Postgres metrics — deferred
+
+The current pipeline includes only a placeholder `documentdb.postgres.up` SQL query. Comprehensive Postgres metrics (replication lag, backends, commits/rollbacks, WAL age, db size, index stats, etc.) will be added in a follow-up by extending the SQL queries in `operator/src/internal/otel/base_config.yaml`. Until then, dashboards in this repository focus on **gateway metrics** and **container/node metrics**.
+
+## Prometheus Integration
+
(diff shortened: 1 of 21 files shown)


docs/operator-public-documentation/preview/monitoring/overview.md
```diff
@@ -0,0 +1,284 @@
...
+- **Gateway container** — MongoDB-compatible API sidecar that exports telemetry via OTLP
+
+The gateway sidecar injector automatically configures each gateway container with:
+
+- `OTEL_EXPORTER_OTLP_ENDPOINT` — points to an OpenTelemetry Collector service
+- `OTEL_RESOURCE_ATTRIBUTES` — sets `service.instance.id` to the pod name for per-instance metric attribution
+
+The recommended monitoring stack collects three signals — **metrics**, **traces**, and **logs** — from these containers and stores them for visualization and alerting.
+
+```
```
xgerman: please make a mermaid graph

documentdb-playground/telemetry/local/README.md
```diff
@@ -0,0 +1,182 @@
...
+kubectl port-forward svc/grafana 3000:3000 -n observability --context kind-documentdb-telemetry
+
+# Access Prometheus
+kubectl port-forward svc/prometheus 9090:9090 -n observability --context kind-documentdb-telemetry
+
+# Validate data is flowing
+./scripts/validate.sh
+
+# Tear down
+./scripts/teardown.sh
```
alaye-ms: Maybe this could have a separate heading for cleanup instead of being in quickstart

alaye-ms (approved):

WentingWu666666 (commented):
**docs/operator-public-documentation/preview/monitoring/overview.md (lines 23-30)**

Instead of asking users to build their own gateway image from a specific branch, it would be better to publish a pre-built gateway image (e.g., to `ghcr.io/documentdb/...`) that includes the OTEL instrumentation. Requiring users to build the image themselves adds friction and is error-prone  most users won't have the build toolchain set up for the gateway. Could we provide a ready-to-use image with telemetry support, or at least track an issue to do so before GA?

WentingWu666666:
Great work on the documentation, dashboards, and overall polish  the metrics reference and alerting rules are really nice!

I'd like to discuss the collector architecture before we merge. The current design uses a central Deployment + DaemonSet, which has some limitations for production:

**1. Replica blind spot**: The `postgresql` receiver connects to the `-rw` service, which only routes to the primary. In a 3-instance cluster, replicas get zero PG metrics (connections, replication lag, index stats). This is a critical gap  replication lag is the most important metric for replicas, and it's exactly the pod that can't be reached.

**2. Hardcoded endpoint**: The PG endpoint `documentdb-preview-rw.documentdb-preview-ns.svc.cluster.local:5432` is baked into the collector config. This doesn't scale to multiple DocumentDB clusters or different namespaces without manual config duplication.

**3. Cross-namespace coupling**: The ExternalName bridge service is needed because the collector lives in a different namespace. This adds operational complexity and breaks if the CR is renamed.

**Alternative: OTel Collector as a sidecar** (injected by the CNPG plugin)

I've been prototyping this approach on `users/wentingwu/otel-poc`:
- Each pod gets its own OTel Collector sidecar, auto-injected by the sidecar-injector plugin
- Connects to PG via `localhost:5432`  works on primary AND replicas
- Operator generates per-cluster ConfigMap  no hardcoded names
- Enable/disable via `monitoring.enabled` on the CR  zero-downtime toggle via ConfigMap hot-reload
- Resource cost: ~70 MB per pod (measured), negligible CPU

| | Central Deployment | Sidecar |
|--|-------------------|---------|
| PG metrics from replicas |  Only primary via `-rw` |  All pods via localhost |
| Multi-cluster |  Manual config per cluster |  Operator auto-generates |
| Cross-namespace bridge |  Required |  Not needed (same pod) |
| Operator-managed |  Static YAMLs |  Reconciled by controller |
| Resource overhead | 1 central + 1 DaemonSet per node | 1 per pod (~70 MB) |

**Scope suggestion**: For the initial merge, I'd suggest we reduce scope to **VM/container metrics only** (CPU, memory, network, filesystem via `kubeletstats` or `hostmetrics`). These are universally useful, don't require PG credentials, and let us ship something quickly while we align on the PG metrics collection architecture (sidecar vs central).

Happy to walk through the sidecar POC and discuss tradeoffs  the docs and dashboards from this PR would pair perfectly with the sidecar approach.

WentingWu666666:
**Suggestion: Add `kubeletstats` receiver to the OTel sidecar for container resource metrics**

Currently, per-container CPU/memory metrics (cAdvisor) are only available via the Prometheus  kubelet scrape path (`kubelet-cadvisor` job). Users who want pure OTLP export don't get container resource metrics at all.

Since the OTel Collector sidecar already supports both push (OTLP exporter) and pull (Prometheus exporter on `:9188`), we should add a `kubeletstats` receiver to the sidecar config. This would make container resource metrics available through both paths automatically:

```yaml
receivers:
  otlp: ...              # gateway app metrics (existing)
  kubeletstats:          # container resource metrics (new)
    auth_type: serviceAccount
    collection_interval: 30s
    metric_groups: [pod, container]
```

This gives users metrics like `container.cpu.usage` and `container.memory.working_set` per container (postgres, gateway, otel-collector)  via OTLP push or Prometheus pull  with no extra DaemonSet or cluster-level infrastructure.

The only additional requirement is RBAC for the pod's ServiceAccount to read `nodes/stats`.

With this change, the `kubelet-cadvisor` Prometheus scrape job can be removed since the sidecar would already expose those metrics on the Prometheus endpoint.

WentingWu666666:
**Suggestion: Split this PR into smaller, focused PRs by metric category**

This PR covers a lot of ground (2900+ lines, 21 files). The operator core changes (sidecar injection, OTel base_config, lifecycle env var injection) look complete  I'd suggest landing that as-is and then adding metrics support incrementally by category:

1. **Container resource metrics**  Add `kubeletstats` receiver to the sidecar for per-container CPU/memory (postgres, gateway, otel-collector). This is the most fundamental  users need to know their DB's resource usage first.
2. **Gateway metrics**  OTLP receiver wiring for gateway-specific `db_client_*` metrics, dashboard.
3. **CNPG metrics** (optional)  CNPG instance manager metrics are already exposed on `:9187`; integrating them into the OTel pipeline is nice-to-have.

Each can be its own PR with matching docs, dashboards, and playground updates. Easier to review, easier to revert, and unblocks the foundational pieces while the rest iterates.

PR #313, opened 2026-03-16
feat: add KEDA autoscaling playground for DocumentDB

xgerman:
## Summary

Adds a self-contained playground demonstrating [KEDA](https://keda.sh/) event-driven autoscaling with DocumentDB. KEDA's [MongoDB scaler](https://keda.sh/docs/2.19/scalers/mongodb/) polls a DocumentDB collection for pending jobs and automatically scales a worker Deployment from 0 to N pods.

**E2E tested on Kind** — full autoscaling cycle verified: 0 → 2 pods on job insert, 2 → 0 pods after drain.

## New files (10)

| File | Description |
|------|-------------|
| `manifests/documentdb-instance.yaml` | DocumentDB CR for Kind (1 node, 2Gi, ClusterIP) |
| `manifests/keda-trigger-auth.yaml` | `ClusterTriggerAuthentication` with DocumentDB connection string |
| `manifests/keda-scaled-object.yaml` | `ScaledObject` with MongoDB trigger on `appdb.jobs` |
| `manifests/job-worker.yaml` | Worker Deployment (scaled by KEDA, starts at 0) |
| `manifests/seed-jobs.yaml` | Job that inserts 10 pending documents |
| `manifests/drain-jobs.yaml` | Job that marks all pending documents as completed |
| `scripts/setup.sh` | Installs KEDA, deploys DocumentDB, configures demo |
| `scripts/teardown.sh` | Cleanup with optional KEDA/DocumentDB removal |
| `scripts/demo.sh` | Interactive walkthrough of scaling behavior |
| `README.md` | Quick start, connection guide, gotchas |

## Key findings from E2E testing

| Finding | Detail |
|---------|--------|
| **`tlsInsecure=true` required** | KEDA's Go MongoDB driver needs `tlsInsecure=true` (not `tlsAllowInvalidCertificates=true`) to skip both cert AND hostname verification with self-signed certs |
| **`directConnection=true` works** | KEDA passes this through correctly |
| **`SCRAM-SHA-256` works** | Standard MongoDB auth handshake compatible |
| **Port 10260** | Must be explicit in connection string |
| **`exposeViaService` required** | DocumentDB CR needs `exposeViaService.serviceType: ClusterIP` for the gateway service to be created |

## Test results

```
Seed 10 pending jobs → ScaledObject ACTIVE: True → HPA TARGETS: 5/5 → 2 worker pods
Drain all jobs       → ScaledObject ACTIVE: False → HPA TARGETS: 0/5 → 0 worker pods (after 30s cooldown)
```

### DIFF
--- a/documentdb-playground/keda-autoscaling/README.md
+++ b/documentdb-playground/keda-autoscaling/README.md
@@ -128,6 +128,10 @@ Interactive walkthrough that seeds jobs, watches scaling, drains jobs, and watch
 > attempts replica set discovery, which fails with DocumentDB because it doesn't expose a standard
 > MongoDB replica set topology.
 
+> **Note:** Remove the `replicaSet=rs0` parameter from the connection string. KEDA's Go driver
+> fails topology negotiation with DocumentDB when `replicaSet` is specified alongside
+> `directConnection=true`. The setup script strips this parameter automatically.
+
 > **Note:** A `ClusterTriggerAuthentication` is used because the DocumentDB credentials Secret
 > (`documentdb-ns`) and the ScaledObject (`app`) are in different namespaces. If you deploy
 > everything in the same namespace, you can use a namespace-scoped `TriggerAuthentication` instead.
--- a/documentdb-playground/keda-autoscaling/README.md
+++ b/documentdb-playground/keda-autoscaling/README.md
@@ -1,0 +1,151 @@
+# KEDA autoscaling with DocumentDB
+
+This playground demonstrates event-driven autoscaling using [KEDA](https://keda.sh/) with DocumentDB. KEDA's [MongoDB scaler](https://keda.sh/docs/2.19/scalers/mongodb/) polls a DocumentDB collection for pending jobs and automatically scales a worker Deployment from 0 to N pods based on the query result.
+
+## Architecture
+
+```mermaid
+flowchart LR
+    subgraph cluster["Kubernetes cluster"]
+        KEDA[KEDA Operator] -->|"polls appdb.jobs<br/>every 5s"| DDB[(DocumentDB<br/>port 10260)]
+        KEDA -->|scales 0 → N| Worker["job-worker<br/>Deployment"]
+        Seed[seed-jobs Job] -->|inserts pending jobs| DDB
+        Drain[drain-jobs Job] -->|marks jobs completed| DDB
+    end
+```
+
+## Prerequisites
+
+| Tool | Version | Purpose |
+|------|---------|---------|
+| [Kind](https://kind.sigs.k8s.io/) | 0.20+ | Local Kubernetes cluster |
+| [kubectl](https://kubernetes.io/docs/tasks/tools/) | 1.30+ | Kubernetes CLI |
+| [Helm](https://helm.sh/docs/intro/install/) | 3.x | Package manager for Kubernetes |
+| DocumentDB operator | — | Must be running in the Kubernetes cluster |
+
+> **Note:** If you don't have a Kubernetes cluster with the DocumentDB operator, use the
+> [development deploy script](https://github.com/documentdb/documentdb-kubernetes-operator/blob/main/operator/src/scripts/development/deploy.sh)
+> to set up Kind with the operator:
+>
+> ```bash
+> cd operator/src
+> DEPLOY=true DEPLOY_CLUSTER=true ./scripts/development/deploy.sh
+> ```
+
+## Quick start
+
+```bash
+# 1. Ensure the DocumentDB operator is running on your Kubernetes cluster
+
+# 2. Deploy KEDA and the demo
+cd documentdb-playground/keda-autoscaling
+./scripts/setup.sh
+
+# 3. Watch pods scale up (10 pending jobs trigger scaling)
+kubectl get pods -n app -w
+
+# 4. Drain jobs to scale back to 0
+kubectl delete job drain-pending-jobs -n app --ignore-not-found
+kubectl apply -f manifests/drain-jobs.yaml
+
+# 5. Clean up
+./scripts/teardown.sh
+```
+
+## How it works
+
+1. **`setup.sh`** installs KEDA and deploys a DocumentDB instance with a worker Deployment.
+2. A `ClusterTriggerAuthentication` stores the DocumentDB connection string (with TLS and auth parameters) in a Secret.
+3. A `ScaledObject` configures KEDA to poll the `appdb.jobs` collection for documents matching `{"status": "pending"}` every 5 seconds.
+4. When the number of pending jobs exceeds 5, KEDA creates an HPA that scales the `job-worker` Deployment from 0 up to 10 pods.
+5. The **seed job** inserts 10 pending documents to trigger scaling.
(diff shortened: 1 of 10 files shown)


xgerman (commented):
## Code Review: KEDA Autoscaling Playground

### Summary
Well-structured playground that demonstrates KEDA event-driven autoscaling with DocumentDB. Clean separation of manifests, scripts, and documentation. The E2E-tested findings (especially `tlsInsecure=true`) are valuable. A few items to address before merge.

---

### 🟠 Major

**1. [Required] README references wrong container image (README.md:135)**
The README says `mongodb/mongodb-community-server:8.0-ubuntu2404` but the actual manifests use `mongo:8.0`. This was fixed in the second commit but the README note was not updated.

```diff
-> **Note:** The `mongodb/mongodb-community-server:8.0-ubuntu2404` image is used for the seed and
+> **Note:** The `mongo:8.0` image is used for the seed and
```

**2. [Required] `setup.sh` should use `kubectl get documentdb` for connection string (setup.sh:128-136)**
The setup script manually constructs the connection string from the secret and hardcoded service name pattern. The DocumentDB resource provides the connection string in `.status.connectionString`. Using it would be more resilient to service naming changes:

```bash
RAW_CONN=$(kubectl get documentdb "$DOCUMENTDB_NAME" -n "$DOCUMENTDB_NAMESPACE" \
    -o jsonpath='{.status.connectionString}')
conn_string=$(eval echo "$RAW_CONN")
```

Note: The status connection string uses `tlsAllowInvalidCertificates=true`, but KEDA needs `tlsInsecure=true`. You may still need a `sed` replacement or keep the manual construction with a comment explaining why.

**3. [Required] `ClusterTriggerAuthentication` missing `namespace` on secret reference (keda-trigger-auth.yaml)**
`ClusterTriggerAuthentication` looks up secrets from the KEDA operator namespace by default. The `secretTargetRef` should explicitly specify the namespace where the secret lives, otherwise it depends on KEDA's default namespace behavior:

```yaml
spec:
  secretTargetRef:
    - parameter: connectionString
      name: documentdb-keda-connection
      namespace: keda          # explicit namespace
      key: connectionString
```

This works today because `setup.sh` creates the secret in the KEDA namespace, but it's fragile — if someone changes `KEDA_NAMESPACE`, the manifest and secret would diverge silently.

---

### 🟡 Minor

**4. [Suggestion] Hardcoded credentials in setup.sh (setup.sh:85-86)**
The password `KedaDemo2024!` is hardcoded. For a playground this is acceptable, but consider adding a comment that users should change it, or accept it via an environment variable:

```bash
DOCUMENTDB_USER="${DOCUMENTDB_USER:-docdbadmin}"
DOCUMENTDB_PASS="${DOCUMENTDB_PASS:-KedaDemo2024!}"
```

**5. [Suggestion] `documentdb-instance.yaml` includes `sidecarInjectorPluginName` (line 17)**
This field is used for CNPG-I sidecar injection. While other playgrounds include it, it's not required for the KEDA demo to function. If you keep it, add a comment explaining its purpose, since it may confuse users who just want KEDA autoscaling.

**6. [Suggestion] `demo.sh` uses `timeout` command (demo.sh:36,47)**
The `timeout` command behaves differently on macOS (`gtimeout` via coreutils) vs Linux. Since this targets Kind (which can run on macOS), consider a note or a fallback:

```bash
TIMEOUT_CMD="timeout"
command -v gtimeout &>/dev/null && TIMEOUT_CMD="gtimeout"
```

**7. [Suggestion] No `.gitignore` or `.helmignore`**
Not blocking, but adding a minimal `.gitignore` would prevent accidental commits of local test artifacts.

---

### 🟢 Nitpick

**8. [Nitpick] Shebang inconsistency**
Scripts use `#!/bin/bash` while the LightRAG playground uses `#!/usr/bin/env bash`. The `env` form is more portable. Not blocking — just a consistency note.

**9. [Nitpick] `seed-jobs.yaml` and `drain-jobs.yaml` use `mongo:8.0` without pinning a digest**
For reproducibility, consider pinning to a specific tag like `mongo:8.0.4` or adding a comment noting the tag is intentionally floating.

---

### ✅ Positive Feedback

- **Excellent README** — the connection string table (lines 72-78) with the `tlsInsecure` finding is genuinely useful documentation that will save other users from the same debugging session.
- **Clean script structure** — `setup.sh` is well-organized with clear functions, color-coded output, and proper prerequisite checks.
- **Good teardown ordering** — deleting the `ScaledObject` before the worker deployment avoids KEDA errors. This attention to deletion ordering shows operational experience.
- **Idempotent `--dry-run=client | kubectl apply`** pattern for secrets and namespaces is the right approach.
- **Consistent labeling** — `app.kubernetes.io/part-of: keda-documentdb-demo` across all resources enables clean bulk operations.
- **Mermaid architecture diagram** in the README is a nice touch.

### Verdict
`/request-changes` — Fix the README image reference (#1) and consider the connection string approach (#2). The rest are suggestions.

hossain-rayhan (approved):

xgerman:
Addressed all review comments in commit 9f103de:

- **No hardcoded namespaces**: Removed `namespace:` from all 5 manifest YAML files; scripts now use `-n "$APP_NAMESPACE"` flag
- **Configurable secret name**: `teardown.sh` now uses `$DOCUMENTDB_SECRET` variable instead of hardcoded `documentdb-credentials`
- **macOS portability**: Replaced `timeout` command in `demo.sh` with `perl -e 'alarm N; exec @ARGV'` (portable on macOS/Linux)
- **Manifest placeholders**: `documentdb-instance.yaml` uses sed-substituted placeholders for name/secret
- **DCO sign-off**: Commit includes `Signed-off-by` trailer

WentingWu666666 (commented):
## KEDA Playground - Local Kind Cluster Test Results

I tested this playground end-to-end on a local Kind cluster (K8s v1.35.0) and the **full autoscaling cycle works** (0  2  0 pods) 

### Issues Encountered

1. **Connection string TLS parameter**: KEDA's Go MongoDB driver requires `tlsInsecure=true`, not `tlsAllowInvalidCertificates=true`. The setup script handles this correctly via `sed`, but it's worth documenting this distinction for users who manually construct connection strings.

2. **`replicaSet=rs0` must be removed**: When combined with `directConnection=true`, the `replicaSet` parameter causes KEDA to fail topology negotiation. The setup script strips it, but the README should mention this as a known requirement.

3. **Cross-namespace secret and DNS resolution**: The connection string from DocumentDB uses the ClusterIP, which doesn't resolve across namespaces. The setup script replaces it with the DNS name (`<svc>.<ns>.svc.cluster.local`), which is correct. Consider documenting this for users deploying in different namespace topologies.

### Suggestion

Consider adding a **Troubleshooting** section to the README covering these connection string gotchas, since users who deviate from the script (or debug failures) will likely hit these issues.

WentingWu666666 (commented):
## Correction to my previous comment

The connection string issues I mentioned (tlsInsecure, replicaSet removal, DNS name) are **already handled correctly by the setup script**. I hit them only because I replicated steps manually on Windows (no bash available) instead of running `setup.sh` directly.

The script is well-written and accounts for all these edge cases. 

My only remaining suggestion: consider adding a brief **Troubleshooting** section to the README for users who debug or customize the setup outside the script, explaining *why* these transformations are needed.

WentingWu666666 (approved):

PR #341, opened 2026-04-16
feat: propagate mutable spec fields and validate immutable fields

### LINKED ISSUE #298: feat: Support PVC resize by propagating pvcSize changes to CNPG Cluster
## Problem

When a user updates `spec.resource.storage.pvcSize` on an existing DocumentDB CR, the change is accepted by the API server but **never propagated** to the underlying CNPG Cluster object or PVCs. The resize is silently ignored.

### Verified behavior on Kind cluster:

| Layer | After patching `pvcSize` from `1Gi` to `2Gi` |
|-------|------|
| DocumentDB CR | Updated to `2Gi`  |
| CNPG Cluster `spec.storage.size` | Stayed at `1Gi`  |
| PVCs | Stayed at `1Gi`  |

### Root cause

In `operator/src/internal/controller/physical_replication.go`, `TryUpdateCluster()` returns early (`nil, -1`) when `ReplicaCluster` is nil, which is the case for non-replicating clusters. This means storage size changes are never compared or patched on the CNPG Cluster object. Even for replicating clusters, storage is not in the patched field set.

## Proposed solution

1. In the reconciliation loop, detect when `pvcSize` in the DocumentDB CR differs from the current CNPG Cluster's `spec.storage.size`
2. Patch the CNPG Cluster's `spec.storage.size` field to match the desired size
3. CNPG will then handle the actual PVC expansion (requires `allowVolumeExpansion: true` on the StorageClass)
4. Add validation: reject shrink attempts (PVC expansion is one-way)

### Files to modify

- `operator/src/internal/controller/physical_replication.go`  Add storage size comparison and patching
- `operator/src/internal/controller/documentdb_controller.go`  Ensure storage changes trigger CNPG cluster update
- Consider adding a webhook to reject `pvcSize` decreases

WentingWu666666:
## Summary

Extends the consolidated `SyncCnpgCluster()` entry point (from PR #336) to propagate additional mutable spec fields to the underlying CNPG Cluster, and adds validation to reject changes to immutable fields.

This is part 2 of the spec change propagation work (fixes #298, #308).

### Mutable Field Sync (`cnpg_sync.go`, `cnpg_patch.go`)

Added propagation for 6 new mutable fields via atomic JSON Patch:

| DocumentDB Field | CNPG Cluster Path | Restart Needed? |
|---|---|---|
| `instancesPerNode` | `/spec/instances` | No (CNPG native) |
| `resource.storage.pvcSize` | `/spec/storage/size` | No (CNPG native) |
| `postgresImage` | `/spec/imageName` | No (CNPG native — image mismatch detected via `checkPodImageIsOutdated`) |
| `logLevel` | `/spec/logLevel` | No (CNPG native — PodSpec drift detected via `checkPodSpecIsOutdated`) |
| `affinity` | `/spec/affinity` | No (CNPG native — PodSpec drift detected via `ComparePodSpecs`) |
| `timeouts.stopDelay` | `/spec/maxStopDelay` | No (CNPG native — maps to `terminationGracePeriodSeconds`, drift triggers rollout) |

All 6 fields are handled natively by CNPG's PodSpec drift detection or image comparison logic. The operator only needs to patch the CNPG Cluster spec — no manual restart annotation is required for any mutable spec field. (The original implementation set `specChanged = true` for 4 of these fields, which was removed after verifying via CNPG v1.28.1 source and live Kind cluster testing that CNPG handles all restarts natively.)

### Immutable Field Validation — Hybrid CEL + Webhook Approach

Immutable fields are protected using two complementary mechanisms:

**CEL transition rules** (`documentdb_types.go`) — for simple string fields:

| Field | CEL Rule | Reason |
|---|---|---|
| `documentDbCredentialSecret` | `self == oldSelf` | Baked into running pods at creation |
| `resource.storage.storageClass` | `self == oldSelf` | Cannot migrate existing PVCs |
| `sidecarInjectorPluginName` | `self == oldSelf` | Structural to CNPG plugin configuration |

This follows the same pattern used by CNPG (e.g., `serviceAccountName`), Cilium, Crossplane, Karpenter, and 780+ other Kubernetes projects for simple immutability.

**Webhook validation** (`documentdb_webhook.go`) — for complex fields:

| Field | Reason for Webhook |
|---|---|
| `bootstrap` | Optional pointer field — CEL transition rules don't fire when field goes from set→nil, losing protection |
| `resource.storage.pvcSize` (shrink) | Requires `resource.ParseQuantity()` comparison; PVC expansion is one-way |
| Image rollback | Complex semver parsing + status field access |
| Schema ≤ binary version | Complex version parsing logic |

None of these immutable fields are enforced by CNPG — our validation is the only line of defense.

### Tests

- 8 new sync tests covering each mutable field, combined changes, and restart annotation behavior
- Webhook tests for bootstrap immutability (change, nil-to-nil, unchanged, set→nil removal)
- PVC shrink rejection and grow allowance tests

## Fixes

- Fixes #298 (PVC resize not propagated)
- Fixes #308 (field mutability policy)

### DIFF
--- a/operator/documentdb-helm-chart/crds/documentdb.io_dbs.yaml
+++ b/operator/documentdb-helm-chart/crds/documentdb.io_dbs.yaml
@@ -1311,7 +1311,6 @@ spec:
                   SchemaVersion controls the desired schema version for the DocumentDB extension.
 
                   The operator never changes your database schema unless you ask:
-                    - Set documentDBVersion → updates the binary (safe to roll back)
                     - Set schemaVersion → updates the database schema (irreversible)
                     - Set schemaVersion: "auto" → schema auto-updates with binary
 
--- a/operator/documentdb-helm-chart/crds/documentdb.io_dbs.yaml
+++ b/operator/documentdb-helm-chart/crds/documentdb.io_dbs.yaml
@@ -1153,6 +1153,9 @@ spec:
                   for the DocumentDB gateway (expects keys `username` and `password`). If omitted,
                   a default secret name `documentdb-credentials` is used.
                 type: string
+                x-kubernetes-validations:
+                - message: credential secret cannot be changed after cluster creation
+                  rule: self == oldSelf
               environment:
                 description: |-
                   Environment specifies the cloud environment for deployment
@@ -1294,6 +1297,9 @@ spec:
                           StorageClass specifies the storage class for DocumentDB persistent volumes.
                           If not specified, the cluster's default storage class will be used.
                         type: string
+                        x-kubernetes-validations:
+                        - message: storage class cannot be changed after cluster creation
+                          rule: self == oldSelf
                     required:
                     - pvcSize
                     type: object
@@ -1304,8 +1310,13 @@ spec:
                 description: |-
                   SchemaVersion controls the desired schema version for the DocumentDB extension.
 
-                  This field decouples the extension binary (image) update from the schema update
-                  (ALTER EXTENSION documentdb UPDATE), providing a rollback-safe upgrade window.
+                  The operator never changes your database schema unless you ask:
+                    - Set documentDBVersion → updates the binary (safe to roll back)
+                    - Set schemaVersion → updates the database schema (irreversible)
+                    - Set schemaVersion: "auto" → schema auto-updates with binary
+
+                  Once the schema has been updated, the operator blocks image rollback below the
+                  installed schema version to prevent running an untested binary/schema combination.
 
                   Values:
                     - "" (empty, default): Two-phase mode. Image upgrades happen automatically,
@@ -1314,7 +1325,7 @@ spec:
                       as it allows rollback by reverting the image before committing the schema change.
                     - "auto": Schema automatically updates to match the binary version whenever
                       the binary is upgraded. This is the simplest mode but provides no rollback
-                      safety window. Recommended for development and testing environments.
+                      safety window.
                     - "<version>" (e.g. "0.112.0"): Schema updates to exactly this version.
                       Must be <= the binary version.
                 pattern: ^(auto|[0-9]+\.[0-9]+\.[0-9]+)?$
@@ -1323,6 +1334,10 @@ spec:
                 description: SidecarInjectorPluginName is the name of the sidecar
                   injector plugin to use.
                 type: string
+                x-kubernetes-validations:
+                - message: sidecar injector plugin name cannot be changed after cluster
+                    creation
+                  rule: self == oldSelf
               timeouts:
                 properties:
                   stopDelay:
--- a/operator/documentdb-helm-chart/crds/documentdb.io_dbs.yaml
+++ b/operator/documentdb-helm-chart/crds/documentdb.io_dbs.yaml
@@ -1152,6 +1152,8 @@ spec:
                   DocumentDbCredentialSecret is the name of the Kubernetes Secret containing credentials
                   for the DocumentDB gateway (expects keys `username` and `password`). If omitted,
                   a default secret name `documentdb-credentials` is used.
+
+                  NOTE: Immutable today; will be relaxed to support credential rotation (see #331).
                 type: string
                 x-kubernetes-validations:
                 - message: credential secret cannot be changed after cluster creation
@@ -1291,6 +1293,7 @@ spec:
                       pvcSize:
                         description: PvcSize is the size of the persistent volume
                           claim for DocumentDB storage (e.g., "10Gi").
+                        minLength: 1
                         type: string
                       storageClass:
                         description: |-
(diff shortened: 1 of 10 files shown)


operator/documentdb-helm-chart/crds/documentdb.io_dbs.yaml
```diff
@@ -1153,6 +1153,9 @@ spec:
                   for the DocumentDB gateway (expects keys `username` and `password`). If omitted,
                   a default secret name `documentdb-credentials` is used.
                 type: string
+                x-kubernetes-validations:
+                - message: credential secret cannot be changed after cluster creation
```
xgerman, line 1159: we were thinking of allowing this for credential rotation. There is a ticket. Please reference the task so we can make it modifiable ocne implemented

operator/documentdb-helm-chart/crds/documentdb.io_dbs.yaml
```diff
@@ -1304,8 +1310,12 @@ spec:
                 description: |-
                   SchemaVersion controls the desired schema version for the DocumentDB extension.
 
-                  This field decouples the extension binary (image) update from the schema update
-                  (ALTER EXTENSION documentdb UPDATE), providing a rollback-safe upgrade window.
+                  The operator never changes your database schema unless you ask:
```
xgerman, line 1316: shoudo point out that auto is only recommended for single-region clusters.

xgerman (changes requested):

alaye-ms (approved):

operator/documentdb-helm-chart/crds/documentdb.io_dbs.yaml
WentingWu666666: Added a comment referencing #331 above the CEL rule:

```
// NOTE: Immutable today; will be relaxed to support credential rotation (see #331).
```

This makes it clear the immutability is intentional but temporary. Relaxing the CEL rule later is a non-breaking change (loosening validation).

operator/documentdb-helm-chart/crds/documentdb.io_dbs.yaml
WentingWu666666: Updated the `auto` description to include:

> Only recommended for single-region clusters.

This is now in both the Go type comment and the generated CRD YAML.

operator/src/internal/webhook/documentdb_webhook.go
```diff
@@ -210,6 +213,61 @@ func (v *DocumentDBValidator) validateImageRollback(newDB, oldDB *dbpreview.Docu
...
+
+	oldQty, errOld := resource.ParseQuantity(oldSize)
+	newQty, errNew := resource.ParseQuantity(newSize)
+	if errOld != nil || errNew != nil {
+		return nil // Let Kubernetes API server handle invalid quantities
+	}
+
+	if newQty.Cmp(oldQty) < 0 {
+		return field.ErrorList{field.Forbidden(
+			field.NewPath("spec", "resource", "storage", "pvcSize"),
```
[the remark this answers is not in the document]
WentingWu666666: Good catch. Applied the suggested approach  `validateStorageResize` now returns `field.Invalid` errors when `ParseQuantity` fails instead of silently returning nil. Added tests for both invalid old and invalid new quantities.

operator/src/internal/webhook/documentdb_webhook.go
```diff
@@ -210,6 +213,61 @@ func (v *DocumentDBValidator) validateImageRollback(newDB, oldDB *dbpreview.Docu
...
+	return allErrs
+}
+
+// validateStorageResize ensures PVC size can only grow, never shrink.
+func (v *DocumentDBValidator) validateStorageResize(newDB, oldDB *dbpreview.DocumentDB) field.ErrorList {
+	oldSize := oldDB.Spec.Resource.Storage.PvcSize
+	newSize := newDB.Spec.Resource.Storage.PvcSize
+	if oldSize == "" || newSize == "" || oldSize == newSize {
+		return nil
+	}
```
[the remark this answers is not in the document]
WentingWu666666: Added `+kubebuilder:validation:MinLength=1` to the `PvcSize` field in the CRD types, so empty strings are now rejected at the API server level before reaching the webhook. Also removed the empty-string early return from the webhook since the CRD now handles it.

operator/documentdb-helm-chart/crds/documentdb.io_dbs.yaml
WentingWu666666: Updated in 30de50f: replaced issue number with plain language. Now reads: 'Immutable today, will be relaxed in a future release to support credential rotation.'

xgerman (approved):

hossain-rayhan (approved):

guanzhousongmicrosoft (approved):

PR #342, opened 2026-04-16
feat(otel): add OTel Collector sidecar injection via CNPG plugin

WentingWu666666:
## Summary

Adds an OpenTelemetry (OTel) Collector sidecar to DocumentDB pods via the CNPG plugin mechanism. The sidecar scrapes PostgreSQL metrics using SQL queries and exposes them via a Prometheus endpoint. Users control the feature through `spec.monitoring` on the DocumentDB CRD.

## Architecture

```
DocumentDB Pod
 postgres         (database)
 documentdb-gateway  (MongoDB wire-protocol proxy)
 otel-collector   (injected by sidecar-injector CNPG plugin when monitoring enabled)
        Reads: static.yaml (embedded SQL metric queries  updated on operator upgrade)
        Reads: dynamic.yaml (per-cluster exporters/attributes  updated on spec change)
        Exposes: Prometheus /metrics endpoint
```

**Rolling restart flow:** operator detects plugin parameter changes  operator adds `kubectl.kubernetes.io/restartedAt` annotation  CNPG performs the pod rollout.

### CRD API (`api/preview/documentdb_types.go`)
- `MonitoringSpec`  top-level monitoring toggle (`enabled: bool`)
- `ExporterSpec`  configures metric export destinations
- `PrometheusExporterSpec`  Prometheus endpoint port (default: 9187)
- `OTLPExporterSpec`  optional OTLP gRPC endpoint

### OTel Config Generation (`internal/otel/`)
- **`base_config.yaml`**  static config embedded via `go:embed`; contains SQL receiver queries (e.g. `documentdb.postgres.up`) and batch processor. Adding new metrics here requires only a YAML edit  no Go code changes.
- **`config.go`**  generates per-cluster `dynamic.yaml` (resource attributes, exporters, pipeline wiring) and a truncated SHA-256 hash for change detection.

### CNPG Integration (`internal/cnpg/`)
- **`cnpg_cluster.go`**  adds OTel plugin parameters (image, configmap name, port, config hash) when monitoring is enabled; removes them when disabled.
- **`cnpg_sync.go`**  syncs OTel params via JSON Patch. When params change, the operator adds a restart annotation in the same atomic patch to trigger a rolling restart.

### Sidecar Injector Plugin (`cnpg-plugins/sidecar-injector/`)
- **`lifecycle.go`**  injects `otel-collector` container when both `otelCollectorImage` and `otelConfigMapName` params are present. Mounts ConfigMap as volume, injects `PGUSER`/`PGPASSWORD` from the app secret, sets `OTEL_EXPORTER_OTLP_ENDPOINT` to `localhost:4317`, and `POD_NAME` via downward API.
- **`config.go`**  parses plugin parameters into typed `Configuration` struct.

### Controller (`internal/controller/documentdb_controller.go`)
- `reconcileOtelConfigMap`  creates/updates the OTel ConfigMap (with owner reference) when monitoring is enabled.
- `deleteOtelConfigMap`  deletes the ConfigMap when monitoring is disabled.

### E2E Tests (`.github/workflows/test-E2E.yml`)
- New test step: enables monitoring  verifies 3/3 containers + ConfigMap + Prometheus `/metrics` returns `documentdb_postgres_up`  disables monitoring  verifies 2/2 containers + ConfigMap deleted.

## Key Behaviors

| Scenario | What Happens |
|----------|-------------|
| Enable monitoring | Operator creates OTel ConfigMap  adds plugin params  restart annotation  CNPG rolls pods  sidecar injected |
| Disable monitoring | Operator deletes ConfigMap  removes plugin params  restart annotation  CNPG rolls pods  sidecar removed |
| Change exporter settings (e.g. port) | Operator regenerates dynamic.yaml  hash changes  restart annotation  pods pick up new config |
| Operator upgrade adds new metrics | New `base_config.yaml` embedded in binary  ConfigMap content changes  hash changes  restart annotation  pods pick up new config |

## Usage

```yaml
apiVersion: documentdb.io/preview
kind: DocumentDB
metadata:
  name: my-cluster
spec:
  monitoring:
    enabled: true
    exporter:
      prometheus:
        port: 9090      # optional, default: 9187
      otlp:
        endpoint: "http://otel-gateway:4317"  # optional
```

## Tests

**Unit tests (15 in `otel/config_test.go`):**
- ConfigMap naming, base_config.yaml embed validation
- GenerateConfigMapData: static/dynamic YAML generation, Prometheus/OTLP/both/none exporter combinations
- HashConfigMapData: determinism, collision resistance, key-order independence
- ResolvePrometheusPort: nil spec, unconfigured, default, custom port

**CNPG integration tests (3 new in `cnpg_sync_test.go`):**
- OTel params added when monitoring enabled
- OTel params removed when monitoring disabled
- Config hash change triggers restart annotation

**Controller tests (4 new in `documentdb_controller_test.go`):**
- reconcileOtelConfigMap: create, update, skip when disabled, owner reference error
- deleteOtelConfigMap: delete existing, no-op when absent, propagate non-NotFound errors

**E2E test (1 new step in `test-E2E.yml`):**
- Full enable  verify sidecar + ConfigMap + metrics  disable  verify cleanup

**Manual Kind cluster testing:**
- All 4 key behaviors verified on a live Kind cluster (see PR conversation for details)


.github/workflows/test-E2E.yml
```diff
@@ -271,6 +271,132 @@ jobs:
         
         echo "✅ DocumentDB status validation passed"
 
+    - name: Test OTel monitoring sidecar enable/disable
```
guanzhousongmicrosoft, line 274: Best to put in a separate script or action

xgerman:
**[Suggestion] 🟠 Use a dedicated least-privilege OTel monitoring user instead of the app user**

The OTel Collector sidecar currently reads `PGUSER`/`PGPASSWORD` from the `<cluster>-app` secret, which is the CNPG application user. This user may have broad read/write access to application databases.

For metrics collection, the collector only needs to run health-check queries (`SELECT 1 as up`) and, in the future, read `pg_stat_*` views. Following the principle of least privilege, consider creating a dedicated monitoring user with only the `pg_monitor` role:

```sql
CREATE ROLE otel_monitor WITH LOGIN;
GRANT pg_monitor TO otel_monitor;
```

Benefits:
- **Least privilege**: The monitoring sidecar can't accidentally read or modify application data
- **Audit clarity**: Monitoring queries are attributable to a separate role in `pg_stat_activity`
- **Future-proof**: When you add `pg_stat_replication`, `pg_stat_bgwriter`, etc. metrics, `pg_monitor` already grants read access to all stats views without needing superuser
- **Blast radius**: If the OTel Collector is compromised (e.g., via a malicious config), the attacker only gets read-only stats access, not app data

Implementation approach:
1. Operator creates the monitoring user via a SQL job (similar to how DocumentDB credentials are managed)
2. Store credentials in a dedicated secret (e.g., `<cluster>-otel-monitor`)
3. Sidecar reads from that secret instead of `<cluster>-app`

This could be a follow-up PR — the current approach works correctly for the initial `SELECT 1` metric. But it should be addressed before adding richer metrics that read from stats views.

xgerman:
**[Suggestion] 🟡 Extensibility gap vs CNPG monitoring — consider user-defined custom queries in a future patch**

I compared our OTel monitoring approach with CNPG's built-in monitoring and wanted to flag an extensibility gap for future consideration.

### How CNPG does it

CNPG has a mature, user-extensible monitoring system:
- **~10 built-in queries**: backends, waiting locks, database size, replication lag, WAL, replication slots, postmaster uptime
- **User-extensible**: Users add custom SQL→metric mappings via ConfigMaps/Secrets (`customQueriesConfigMap` / `customQueriesSecret`) — no operator changes needed
- **Toggleable defaults**: `disableDefaultQueries: true` to use only custom queries
- **Query caching**: `metricsQueriesTTL` controls re-query frequency

### How we do it (this PR)

- **1 built-in query**: `SELECT 1 as up` (liveness only)
- **Not user-extensible**: Queries are in `base_config.yaml` via `go:embed`. Users can't add queries without an operator upgrade.
- **Dual export (advantage!)**: Prometheus + OTLP — CNPG only supports Prometheus

### What we do better
Our OTLP support is a genuine advantage for users with centralized OTel backends (Grafana Tempo, Datadog, etc.). The static/dynamic config split and config hash change detection are also clean.

### Suggested future extensibility

A `customQueries` field that mirrors CNPG's model but uses the OTel sqlquery receiver format:

```yaml
spec:
  monitoring:
    enabled: true
    customQueries:
      - configMapRef:
          name: my-custom-metrics
          key: queries.yaml
```

Where `queries.yaml` uses the OTel `sqlquery` receiver format:

```yaml
queries:
  - sql: "SELECT count(*) as total FROM pg_stat_activity WHERE state = 'active'"
    metrics:
      - metric_name: documentdb.postgres.active_connections
        value_column: total
        data_type: gauge
```

The operator would merge user-provided queries into the generated ConfigMap alongside `base_config.yaml`.

### Not blocking this PR

This is a follow-up suggestion — the current architecture (static/dynamic split, `go:embed` for base config) is well-designed and supports adding this later without breaking changes. The `base_config.yaml` should also be enriched with more built-in queries (backends, replication lag, database size) to match CNPG's baseline — that's a YAML-only change that can land in a subsequent patch.

WentingWu666666:
Great suggestion @xgerman! Agreed  using the app user for OTel metrics violates least-privilege, especially as we add richer metrics (pg_stat_*, replication lag, etc.).

For this PR, the current approach works correctly for the initial `SELECT 1 as up` query. I've opened a follow-up issue to track the dedicated monitoring user work:

 #344  `feat(otel): use dedicated least-privilege monitoring user for OTel Collector sidecar`

This will be addressed before we add richer metric queries that read from stats views.

WentingWu666666:
Excellent analysis @xgerman! The comparison with CNPG's monitoring model is very helpful. Agreed on:

- Our OTLP support is a genuine advantage
- The static/dynamic config split was designed to support extensibility later
- `customQueries` via ConfigMap refs is the right API shape for user-defined metrics

I've opened a follow-up issue to track this:

 #343  `feat(otel): support user-defined custom metric queries via ConfigMap references`

For the immediate next step, we can also enrich `base_config.yaml` with more built-in queries (backends, replication lag, database size) as a YAML-only change in a subsequent patch.

.github/workflows/test-E2E.yml
WentingWu666666: This is already in a separate script  the workflow step just calls `operator/src/scripts/test-scripts/test-otel-monitoring.sh`. The inline `run:` block is just a thin wrapper (chmod + invoke):

`yaml
chmod +x operator/src/scripts/test-scripts/test-otel-monitoring.sh
operator/src/scripts/test-scripts/test-otel-monitoring.sh
`

All test logic lives in the shell script. Let me know if you'd prefer it as a reusable GitHub Action instead.

guanzhousongmicrosoft (approved):

xgerman (approved):

PR #347, opened 2026-04-20
Allow use of certs for PG connections

alaye-ms:
This PR modifies the PgHBA so that the localhost connections (only within the pod, i.e. gateway to pg connections) are inherently trusted. 

It also changes the replication connection which is used by local and remote replicas to require a cert. This is already used for local HA replicas and so no change is needed. However, for remote replicas the cert would need to be propagated. If provided, we pass in a cert from the user to validate client connections. If none is provided, we will set it to trust all instead. This should change to a default cert being provided upon the creation of our higher level, managing operator.


operator/src/internal/controller/physical_replication.go
```diff
@@ -136,15 +136,40 @@ func (r *DocumentDBReconciler) AddClusterReplicationToClusterSpec(
...
+				Key: "tls.key",
+			}
+			cnpgCluster.Spec.Certificates = &cnpgv1.CertificatesConfiguration{
+				ReplicationTLSSecret: replicationContext.ReplicationTLSSecret,
+				ClientCASecret:       replicationContext.ClientCASecret,
+			}
+		} else {
+			// If we don't have a cert, we just need to trust
+			cnpgCluster.Spec.PostgresConfiguration.PgHBA = []string{
+				"host all postgres localhost trust",
```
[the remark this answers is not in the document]
alaye-ms: Needs to be backwards compatible. Will adjust once there's a central operator

WentingWu666666 (changes requested):
Thanks for tightening the pg_hba defaults -- removing `0.0.0.0/0 trust` is the right move. A few concerns on the cross-cluster cert path before this lands.

### 1. `sslmode=require` doesn't actually verify the primary (replica side is unauthenticated)

`physical_replication.go:151` sets `sslmode=require` on the `ExternalCluster`, and `SSLCert`/`SSLKey` are wired but `SSLRootCert` is not. `require` only encrypts the channel -- it does not check the server's cert against any CA. The result is asymmetric:

- Primary verifies the replica via `cert` auth against `clientCASecret` :white_check_mark:
- Replica does not verify the primary at all :x:

A network attacker on the cross-region/cross-cloud path can impersonate the primary, accept the replica's WAL stream and feedback, and harvest data. The PR description and `setup.md` describe this as "mutual TLS," but it's actually one-way TLS plus client cert auth.

**Fix:** set `sslmode=verify-ca` (or `verify-full` if the cross-cluster hostnames are stable enough) and populate `SSLRootCert` from the same secret that already carries `ca.crt`.

### 2. `ClientCASecret` fallback documentation is incorrect and will mislead users

`setup.md:176`:
> If omitted, the operator falls back to the CA embedded in `replicationTLSSecret`.

Two issues:
- TLS secrets contain `tls.crt` (leaf) and `tls.key`, not a CA. There's nothing to "fall back" to inside `replicationTLSSecret`.
- The actual fallback (when `ClientCASecret==""`) is CNPG's internally-managed CA. That CA cannot validate a user-issued client cert, so omitting `clientCASecret` while supplying `replicationTLSSecret` will break replication.

The playground works only because cert-manager writes `ca.crt` into the same secret and you reuse the secret name for both fields:
```yaml
replicationTLSSecret: cross-region-client-cert
clientCASecret: cross-region-client-cert
```

Please rewrite this row to say both fields are required for user-issued certs, and that you can point both at the same secret if it contains `tls.crt`/`tls.key`/`ca.crt`.

### 3. The `commonName: streaming_replica` requirement is undocumented

Postgres `cert` auth (without an explicit `pg_ident` map) requires the client cert's CN to equal the PG role. The playground gets this right (`commonName: streaming_replica`), but `setup.md` doesn't mention it. A user generating their own cert will hit `FATAL: certificate authentication failed for user "streaming_replica"` with no doc breadcrumb. One sentence in the field table would prevent this.

### 4. Cert rotation works but isn't documented

The `cnpg.io/reload: ""` label on the secret does the heavy lifting (CNPG hot-reloads, cert-manager auto-renews). Worth one short paragraph in `setup.md`:
- For cert-manager users: rotation is automatic if the `cnpg.io/reload: ""` label is present
- For users managing certs manually: update the secret and the label triggers a reload
- Mention what happens at expiry (replication breaks, no specific status condition surfaces today)

### 5. (Lower priority) No validation of the secret at reconcile time

The operator never `Get`s the secret to confirm it exists or has the expected keys; it just writes the name into CNPG's spec. When the secret is missing or malformed, the failure surfaces at the CNPG/pod layer with no DocumentDB-level condition. Consistent with how gateway TLS is handled, so not a blocker, but worth a follow-up issue -- replication misconfiguration is higher-stakes than a one-time setup error.

### Nit
The setup.md callout reads:
> **Insecure by default** -- If no cert is provided, the operator defaults to trusting all external replication connections.

Given (1) above, even the "secure" path isn't fully mutual. Consider also calling out the `verify-ca` plan once #1 is fixed, or flipping the default behavior so cross-region without certs is rejected rather than silently permissive.

docs/operator-public-documentation/preview/multi-region-deployment/overview.md
```diff
@@ -156,7 +156,7 @@ an equal or greater volume of available storage compared to the primary.
 Enable TLS for all connections:
 
 - **Client-to-gateway:** Encrypt application connections (see [TLS configuration](../configuration/tls.md))
-- **Replication traffic:** PostgreSQL SSL for inter-cluster replication
+- **Replication traffic:** Cross-Kubernetes-cluster streaming replication is authenticated with mutual TLS using the `streaming_replica` PostgreSQL role. Configure `spec.clusterReplication.replicationTLSSecret` and optionally `spec.clusterReplication.clientCASecret` to wire up the client certificate and CA. See [Securing replication with TLS](setup.md#securing-replication-with-tls) for the complete setup.
```
WentingWu666666: not mutual TLS

docs/operator-public-documentation/preview/multi-region-deployment/setup.md
```diff
@@ -151,6 +151,47 @@ spec:
       - name: member-westus3-cluster
 ```
 
+#### Securing replication with TLS
+
+Cross-Kubernetes-cluster streaming replication flows over the network between
+member Kubernetes clusters, so the operator secures it with mutual TLS instead
```
WentingWu666666: not mutual TLS

docs/operator-public-documentation/preview/multi-region-deployment/setup.md
```diff
@@ -151,6 +151,47 @@ spec:
...
+configures PostgreSQL to only accepts replication connections over TLS with a valid
+client certificate (`hostssl replication streaming_replica all cert` in `pg_hba.conf`).
+Each member Kubernetes cluster must use the same replication certificate and CA,
+so any replica can authenticate to any primary after a failover. Put the cert into
+a Kubernetes Secret, then pass the name in using the following fields.
+
+| Field | Type | Required | Description |
+| --- | --- | --- | --- |
+| `replicationTLSSecret` | string | Yes for secure multi-region | Name of a Kubernetes Secret that contains the `streaming_replica` client certificate and key. Must contain `tls.crt` and `tls.key`. Must be the same name in every member Kubernetes cluster. |
+| `clientCASecret` | string | Optional | Name of a Kubernetes Secret that contains the CA certificate (`ca.crt`) used to verify the client certificate. If omitted, the operator falls back to the CA embedded in `replicationTLSSecret`. Must be the same name in every member Kubernetes cluster. |
```
WentingWu666666: AI traced through the CNPG source and believe the actual fallback is different. In [cluster_funcs.go:654-659](https://github.com/cloudnative-pg/cloudnative-pg/blob/v1.28.1/api/v1/cluster_funcs.go#L654-L659):
```
 func (cluster *Cluster) GetClientCASecretName() string {
     if cluster.Spec.Certificates != nil && cluster.Spec.Certificates.ClientCASecret != "" {
         return cluster.Spec.Certificates.ClientCASecret
     }
     return fmt.Sprintf("%v%v", cluster.Name, ClientCaSecretSuffix)  // → "{cluster-name}-ca"
 }
```
When `ClientCASecret` is empty, CNPG falls back to `{cluster-name}-ca` — its auto-generated internal cluster CA, not the CA from `replicationTLSSecret`.

This matters for cross-cluster replication because each CNPG cluster generates its own independent CA. If `clientCASecret` is omitted:
 - Cluster-A trusts certs signed by Cluster-A's CA
 - Cluster-B presents a cert signed by Cluster-B's CA
 - → replication auth would fail silently

Could you double-check this? If confirmed, the doc should probably say something like:

 "If omitted, CNPG falls back to the cluster's own internal CA ({cluster-name}-ca), which will not work for cross-cluster replication since each cluster has a different CA. For multi-region setups, this field is effectively required."

WentingWu666666 (approved):

docs/operator-public-documentation/preview/multi-region-deployment/setup.md
alaye-ms: Confirmed and corrected

xgerman (changes requested):
## Documentation review

Applied the project's `.github/agents/documentation-agent.md` rubric (Microsoft Writing Style Guide + MkDocs rules + "never use 'cluster' alone" disambiguation) to the doc surface touched by this PR: the new `setup.md` section, the rewritten `overview.md` bullet, the godoc that drives `api-reference.md`, and the playground YAML it links to.

### Summary

The structure is right (concept → field table → cross-reference → playground link) and the cross-link layering follows the rubric's "playground = scripts, public docs = concepts" rule. But the new prose ships with a broken Markdown table, an apostrophe typo, and a subject-verb agreement bug in the security claim itself; the godoc violates the "never use 'cluster' alone" rule (and it renders straight into the published API reference); and the "Insecure by default" framing buries what is functionally a security regression in a single soft `!!! important` line with no terminal period.

### 🔴 Required before merge

1. **Broken Markdown table** in `setup.md` field reference — 3 header cells, 4 separator cells:
   ```
   | Field | Type | Description |
   | --- | --- | --- | --- |
   ```
   Drop the trailing `| ---`. Likely fails `mkdocs build --strict`.

2. **Apostrophe typo** in the `clientCASecret` row: *"the CNPG operator falls back to **it's** own generated CA"* — should be `its` (possessive).

3. **Subject-verb agreement bug in the security claim** (paragraph above the field table):
   > "configures PostgreSQL to **only accepts** replication connections over TLS"
   
   `to + bare infinitive` ⇒ `only accept`. Modifier placement is also weak; preferred: *"configures PostgreSQL to accept replication connections only over TLS with a valid client certificate"*. This sentence is *the* security claim the page makes — it needs to read cleanly.

4. **No guidance for certificate rotation.** When the Secret rotates, CNPG hot-reloads server-side material only if the Secret carries `cnpg.io/reload: ""` (the playground sample does, but the doc never says so), and the `SSLCert`/`SSLKey` projected into external clusters may not reload until the pods restart. Production users will hit this. Add a short "Certificate rotation" subsection.

### 🟠 Major

5. **Strengthen the "Insecure by default" admonition.** Promote to `!!! warning` (or `!!! danger`), terminate the sentence with a period, replace informal "cert" with "certificate", and make it prescriptive:
   > **Always set `replicationTLSSecret` and `clientCASecret` for multi-Kubernetes-cluster replication exposed to any untrusted network.** When both fields are unset, the operator falls back to trust-based authentication on the replication endpoint — anyone who can reach the PostgreSQL service on the primary can act as a replica.

6. **Field reference table is missing Required and Default columns.** The rubric explicitly calls for `Field, Type, Required, Default, Description`. This matters here because `clientCASecret` is `+optional` in the API but *functionally required* once `replicationTLSSecret` is set (the doc itself says replication will fail otherwise). Surface that landmine in the table.

7. **`replicationTLSSecret` description omits half its job.** The doc and godoc both describe only the *client* role. The same Secret is also consumed as the *server* cert by CNPG's `Certificates.ReplicationTLSSecret`. The dual role is what forces the constraint that the cert's CN must equal `streaming_replica` — and the playground YAML silently relies on this (`commonName: streaming_replica`). Add one sentence: *"The certificate's Common Name (CN) must equal `streaming_replica`, because PostgreSQL `cert` authentication maps CN → role."*

8. **Godoc violates the "never use 'cluster' alone" rule** (`documentdb_types.go`). Both new fields say *"the operator references this secret in **clusters** participating in replication"* and *"NOTE: It needs to be the same for all **clusters**"*. These render straight into `api-reference.md` (which the PR correctly regenerated). Qualify as "Kubernetes clusters" or "member Kubernetes clusters" and regenerate.

9. **Overview bullet is inconsistent with its siblings.** The new bullet is now ~5× the length of the surrounding one-liners; overview pages are supposed to be scannable. Tighten to:
   > **Replication traffic:** Cross-Kubernetes-cluster streaming replication uses TLS with the `streaming_replica` role. See [Securing replication with TLS](setup.md#securing-replication-with-tls).

10. **`ClientCASecret` `+optional` contradicts documented behavior.** The godoc says optional; the doc says replication fails without it. Preferred fix: add a kubebuilder CEL `XValidation` so the API enforces it:
    ```
    +kubebuilder:validation:XValidation:rule="!has(self.replicationTLSSecret) || has(self.clientCASecret)",message="clientCASecret is required when replicationTLSSecret is set"
    ```
    Landmines hidden in `+optional` aren't a GA-worthy story.

### 🟡 Minor

11. **Terminology drift in `setup.md`** — uses both "replication cert" (informal) and "replication certificate" (formal) within two paragraphs. Pick one — the formal version reads better in published docs.

12. **Self-signed playground example without a warning.** The playground YAML uses `selfSigned: {}`. That works for PG cert-auth (which validates by CN, not chain) but sets a poor production precedent. Setup.md should say: *"This example uses a self-signed certificate for brevity. In production, issue the certificate from a managed CA (Vault, cert-manager `CAIssuer`, ACME)."*

13. **Pattern regex rejects dotted Secret names.** `^[a-z0-9]([-a-z0-9]*[a-z0-9])?$` rejects K8s-valid dotted names (e.g., `my.cert.secret`). RFC 1123 DNS-subdomain allows dots. Either widen the regex or document the constraint.

14. **Anchor verification not confirmed.** The new `setup.md#securing-replication-with-tls` anchor is referenced from `overview.md`. Worth running `cd docs && mkdocs build --strict` once to confirm both the anchor and the malformed table (issue #1) are caught.

15. **Front matter not added.** The two doc pages lack the YAML front matter the rubric recommends (`title`, `description`, `tags`) for AI-friendliness. Pre-existing, but easy to add while you're editing the file.

### 🟢 Nitpicks

16. **Inline `NOTE:` in godoc renders awkwardly in the API ref.** The rendered `api-reference.md` shows `NOTE: It needs to be the same for all clusters` mid-cell. Material admonitions don't render inside table cells. Either drop "NOTE:" or move the constraint into prose docs.

### ✅ Positive feedback

- Cross-references are well placed: overview → setup → playground YAML → API reference. That's exactly the layered structure the rubric calls for.
- Playground YAML is referenced via absolute GitHub URL (not a relative path that escapes `docs_dir`) — the rule the rubric explicitly calls out, gotten right.
- `setup.md` section follows clear concept → field table → cross-references → admonitions flow. Copy-edit and the strengthened warning bring it to merge-ready.
- `api-reference.md` was correctly regenerated rather than hand-edited.
- Sentence-case headings (`#### Securing replication with TLS`) follow the rule. ✅

### Required-before-merge shortlist

1. Fix the table separator (4 cells → 3).
2. `it's` → `its`.
3. `only accepts` → `only accept` (and ideally rephrase the whole clause).
4. Add a Certificate rotation note (or admonition).
5. Strengthen the "Insecure by default" admonition to a `!!! warning` with prescriptive language.
6. Qualify "cluster" → "Kubernetes cluster" in both new godoc blocks (and regenerate `api-reference.md`).
7. Either add CEL `XValidation` enforcing `clientCASecret` when `replicationTLSSecret` is set, or rewrite the optional/required markers to match documented behavior.

`/request-changes` on docs.

---

<sub><em>Posted via the project's documentation-agent rubric (`.github/agents/documentation-agent.md`). Code-quality and security findings will follow in a separate review.</em></sub>

xgerman (changes requested):
## Code-quality & security review

Applied the project's `.github/agents/code-review-agent.md` rubric to the Go changes in this PR (controllers, CNPG patch ops, replication context, tests). The doc-side review is already posted as a separate review.

### Summary

The PR's central intent — tightening the base PgHBA and adding an opt-in TLS path for replication — is the right direction, and the patch-op test for `/spec/certificates` is exactly the kind of contract-level test this code needs. But two things give me pause: (a) the JSON patch op uses `replace` on a path that may not exist (the exact pattern this file already fixed for `/metadata/annotations`), and (b) the "no cert" branch of the new code silently sets `host replication streaming_replica all trust` on multi-region clusters — that's a security regression versus what the PR title implies, with no warning, status condition, or event surfacing it to the operator. There are also several drive-by formatting changes from unrelated files mixed into the diff that should be split out.

`/request-changes` on code (in addition to the doc review).

### 🔴 Critical

#### C1. JSON Patch `replace` on `/spec/certificates` will fail when the field is absent
`operator/src/internal/controller/physical_replication.go` `getReplicasChangePatchOps`:

```go
*patchOps = append(*patchOps, cnpg.JSONPatch{
    Op:    cnpg.PatchOpReplace,
    Path:  cnpg.PatchPathCertificates,
    Value: desired.Spec.Certificates,
})
```

RFC 6902 §4.3: *"The target location MUST exist for the operation to be successful."* For a CNPG cluster that has never had TLS configured (i.e., the original state for every existing deployment in the wild), `spec.certificates` is absent and this `replace` will reject the entire atomic patch. That's the exact failure mode this file already learned about for `/metadata/annotations` — see `cnpg_sync.go:204-219` where the code explicitly branches on `current.Annotations == nil` and uses `add` instead. The cert patch needs the same gate.

Suggested fix:
```go
if current.Spec.Certificates == nil && desired.Spec.Certificates == nil {
    // no-op: nothing to patch
} else if current.Spec.Certificates == nil {
    *patchOps = append(*patchOps, cnpg.JSONPatch{
        Op:    cnpg.PatchOpAdd,
        Path:  cnpg.PatchPathCertificates,
        Value: desired.Spec.Certificates,
    })
} else {
    *patchOps = append(*patchOps, cnpg.JSONPatch{
        Op:    cnpg.PatchOpReplace,
        Path:  cnpg.PatchPathCertificates,
        Value: desired.Spec.Certificates,
    })
}
```

This signature change means `getReplicasChangePatchOps` needs `current` as well as `desired`. Worth doing — the bug as written breaks any cluster that hasn't already had TLS configured.

**`[Required]`** — please verify locally against a CNPG cluster that has never had `spec.certificates` set, then either fix the bug or post the test/run that proves it isn't one.

#### C2. Trust-all replication fallback on the multi-region path is a silent security regression
`physical_replication.go` `AddClusterReplicationToClusterSpec`:

```go
} else {
    // If we don't have a cert AND we're multi-regional, we just need to trust (for now)
    cnpgCluster.Spec.PostgresConfiguration.PgHBA = []string{
        "host all all localhost trust",
        "host replication streaming_replica all trust",
    }
}
```

This means: any DocumentDB resource that enables `clusterReplication` (i.e., has a `clusterList`) without setting `replicationTLSSecret` gets PgHBA rewritten to accept replication connections from **any source IP** as `streaming_replica` with **no authentication**. Combined with the PR's removal of `host all all 0.0.0.0/0 trust` from the base spec, the result is a multi-region cluster whose replication endpoint is wide open to anyone who can route to the PostgreSQL service IP across Kubernetes clusters — which, in a fleet/multi-cloud setup, is everyone on the inter-cluster network. The PR title is "Allow use of certs for PG connections" but the code path is "default to insecure replication when no cert is set."

Three changes needed before merge, in priority order:

1. **Fail loud, not silent.** When TLS is off but `clusterReplication` is enabled with more than one member, set a `Status.Condition` of type `ReplicationSecured` with `Status=False`, `Reason=TrustModeFallback`, and a message pointing at `replicationTLSSecret`. Emit a `Warning` event with the same payload. Right now nothing surfaces this to `kubectl describe`.
2. **Refuse it by default behind a feature gate.** Add a `featureGate: AllowInsecureClusterReplication` (default `false`) and have `AddClusterReplicationToClusterSpec` return an error if the user enabled multi-region replication without TLS and without the feature gate. That preserves "fail-safe" semantics and gives sophisticated users an explicit escape hatch.
3. **Tighten the `streaming_replica all trust` line.** Even in the explicit-opt-in case, prefer constraining to the inter-cluster service CIDR (`samenet`, or the explicit service IP set the operator already enumerates) rather than literal `all`. The operator already knows the remote service names — it can derive their addresses.

**`[Required]`** — at minimum #1 (loud status + event); ideally all three.

### 🟠 Major

#### M1. PgHBA whole-array overwrite is fragile
The fallback branch does `cnpgCluster.Spec.PostgresConfiguration.PgHBA = []string{...}`, dropping anything `GetCnpgClusterSpec` set. Today both lists happen to start with `host all all localhost trust`, but the moment someone adds a third rule to the base spec (e.g., for a new cron user, monitoring scrape, change-stream service), the multi-region-no-TLS branch silently drops it and you get a CVE-shaped surprise. Use `append` (and dedupe), or factor a single composer function that takes both inputs.

**`[Required]`** — even if you don't fix the trust-all fallback today, this coupling needs to go.

#### M2. Missing test: TLS-on path preserves base PgHBA
The new test suite covers:
- `Certificates == nil` and fallback PgHBA applied  ✅
- `Certificates.ClientCASecret` propagated when TLS set  ✅
- TLS not enabled when only `ClientCASecret` is set without `ReplicationTLSSecret`  ✅

But there's no test asserting that when `ReplicationTLSSecret` IS set, `cnpgCluster.Spec.PostgresConfiguration.PgHBA` is NOT clobbered (i.e., still contains the base `hostssl replication streaming_replica all cert` rule from `GetCnpgClusterSpec`). That's the test that would have caught M1 above as a regression. Add it.

**`[Required]`** per `/needs-tests`.

#### M3. Missing test: TLS-on path populates SSL fields on remote externals
The first test asserts `sslmode`/`SSLCert`/`SSLKey` are absent on remote externals when TLS is off, but no symmetric test asserts they ARE present (and contain the expected secret-key-selector shape) when TLS is on. The TLS-on test only checks `Spec.Certificates`. Two lines, big coverage gain.

**`[Suggestion]`** — would block on M2 but happy to call this `[Suggestion]`.

#### M4. No CEL `XValidation` enforcing `clientCASecret` when `replicationTLSSecret` is set
Already covered in the doc review (point 10). Restating here as a code concern: this is a configuration landmine that the API layer can prevent. Add:

```go
// +kubebuilder:validation:XValidation:rule="!has(self.replicationTLSSecret) || has(self.clientCASecret)",message="clientCASecret is required when replicationTLSSecret is set"
type ClusterReplication struct { ... }
```

(Heads-up — per the repo memory, Go 1.26 `gofmt` rewrites `''` to `''` in CEL rules; if you write the rule with empty-string comparisons use `size(...) > 0` instead. The version above uses `has()` which is safe.)

**`[Required]`** — same reasoning as the doc review: GA stories shouldn't include "this field is optional but if you don't set it nothing works."

#### M5. `Pattern` regex narrower than the K8s Secret-name spec
`^[a-z0-9]([-a-z0-9]*[a-z0-9])?$` rejects dotted names. K8s Secret names follow RFC 1123 DNS subdomain, which permits dotted segments. Either widen to the standard pattern or document the constraint. Cert-manager users sometimes use dotted names; this would reject otherwise-valid secrets.

**`[Suggestion]`**

#### M6. The "self" external-cluster keeps `user: postgres` while all others switch to `streaming_replica`
The test asserts this on purpose:
```go
if ec.Name == replicationContext.CNPGClusterName {
    // Self cluster still uses the superuser for self-loopback.
    Expect(ec.ConnectionParameters["user"]).To(Equal("postgres"))
    continue
}
```

That works *because* the new base PgHBA still has `host all all localhost trust`, so a `postgres` connection on the local pod is auth-free. Two concerns:
- This is undocumented in the production code (only the test explains it). Add a comment in `AddClusterReplicationToClusterSpec` where the self-cluster is built (`physical_replication.go:127-137`).
- If anyone tightens the local-loopback line in the future (or moves the self-cluster's connection to a non-localhost service host), this silently breaks self-loopback. Worth a defensive constant somewhere, or at minimum a test that fails loudly if the localhost trust line is ever removed.

**`[Suggestion]`**

### 🟡 Minor

#### m1. Drive-by formatting churn in unrelated files
The diff includes pure-whitespace changes in:
- `operator/src/internal/utils/constants.go` (`=` realignment for `DEFAULT_DOCUMENTDB_IMAGE` and `DEFAULT_OTEL_COLLECTOR_IMAGE`)
- `operator/src/internal/utils/pv_recovery.go` (trailing whitespace + `=` realignment)
- `operator/src/internal/cnpg/cnpg_sync.go` (reformat of a struct literal that compiles identically)
- `operator/src/internal/controller/documentdb_controller.go` (single blank-line removal at L1019)

None of these relate to "allow use of certs for PG connections." They pollute the diff, complicate `git blame`, and risk merge conflicts with concurrent work. Split into a separate "chore: gofmt drive-by" PR (or revert from this one).

#### m2. `deploy-fleet-bicep.sh` `HUB_CERTIFICATE_AUTHORITY` change is unrelated
The 5-line shell change adds `HUB_CERTIFICATE_AUTHORITY` env var via `kubectl set env` with a comment "temp fix while joinMC.sh is out of date." This is a Fleet-bootstrap concern, not a PG-cert concern, and even ships with a "temp" tag. Belongs in its own PR with its own description of what's broken in `joinMC.sh` and what tracking issue this temp fix references.

#### m3. `ReplicationContext` fields broke the access pattern
The struct mixes exported state (`CNPGClusterName`, `Environment`, ...) with private state accessed via methods (`state`, `targetLocalPrimary`, `currentLocalPrimary`). The two new fields chose exported. Consistent enough with sibling fields, but if there's an unwritten convention that runtime/replication-decision state goes private with accessors, these should too. Worth a 30-second chat with whoever set up this struct.

**`[Question]`**

### 🟢 Nitpicks

#### n1. `// (for now)` comment in production code
```go
// If we don't have a cert AND we're multi-regional, we just need to trust (for now)
```

If "for now" is real, file an issue and replace with `// TODO(#NNN): require TLS for multi-region replication once the higher-level managing operator provisions a default cert (see PR #347 discussion).` If "for now" is permanent, drop the parenthetical.

#### n2. PR title vs scope
"Allow use of certs for PG connections" undersells the diff. The PR also tightens the base PgHBA (a security improvement worth calling out), switches replication user from `all` to `streaming_replica`, and introduces a multi-region fallback path with new defaults. Consider splitting or expanding the title — and yes, this matters for the squash-merge commit message that future readers will see in `git log`.

### ❓ Questions

#### Q1. How does this PR interact with single-region deployments that already exist?
The base PgHBA change (`host all all 0.0.0.0/0 trust` → `host all all localhost trust`) is a behavioral change for *every existing single-region cluster*, not just multi-region. Anything that connects to PostgreSQL via the service IP rather than 127.0.0.1 will start failing after the operator upgrades the CNPG spec. The gateway sidecar should be fine (in-pod loopback) — but what about:

- Any user app the customer has running against the cluster's RW service?
- The CNPG operator's own probe connections (do they use localhost or the service IP)?
- Monitoring scrapes (Prometheus exporters, OTEL)?
- The kubectl plugin's `status`/`health` commands?

If any of these depend on the IP-based trust line, this PR breaks them on upgrade with zero migration story. Please confirm — ideally with the E2E suite run against an upgrade path (`upgrade` label is already in the suite).

#### Q2. Does `hostssl replication streaming_replica all cert` work with CNPG's auto-provisioned `streaming_replica` cert in single-region deployments?
CNPG normally generates a per-cluster `streaming_replica` cert for HA replicas. The new base PgHBA assumes that cert is signed by a CA the operator trusts. When `replicationTLSSecret` is empty in a *single-region* cluster, does CNPG's default cert satisfy the `cert` auth method? The cnpg_cluster_test count change to 2 suggests it should, but I'd want this proven by an HA-cluster E2E run (`scale` label).

### ✅ Positive feedback

- The `getReplicasChangePatchOps` cert tests are exactly the right shape — they assert the *patch-op contract* rather than the resulting cluster spec, which makes them resilient to refactors. Nice work.
- The `util.None` networking-strategy choice for the unit tests is sharp — avoids pulling in Istio/Fleet plumbing.
- Tightening the base PgHBA (`0.0.0.0/0` → `localhost`) on the single-region path is a genuine security improvement, well overdue. (Just needs the upgrade story in Q1 worked out.)
- `cnpg_cluster_test.go` was updated to reflect the new PgHBA rule count (3 → 2) — that's the kind of small but careful test maintenance that prevents future regressions.
- The patch-op `nil`-value test (`emits a replace patch for spec.certificates with a nil value when TLS is disabled`) explicitly documents that the rollback path is intended to be `replace → nil`, not "skip the patch." That's well-thought-out intent — once the C1 bug is fixed, the test still encodes the right design.
- Moving the replication user from the catch-all `all` to the dedicated `streaming_replica` role is the right call regardless of TLS — better least-privilege.

### Required-before-merge shortlist

1. Fix the `replace`-on-absent-path bug for `/spec/certificates` (C1).
2. Make the trust-all replication fallback fail loud — status condition + warning event at minimum (C2 #1).
3. Add CEL `XValidation` requiring `clientCASecret` when `replicationTLSSecret` is set (M4).
4. Add the missing TLS-on PgHBA test (M2) and ideally the TLS-on SSL-fields test (M3).
5. Replace the whole-array PgHBA overwrite with an `append`/composer (M1).
6. Answer Q1 — what's the upgrade story for existing single-region clusters?
7. Split out the unrelated drive-by formatting and the `HUB_CERTIFICATE_AUTHORITY` change (m1, m2).

`/request-changes` `/needs-tests`

---

<sub><em>Posted via the project's code-review-agent rubric (`.github/agents/code-review-agent.md`). Doc-side review posted separately above.</em></sub>

alaye-ms:
Will re-open when mTLS is enabled

PR #348, opened 2026-04-20
feat: long-haul (canary) test driver for the operator (#220)

### LINKED ISSUE #220: Set up long haul test cluster
Set up a long hauil test cluster:
1) Constantly writes/reads - ensure no data is lost
2) Perform constant managment operations like add region, remove region; local HA, not local HA; scale up/down, backup/restiore, etc.

In particular we need to test updates.

WentingWu666666:
Closes #220.

## What this PR adds

A standalone **long-haul (canary) test driver** for the DocumentDB Kubernetes Operator, plus the GitHub Actions plumbing to keep it running on a long-lived AKS cluster.

The driver lives under `test/longhaul/` and is its own Go module. It is **not** a Ginkgo suite — it is a long-running binary (`cmd/longhaul`) deployed as a `Deployment` into a namespace alongside a real DocumentDB CR. It exercises the operator with a continuous, realistic workload while a weighted-random scheduler injects disruptive operations, and it journals every event so failures can be reproduced.

### Pieces that ship in this PR

- **Continuous data plane** (`workload/`): writer + verifier goroutines, sequence / checksum gap detection, lag-aware reads — the durability oracle for FM1/FM3/FM7.
- **Operation scheduler** (`operations/`): weighted random selection, per-category cooldowns, steady-state gating, and an outage-policy verdict per operation. Operations implemented today: `ScaleUp`, `ScaleDown`, `UpgradeDocumentDB` (auto-skips when `instancesPerNode < 2` because rolling restart on a single instance is unavoidable downtime).
- **Real cluster client** (`monitor/k8sclient.go`): reads the DocumentDB CR, watches pods, pulls metrics-server samples — no fakes.
- **Health monitor + leak detector** (`monitor/health.go`, `monitor/leakdetect.go`): steady-state detection plus linear-regression slope on memory/CPU samples.
- **Append-only journal** (`journal/`): every disruption window, every workload event, every verdict. The disruption-window oracle is what makes verdicts reproducible.
- **Reporter + alerter** (`report/`): writes the `longhaul-report` ConfigMap so a human can `kubectl get cm longhaul-report -o yaml` for a single source of truth on pass/fail. Alerts surface in the monitor workflow.
- **In-cluster packaging** (`Dockerfile`, `deploy/longhaul.yaml`): runs as a `Deployment` (not a `Job`) — if the process exits, it restarts; the ConfigMap is the durable verdict, not the pod's exit code.
- **Two GHA workflows chained via `workflow_run`**:
  - `LONGHAUL - Build Test Driver Image` — builds and pushes to GHCR.
  - `LONGHAUL - Deploy Test Driver to AKS` — uses a long-lived ServiceAccount-token kubeconfig (`LONGHAUL_KUBECONFIG` repo secret) to roll the Deployment.
- **`LONGHAUL - Monitor`** — hourly cron that pulls the report ConfigMap and posts a workflow summary; alerts on stale or failing reports.
- **Tier-1 unit tests** across `config`, `journal`, `monitor`, `report`, `operations` — no cluster needed, run with `go test ./...` from `test/longhaul/`.
- **Docs** — `docs/designs/long-haul-test-design.md` (problem statement, eight failure modes, design principles, architecture, operations catalog with status, env-var table, phase tracker) and `test/longhaul/README.md` (how to run, what every env var does, why `Deployment` over `Job`).

## Why this is structured the way it is

- **Long-haul is what happens when the failure modes you need to detect require more time and operations than fit in any bounded test run.** Memory leaks, CR-history drift, upgrade-under-state failures — these don't speed up. The eight failure modes in the design doc are why every component exists.
- **No drain before disruption.** Workload runs through scale, upgrade, and (future) failover ops — that's exactly the bug class we're trying to find.
- **`Deployment`, not `Job`.** A `Job` that finishes is a `Job` no one looks at. The `longhaul-report` ConfigMap is the durable verdict, and the monitor cron makes sure a human sees stale runs.
- **Single cluster today, two clusters later.** A baseline DocumentDB cluster (control group) is in the design but not in this PR — the single-cluster driver already exercises the primary failure modes.

## Live canary

I've stood up a long-running cluster — the driver Deployment lives there and gets re-rolled by the deploy workflow on every successful image build:

- AKS cluster: [longhaul-aks (longhaul-rg, sub 81901d5e…)](https://ms.portal.azure.com/#@microsoft.onmicrosoft.com/resource/subscriptions/81901d5e-31aa-46c5-b61a-537dbd5df1e7/resourceGroups/longhaul-rg/providers/Microsoft.ContainerService/managedClusters/longhaul-aks/workloads)
- Everything (DocumentDB CR + driver Deployment + `longhaul-report` ConfigMap) runs in `documentdb-test-ns`.
- Useful pokes:
  - `kubectl logs deploy/longhaul-test -n documentdb-test-ns`
  - `kubectl get cm longhaul-report -n documentdb-test-ns -o yaml`

## What's intentionally out of scope

Tracked in the design doc's phase table:

- **Two-cluster topology** (Phase 3): primary + baseline, so signals become attributable.
- **Backup / restore operations** (Phase 2a) and **chaos primitives beyond scale** (Phase 2d): drain node, kill primary pod, operator restart.
- **Operator version upgrade operation**: we exercise DocumentDB version upgrades today; operator upgrades will land once the release workflow is wired to bump the canary's target.
- **`OutagePolicy.AllowedDowntime` enforcement**: the field is reserved on the struct; current verdicts use `AllowedWriteFailures` + `MustRecoverWithin`. Called out explicitly in the design doc.

## How to review

If you only want to skim, three files give you the shape of the PR:

1. `docs/designs/long-haul-test-design.md` — the why.
2. `test/longhaul/cmd/longhaul/main.go` — the wiring.
3. `test/longhaul/operations/scheduler.go` — the heart of the disruption loop.

If you want to run it locally:

```bash
cd test/longhaul
LONGHAUL_CLUSTER_NAME=documentdb-sample LONGHAUL_NAMESPACE=documentdb-test-ns \
  LONGHAUL_MAX_DURATION=10m \
  go run ./cmd/longhaul
```

Happy to walk through any area on a call.


docs/designs/long-haul-test-design.md
```diff
@@ -0,0 +1,439 @@
...
+**Steady State Definition:**
+```
+- All pods in Ready state
+- DocumentDB CR conditions: all True
+- Replication lag < threshold (if replicated)
+- No new pod restarts in last 5 min
+- Workload success rate > 99.9%
+- No unresolved backup failures
+```
+
```
xgerman, line 145: we should probably hook this up to a monitoring system... do we need alerts, etc.?

docs/designs/long-haul-test-design.md
```diff
@@ -0,0 +1,439 @@
...
+- Cluster state snapshot (topology, pod count, primary node)
+- Associated errors (if any)
+
+**Key use case:** When a write failure occurs, the journal shows whether it happened during an expected disruption window (tolerable) or during steady state (bug).
+
+---
+
+## Canary Model
+
+The long-haul test is a **single persistent canary** running on a dedicated AKS cluster. Existing Kind-based integration tests (45-60 min, PR-gated) already cover short-lived validation — there is no need for a separate smoke mode.
```
xgerman: do we need to consider canary's in different regions - how do we deal with regional outages?

xgerman (changes requested):
## Review

**Verdict:** request-changes — lightly. No blocking issues for the skeleton itself; the skip gate is airtight and the config tests are appropriate. But there are four structural items the Phase 1a framing bakes in that will be expensive to unwind in Phase 1b/1c if not settled now.

**Verified first:** `test-unit.yml` uses `operator/src` as working-dir and runs `go list ./...`, which enumerates the new packages. With `LONGHAUL_ENABLED` unset the `BeforeSuite` calls `Skip()` and the suite registers as a single "skipped" line. Zero CI overhead. ✅

---

### 🟠 Major

**M1. Module placement will pollute `operator/src/go.mod`** — `operator/src/test/longhaul/` (entire tree)

The existing `test/e2e/` is a **separate Go module** at `test/e2e/go.mod` — a deliberate isolation so `mongo-driver/v2`, CNPG test utilities, and other test-only deps never enter the operator's dependency graph. This PR puts longhaul inside the operator module.

For Phase 1a (stdlib + Ginkgo/Gomega, already in the operator `go.mod`) this is harmless. For the phases *this PR's design* describes:

- **Phase 1b** adds `go.mongodb.org/mongo-driver` (design line 75)
- **Phase 1c** adds a file-backed journal
- **Phase 1d** adds OTel / Prometheus client libs
- **Phase 2f** adds k8s Job manifests + likely controller-runtime client

…all would land in `operator/src/go.mod` and be forced onto every operator build. **Resolve before the first non-stdlib dep is introduced.** Preferred: move to `test/longhaul/` with its own `go.mod` (mirrors e2e). Alternative: nest a `go.mod` under `operator/src/test/longhaul/` and document the convention in `CONTRIBUTING.md`.

**M2. Design does not engage with the existing e2e suite** — `docs/designs/long-haul-test-design.md` throughout

Neither the PR description nor the design doc reference `docs/designs/e2e-test-suite.md` or `test/e2e/`. That suite already ships primitives long-haul will need:

- Ginkgo v2 harness with `SynchronizedBeforeSuite` for cluster bootstrapping (`test/e2e/pkg/testenv`)
- Label taxonomy (`test/e2e/labels.go`)
- Run-ID tagging for artifact separation (`test/e2e/runid.go`)
- envsubst-based manifest rendering
- A mongo-driver v2 wire-protocol client (same driver the design proposes)
- `.github/actions/setup-test-environment` composite action for cluster provisioning

The "Directory Structure" section (lines 213–244) proposes reinventing `monitor/`, `config/`, suite bootstrap, and mongo client from scratch. **The design should commit to either reuse — most naturally by making longhaul a sibling module that depends on `test/e2e/pkg/` — or explicitly enumerate what it forks and why.** Otherwise, six months from now we'll have two drifting implementations of "connect to a DocumentDB CR and open a mongo client" with different flag names, different retry policies, and different label conventions.

**M3. Four design gaps that will cost rework, not captured in §Open Questions** — `long-haul-test-design.md:443-445`

1. **Event journal durability across process/pod restarts.** Line 239 says "In-memory + file-backed" but never says where the file lives. If the canary runs as a Kubernetes Job (line 305), the pod's ephemeral filesystem disappears on restart — and **Phase 2e explicitly plans for operator restart / pod eviction chaos**. The journal is the only post-mortem artifact and it will be destroyed by the very events it is supposed to record. **Commit to a PVC mount or an external sink (object storage, Loki) in the design.**
2. **Workload resumption idempotency.** Writers use a unique index on `(writer_id, seq)` (line 80). If a writer or the Job pod restarts, where does `seq` restart from? Reset to 0 → writes collide. Bootstrap from `max(seq)` per `writer_id` → the oracle must tolerate a gap between crash and resume without flagging data loss. Neither path is described.
3. **Teardown on abort.** "On failure: collect artifacts, preserve cluster state" (line 188) handles the fatal path. But Ctrl-C, CI cancellation, or Job preemption mid-run leaves writer data in the target DB, possibly an in-progress restore cluster (Phase 2a), possibly a half-scaled topology (Phase 1e). No cleanup hooks, resumable state files, or leftover-run detector.
4. **Latency-regression baseline policy.** Failure tiers list "latency regression" as a warning (line 197), but no baseline/threshold policy is defined. **List this as an Open Question at minimum** so Phase 1b doesn't invent a threshold policy in a PR review thread.

These shape the Phase 1b/1c interfaces and should be addressed (or explicitly deferred in §Open Questions) before the design doc merges.

**M4. `MaxDuration=30m` default contradicts "run until failure"** — `config.go:594`, `long-haul-test-design.md:37,188`

The design repeatedly commits to the Strimzi run-until-failure canary model. The Phase 1a default is 30m; the env doc (line 259) says `0s` opts into infinite. Fine for local development — exactly wrong for the Phase 2f canary Job. **Pick one:** either (a) document that the canary Job manifest will set `LONGHAUL_MAX_DURATION=0s` explicitly and mention this in the README "CI Safety" section, or (b) flip the default to `0` and require local-dev to pass `LONGHAUL_MAX_DURATION=10m`.

---

### 🟡 Minor (cheap in-PR fixes)

- **m1** `IsEnabled()` — `config.go:637-640` — doesn't trim whitespace. `LONGHAUL_ENABLED=" true "` → false (silent skip). Add `strings.TrimSpace`. **Silent-failure mode; worth fixing in-PR.**
- **m2** `Validate()` — `config.go:625-633` — accepts negative `MaxDuration`. `time.ParseDuration("-5m")` succeeds. Add `if c.MaxDuration < 0 { return fmt.Errorf(...) }`.
- **m3** Error strings capitalized — `config.go:627,630` — `ST1005` / Go convention: lowercase. Matching tests use `ContainSubstring("Namespace")` / `"ClusterName"` and will need adjustment.
- **m4** Path inconsistency — `long-haul-test-design.md:340` says `kubectl apply -f test/longhaul/deploy/job.yaml` but the directory structure roots at `operator/src/test/longhaul/`. Fix path, or resolve M1 first.
- **m5** README overstates CI safety — `README.md:543-547` — phrase as "no CI workflow currently sets `LONGHAUL_ENABLED`; do not set it in any PR-gated workflow" rather than as a property of the code.
- **m6** `config` is a very generic package name. If Phase 2 ever imports the operator's own config, it will shadow. Consider `longhaulcfg` or `cfg`. Not blocking.
- **m7** Missing `IsEnabled` test cases — `config_test.go:727-757` — given the silent-skip failure mode, lock it down: leading/trailing whitespace, mixed case (`"True"`, `"YES"`), `"0"` / `"no"` / `"false"` → false.

### 🟢 Nit
- `longhaul_test.go:799` — package-level mutable `testConfig`. Fine for a single-file suite; revisit if Phase 1b splits files.
- `DefaultConfig()` by value vs. `Validate()` pointer receiver — slight asymmetry. Leave it.
- README line 509's `-timeout 0` guidance is correct and rarely documented — nice touch.

---

### Positive feedback

- Skip gate is correctly placed in `BeforeSuite`, **not** `BeforeEach` — single skipped line instead of N per spec. Exactly right.
- `LoadFromEnv` cleanly separates parsing from validation, so `Validate()` can be called on a programmatically-built `Config` in Phase 1b unit tests without fighting env vars.
- The design doc's survey of Strimzi / CloudNative-PG / CockroachDB / Vitess and the "adopt / skip" table (lines 354–361) is unusually high-quality framing for an OSS design doc.
- Phased breakdown (Phase 1a → 3) is appropriately granular — each phase is demoable and reviewable.
- Test coverage on the config package is thorough for the surface area.

---

### Questions for the author

1. **(M1)** Any reason longhaul can't live at `test/longhaul/` as a sibling module to `test/e2e/`? If there is one, please capture it in the design.
2. **(M2)** Have you looked at `test/e2e/pkg/` for reusable primitives (typed clients, envsubst, runid, labels)? What's the plan for avoiding a second mongo-client implementation?
3. **(M3.1)** For the journal: PVC, external object store, or stdout-scraped-by-Loki?
4. **Issue #220** text isn't quoted in the PR — are there SLO targets, acceptance criteria, or a required timeframe (e.g., 72h)? Restating them in the design's Problem Statement would make future reviews easier.
5. Should `CONTRIBUTING.md` or `AGENTS.md` get a one-liner pointing at `operator/src/test/longhaul/README.md` so contributors aren't left wondering "what is this tree?"

---

*Reviewed with the GitHub Copilot code-review agent.*

docs/designs/long-haul-test-design.md
WentingWu666666: Good call! I've added an **Alerting** section to the design (see the Deployment section). The approach is a two-layer architecture:

**Layer 1 (AKS):** Canary writes status to a ConfigMap  no GitHub tokens on the cluster.
**Layer 2 (GitHub Actions):** Hourly cron workflow checks canary health, creates a GitHub Issue on failure (with deduplication). Maintainers get email via GitHub's notification system.

This avoids long-lived token management  GitHub Actions provides GITHUB_TOKEN automatically. Added as Phase 2h in the implementation plan.

docs/designs/long-haul-test-design.md
WentingWu666666: Great question. Multi-region is planned as **Phase 3** in the design — it covers add/remove region operations, cross-region replication verification, and AKS Fleet integration.

For regional outages: the canary's event journal + health monitor will detect when a region goes down. The outage policy system classifies whether errors during a regional disruption are expected (tolerable) or unexpected (bug). We can also run independent canary instances in different regions once we get to Phase 3.

For Phase 1-2, we start with a single-cluster canary to prove out the test framework, then extend to multi-region.

WentingWu666666:
## Response to @xgerman's Review

Thank you for the thorough review! I've addressed all items. Here's a summary:

---

###  Major Items

**M1 (Module placement):** Agreed  extracted long-haul tests to `test/longhaul/` at the repo root as a **separate Go module** (`github.com/documentdb/documentdb-operator/test/longhaul`). This keeps Ginkgo/Gomega and any future test-only deps completely out of the operator's runtime `go.mod`. Follows the three-directory layout described in the updated design doc.

**M2 (E2E engagement):** Designed a **three-directory layout** for test code reuse:
- `test/utils/`  shared utilities (future extraction from PR #346)
- `test/e2e/`  E2E test suite (PR #346)
- `test/longhaul/`  long-haul canary tests (this PR)

Each directory has its own `go.mod` with `replace` directives. The design doc now includes a package-to-use mapping table showing which shared utils each long-haul phase will reuse (mongo client, assertions, DocumentDB CR helpers, etc.). Extraction of shared packages from `test/e2e/pkg/e2eutils/` to `test/utils/` will be coordinated with @xgerman in a follow-up.

**M3 (Design gaps):** Upgraded from Open Questions to **Provisional Design Decisions** in the design doc:

1. **Journal durability:** PVC-backed file at `/data/journal/`, structured JSON lines, on-startup reconstruction from existing file.
2. **Writer seq resumption:** Bootstrap from `max(seq)` per `writer_id` in DB. Oracle tolerates crash-recovery gaps (logged, not flagged as data loss).
3. **Teardown on abort:** Signal handler for SIGTERM/SIGINT  cancel contexts  flush journal  write final status to ConfigMap  exit. Leftover-run detector on startup.
4. **Latency baseline:** Kept as Open Question  establish P99 baseline during first 30min warmup, alert on >2 sustained regression.

These lock the approach for Phase 1b/1c interfaces while leaving implementation details to those PRs.

**M4 (MaxDuration default):** Added notes in both the design doc (Configuration section) and README (CI Safety section) that the persistent canary Job manifest explicitly sets `LONGHAUL_MAX_DURATION=0s`. The 30m default remains as a safety net for local development.

---

###  Minor Items  All Fixed

- **m1:** Added `strings.TrimSpace` to `IsEnabled()` 
- **m2:** Added negative `MaxDuration` validation in `Validate()` 
- **m3:** Lowercased error strings (Go convention ST1005) 
- **m4:** Fixed path: `test/longhaul/deploy/job.yaml` 
- **m5:** Rephrased CI safety: *"No CI workflow currently sets this variable  do not add it to any PR-gated workflow"* 
- **m6:** Valid concern. Keeping `config` for Phase 1a (no shadowing risk yet). Will revisit if the operator's own config needs importing in later phases.
- **m7:** Added 8 new test cases: whitespace (`" true "`, `" yes "`), whitespace-only (`"   "`), mixed case (`"True"`, `"YES"`), explicit false values (`"0"`, `"no"`, `"false"`). Total: 23 specs (was 15). 

---

### Questions

1. **(M1)** Resolved  extracted to separate Go module at `test/longhaul/`.
2. **(M2)** Resolved  three-directory layout with shared utils plan documented.
3. **(M3.1)** PVC-backed file journal  documented as provisional design decision.
4. **(Issue #220)** The design's Problem Statement already quotes the three requirements from #220. SLO targets remain as Open Question #2  they need team input.
5. **(CONTRIBUTING.md/AGENTS.md)** Good idea. Will add a one-liner in a follow-up PR to avoid scope creep here.

---

**Test results:** 23 config specs + 1 canary spec (skipped) = all passing. `go vet` clean.

hossain-rayhan:
The plan looks good. 
I have one concern about the test restart policy. 

We should keep running parallel tests for different combination. Like, operator n version with gateway/extension n-1 version. Operator n-2 version with keeps working fine with latest non-schema changes. 

At least, different operator version tests, not only the latest release restarts the tests.

WentingWu666666:
Thanks for the feedback @hossain-rayhan! Great point about version compatibility.

Since the operator and database images follow independent version tracks, we agree cross-version testing matters  e.g., upgrading the operator while the database is still on the previous version, and vice versa.

That said, we think this is better suited for the **E2E test suite** rather than long-haul. Cross-version compat bugs surface immediately on upgrade (deterministic, finite), while long-haul focuses on **time-dependent** issues  memory leaks, connection pool exhaustion, replication drift  that only appear after hours/days of sustained operation.

We'll make sure the E2E suite covers both operator-first and database-first upgrade paths with version N/N-1 combinations.

docs/designs/long-haul-test-design.md
```diff
@@ -0,0 +1,633 @@
...
+| On-demand backup | Backup | None | Backup CR reaches Completed |
+| Restore to new DocumentDB cluster | Backup | N/A (new cluster) | Restored data matches backup watermark |
+| Scheduled backup verify | Backup | None | Backups created on schedule |
+| Operator upgrade | Update | None (DB pods should NOT restart) | Operator pod rolls, DocumentDB cluster unaffected |
+| DocumentDB binary upgrade | Update | Rolling restart | Pods restart one-by-one, workload continues |
+| Schema upgrade | Update | Varies | Pre-backup, post-upgrade reads/writes OK |
+| Operator restart/leader failover | Chaos | Brief reconcile gap | Reconciliation resumes |
+| Pod eviction (simulating node drain) | Chaos | Brief | Pod rescheduled, workload resumes |
+
+**Sequencing Rules:**
```
xgerman: Constraints?

docs/designs/long-haul-test-design.md
```diff
@@ -0,0 +1,633 @@
...
+LONGHAUL_CLUSTER_NAME=documentdb-sample \
+./longhaul.test -test.v -test.timeout 0
+```
+Runs against whatever Kubernetes cluster your kubeconfig points to (Kind, Minikube, etc.).
+
+**Persistent canary (internal):**
+- Dedicated AKS cluster provisioned once (manually or via IaC)
+- Long haul test deployed as a Kubernetes Job on the same AKS cluster (separate `longhaul` namespace)
+- On new operator release: re-deploy operator via Helm + restart longhaul Job
+- Internal Grafana/OTel dashboard for monitoring (optional)
```
xgerman: Remove internal

docs/designs/long-haul-test-design.md
```diff
@@ -0,0 +1,633 @@
...
+- DocumentDB cluster preserved on failure for investigation
+
+> **Note:** The canary runs on a team-managed AKS cluster. Contributors do not need cluster access —
```
xgerman: Remove, and keep more general

docs/designs/long-haul-test-design.md
```diff
@@ -0,0 +1,633 @@
...
+- Long-haul canary runs as a Kubernetes Job — continuous workload
+- Writes status to a well-known ConfigMap (`longhaul-status` in `longhaul` namespace)
+- Updates include: current state (running/failed/passed), last heartbeat, failure details, journal excerpt
+- No GitHub token needed on the AKS cluster
+
+**Layer 2: GitHub Actions (periodic health check)**
+- Scheduled workflow runs every hour (`cron: '0 * * * *'`)
+- Connects to AKS cluster via Azure federated identity (OIDC, same as auto-upgrade workflow)
+- Checks canary health: pod status, status ConfigMap, recent pod logs
+- If failure detected → creates a GitHub Issue with:
```
xgerman: we need human oversight before creating an issue (human in the loop)

xgerman (changes requested):

docs/designs/long-haul-test-design.md
WentingWu666666: Expanded to a full **Sequencing Constraints** table with explicit rules:

| Constraint | Rule |
|-----------|------|
| Minimum topology | Cannot scale below minNodeCount |
| Region cardinality | Cannot remove region if only 1 exists |
| Replication prerequisite | Cannot disable replication while multi-region is active |
| Concurrent disruptive ops | At most 1 at a time |
| Cooldown | Configurable min interval (default 5 min) |
| Steady-state gate | Previous op must complete before next |
| Backup isolation | Restore runs on separate DocumentDB cluster |

Each operation declares preconditions (checked before) and postconditions (verified after). If a precondition fails, the scheduler picks a different operation.

docs/designs/long-haul-test-design.md
WentingWu666666: Removed "Internal"  now just "Grafana/OTel dashboard for monitoring (optional)". Aligns with the portability goal.

docs/designs/long-haul-test-design.md
WentingWu666666: Already addressed in the portability update (commit 1d63aea). The old AKS-specific Note block was rewritten to:

> The canary Job manifest, RBAC, and test binary are fully portable  they use only core Kubernetes APIs and work on any conformant cluster. The core team runs an automated instance with GitHub Actions-based alerting and auto-upgrade (see below), but any contributor can deploy the same artifacts on their own cluster, observe results locally (pod logs, ConfigMap status), and file bugs if issues are found.

docs/designs/long-haul-test-design.md
WentingWu666666: Good point  changed to **human-in-the-loop**. The alerting is now split into two workflows:

1. **Health Check** (hourly cron): Detects failures and posts a report to the GitHub Actions workflow summary. Does NOT auto-create issues.

2. **Create Issue** (\workflow_dispatch\ only): A maintainer reviews the failure evidence from the health check, then manually triggers this workflow to file a GitHub Issue.

This ensures human oversight  no noisy auto-created issues from transient infra failures or flaky conditions.

hossain-rayhan:
> Thanks for the feedback @hossain-rayhan! Great point about version compatibility.
> 
> Since the operator and database images follow independent version tracks, we agree cross-version testing matters e.g., upgrading the operator while the database is still on the previous version, and vice versa.
> 
> That said, we think this is better suited for the **E2E test suite** rather than long-haul. Cross-version compat bugs surface immediately on upgrade (deterministic, finite), while long-haul focuses on **time-dependent** issues memory leaks, connection pool exhaustion, replication drift that only appear after hours/days of sustained operation.
> 
> We'll make sure the E2E suite covers both operator-first and database-first upgrade paths with version N/N-1 combinations.

We will have long-haul tests only for the latest version then. I wanted to keep testing at least one older version. But that's fine maybe. Because our release cadence is bi-monthly and that's enough testing for now.

hossain-rayhan (approved):

xgerman (changes requested):
## Summary

Substantial PR — a long-haul (canary) test driver as its own Go module under `test/longhaul/`, plus three GHA workflows (build / deploy / hourly monitor) and a 367-line design doc. The architectural shape is sound: `Deployment` (not `Job`) so the ConfigMap is the durable verdict, append-only journal as the disruption-window oracle, weighted-random scheduler, separate writer/verifier goroutines, real cluster client (no fakes). The design doc is good — explicit failure-modes catalogue, named principles, phase tracker.

But several issues are blockers — most importantly **the scale operations target the wrong CR field** and will be rejected by the API server every time, and **the unit-test coverage promised in the PR body is absent for 5 of 6 packages**.

I also reviewed alignment with the e2e suite pattern landing in #346 (`developer/e2e-suite`). Headline: the long-haul driver re-implements scaling, CR access, and Mongo connection plumbing that already exist in `test/e2e/pkg/e2eutils/`, and it gets the scale-target field *wrong* in a way that the e2e helpers would have prevented. Details under §"Alignment with the e2e suite from #346" below.

CI status: 4 build matrix rows show FAILURE, 4 CANCELLED on the merge attempt — the long-haul driver does not appear to be the cause, but please confirm the `Build DocumentDB Packages (amd64)` failure is unrelated before merge.

---

## Critical Issues

### 🔴 [Required] `ScaleUp` / `ScaleDown` patch the wrong CR field

`monitor/k8sclient.go` `ScaleCluster` and `GetCurrentReplicas` both target `spec.nodeCount`:

```go
patch := map[string]interface{}{
    "spec": map[string]interface{}{
        "nodeCount": replicas,
    },
}
```

But the CRD (`operator/documentdb-helm-chart/crds/documentdb.io_dbs.yaml`) pins:

```yaml
nodeCount:
  description: NodeCount is the number of nodes in the DocumentDB cluster. Must be 1.
  maximum: 1
  minimum: 1
```

The actual scalable field is `instancesPerNode` (range 1..3). Every `ScaleUp.Execute` call at runtime patches `nodeCount: 2` and the API server rejects it — the operation always fails, no real scale event is exercised, and `OpsExecuted` over-counts. This silently breaks one of the three operations the PR ships.

**Fix:** Patch `spec.instancesPerNode` instead, and clamp `MinReplicas`/`MaxReplicas` to 1..3. Or — preferred — call `test/e2e/pkg/e2eutils/documentdb.PatchInstances`, which already validates the range and is the pattern PR #346 established (see §"Alignment" below).

Confirms the deployment manifest already half-knew this — `deploy/deployment.yaml` calls the env vars `LONGHAUL_MIN_INSTANCES` / `LONGHAUL_MAX_INSTANCES`, and even comments *"Bounds for spec.instancesPerNode (CRD allows 1-3)"*. The naming is right; the implementation is wrong.

### 🔴 [Required] Deployment ConfigMap env-var keys do not match `config.LoadFromEnv`

`deploy/deployment.yaml` sets:

```yaml
LONGHAUL_MIN_INSTANCES: "1"
LONGHAUL_MAX_INSTANCES: "3"
```

But `config/config.go` only reads:

```go
EnvMinReplicas = "LONGHAUL_MIN_REPLICAS"
EnvMaxReplicas = "LONGHAUL_MAX_REPLICAS"
```

So in production these two env vars are silently dropped and the driver runs with the `DefaultConfig` values `MinReplicas=2, MaxReplicas=5`. Combined with the `nodeCount` bug above, both bounds are nonsense (the field they target is locked to 1).

**Fix:** Pick one name (`INSTANCES` is correct given the CRD field) and use it everywhere — env constants, struct fields, ConfigMap keys, README, design doc.

### 🔴 [Required] Promised unit tests are absent for 5 of 6 packages

PR body claims: *"Tier-1 unit tests across `config`, `journal`, `monitor`, `report`, `operations`."*

Reality (`go test ./...` from `test/longhaul/`):

```
ok      .../config       0.333s
?       .../cmd/longhaul    [no test files]
?       .../journal         [no test files]
?       .../monitor         [no test files]
?       .../operations      [no test files]
?       .../report          [no test files]
?       .../workload        [no test files]
```

Only `config_test.go` + `suite_test.go` exist. Given that `journal.HasPolicyViolation`, `leakdetect.Analyze` (linear regression), `report.GenerateMarkdown`, and `operations/scheduler.selectOperation` (weighted-random + cooldown gating) all encode non-trivial logic, this is the gap most worth closing before merge.

`/needs-tests`

### 🔴 [Required] `instancesPerNode < 2` auto-skip is documented but not implemented

PR body and `test/longhaul/README.md` both state:

> The `upgrade-documentdb` operation auto-skips when `spec.instancesPerNode < 2`…

`operations/upgrade.go` `Precondition`:

```go
func (u *UpgradeDocumentDB) Precondition(ctx context.Context) (bool, string) {
    desired, err := u.readDesiredVersion(ctx)
    if err != nil { ... }
    if desired == "" { return false, "no desired version published" }
    running, err := u.client.GetCurrentDocumentDBImageTag(ctx)
    ...
    if running == desired { return false, ... }
    return true, ""
}
```

There is **no** `instancesPerNode` check anywhere in this file (or in `monitor.ClusterClient`). On the live canary cluster `setup.yaml` deploys with `instancesPerNode: 1`, which means the upgrade operation *will* fire and *will* produce real downtime that no operator change can prevent — the exact false-positive the README promises is gated. Either implement the gate or remove the claim.

---

## Major Issues

### 🟠 [Required] `Verifier.verifyAll` scans every document on every tick — won't survive 24h

`workload/verifier.go`:

```go
ticker := time.NewTicker(verifyInterval)  // 10s
...
for cursor.Next(ctx) {
    var doc WriteDocument
    cursor.Decode(&doc)
    // gap + checksum check
}
```

The verifier finds distinct writers, then `Find` *all* documents per writer, ordered by seq, and walks them every 10s. With 5 writers × 1 write per 100ms = 50 writes/s × 86,400 s/day = 4.3M docs/day. After 24h every verifier tick is reading 4.3M rows from the cluster every 10s. Single-day runs are arguably acceptable; the design doc explicitly targets multi-day runs.

**Fix:** Track `lastVerifiedSeq` per writer in the Verifier struct and `Find({writer_id, seq: {$gte: lastVerifiedSeq}})`. Same gap-detection logic, bounded scan.

### 🟠 [Required] `Writer.writeOne` mis-classifies "succeeded server-side, failed client-side" as a write failure

`workload/writer.go`:

```go
_, err := w.collection.InsertOne(ctx, doc)
if err != nil {
    w.metrics.WriteFailed.Add(1)
    w.journal.RecordWriteFailure()
    return
}
```

If the network drops the ACK after MongoDB committed (very common during the disruptive operations this driver simulates), the row is durably written but the client sees an error and:
1. Increments `WriteFailed` (skews the success rate metric).
2. Calls `RecordWriteFailure` on the active disruption window — and `OutagePolicy.AllowedWriteFailures=50` for ScaleUp can be exceeded by a few of these false-failures alone, producing a false `FAIL` verdict.

Sequence numbers are atomic so the row is unique; you can disambiguate by checking duplicate-key on the *next* write or by reading-back the last seq on the connection. At minimum, swallow `IsDuplicateKeyError` and treat it as success on a writer-level retry.

### 🟠 [Required] `go.mod` declares `go 1.26.0` — diverges from rest of repo

`test/longhaul/go.mod`:
```
go 1.26.0
```

But `test/e2e/go.mod` is `go 1.25.9` and `operator/src/go.mod` is `go 1.25.x`. The README in `test/longhaul/README.md` says "Go 1.25+". This module is an outlier. Several of the indirect deps are dated `*-20260120151049-...` / `*-20260317180543-...` (i.e. pulled from very recent main snapshots).

**Fix:** Pin `go 1.25.x` to match the parent repo. Re-run `go mod tidy` against released versions of the k8s.io/* deps.

### 🟠 [Suggestion] `OutagePolicy.AllowedDowntime` is dead code

The field is declared on `OutagePolicy`, set by `ScaleUp` (`30s`), `ScaleDown` (`60s`), `UpgradeDocumentDB` (`120s`), but never consulted in `DisruptionWindow.ExceededPolicy()`:

```go
func (w *DisruptionWindow) ExceededPolicy() bool {
    if w.Duration() > w.Policy.MustRecoverWithin { return true }
    if w.WriteFailures > w.Policy.AllowedWriteFailures { return true }
    return false
}
```

PR body acknowledges this ("the field is reserved on the struct"). At minimum add a `// TODO(longhaul): enforce AllowedDowntime; today only WriteFailures + MustRecoverWithin gate the verdict.` so the dead-field doesn't get accidentally relied on by a future operation author who reads the field but not the PR description.

### 🟠 [Suggestion] `UpgradeDocumentDB.waitForImage` swallows polling errors silently

`operations/upgrade.go`:

```go
case <-ticker.C:
    tag, err := u.client.GetCurrentDocumentDBImageTag(ctx)
    if err != nil {
        continue  // silent
    }
```

If the Kubernetes API auth is revoked mid-run (kubeconfig rotation, ServiceAccount deleted), this loop spins until `recovery` timeout with no journal record of *why*. Add a `j.Warn("upgrade", fmt.Sprintf("status read error: %v", err))` so the operator knows whether they're looking at a slow upgrade or a broken kubeconfig.

### 🟠 [Suggestion] Scheduler tries every operation's `Precondition` every 10s — API spam

`operations/scheduler.go` `selectOperation` calls `op.Precondition(ctx)` on *every* operation on *every* 10s tick, regardless of whether the cooldown has elapsed. Each precondition for ScaleUp/ScaleDown is one CR GET; UpgradeDocumentDB is two (ConfigMap + CR). At 10s × 3 ops × ~3 calls = 1 call/s on the cluster's API just for precondition dispatch.

Cheaper: filter cooldown *first* (already done in `tryExecute`), but also gate `Precondition` calls behind `IsSteadyState()` (already done) — fine. The remaining sprawl is mostly the upgrade op reading the version ConfigMap once a tick. Acceptable; flagging for awareness.

---

## Alignment with the e2e suite from PR #346

This is the part you specifically asked about. The long-haul module is intentionally separate (long-running binary, not a Ginkgo suite — that's correct), but several pieces it re-implements badly already exist as tested helpers in `test/e2e/pkg/e2eutils/`:

| What longhaul reimplements | What `test/e2e/pkg/e2eutils/` already provides | Verdict |
|---|---|---|
| `monitor.ScaleCluster` patches `spec.nodeCount` (broken) | `documentdb.PatchInstances(ctx, c, ns, name, want)` patches `spec.InstancesPerNode`, validates 1..3, no-ops when already at target | **Use it.** Would have prevented the 🔴 nodeCount bug. |
| `monitor.K8sClusterClient` (untyped `dynamic.Interface` for the CR) | `documentdb.Get` / `documentdb.Eventually` (typed `*v1preview.DocumentDB`) | Likely too heavyweight to import as-is given the separate go.mod; but the field-name *facts* (`InstancesPerNode`, status conditions) should be sourced from the typed API to avoid drift. |
| `cmd/longhaul/main.go` raw `mongo.Connect(options.Client().ApplyURI(...))` | `pkg/e2eutils/mongo/connect.go` already wraps connect/ping with the project's error semantics and TLS handling | Worth a look — at minimum, read `connect.go` so the long-haul connection logic doesn't drift from the suite's. |
| Health = "are pods Ready + CR.status=='Cluster in healthy state'" hard-coded as a string | `assertions/` already has cluster-readiness predicates that tolerate transient status messages | Soft suggestion. |

The current long-haul go.mod is a separate module so direct import is awkward, but **none of the above has to be a copy-paste**. The minimum bar is: when the long-haul driver patches a field on `documentdb.io/preview.DocumentDB`, the field name should match the typed Go struct in `operator/src/api/preview/`, not be a hand-written string. A unit test that round-trips the patch against the typed struct would catch the `nodeCount` bug.

The label/level vocabulary from #346 (`level:medium`, `level:long-haul`, `SkipUnlessLevel`, `--label-filter`) does **not** apply to a deployed Deployment — that's fine, it's a different runtime model. But a future improvement: if `cmd/longhaul/main.go` exposed a Ginkgo-mode entrypoint, the same binary could be invoked under `TEST_DEPTH=Lowest` from the e2e workflow as a 60-minute "shallow longhaul" smoke. Out of scope here, worth a follow-up issue.

### 🟡 [Suggestion] Have longhaul depend on the e2e module for field constants

If you don't want to vendor `e2eutils.documentdb`, at least:

```go
import preview "github.com/documentdb/documentdb-operator/operator/src/api/preview"
// ...
const fieldInstancesPerNode = "instancesPerNode"  // matches preview.DocumentDBSpec.InstancesPerNode bson tag
```

…with a compile-time assertion test that the JSON tag of `preview.DocumentDBSpec.InstancesPerNode` is `"instancesPerNode"`. Drift-prevention without a runtime dep.

---

## Suggestions

### 🟡 [Nitpick] `ClusterHealth.WritesHealthy` is unused

Defined on the struct, never set, never read. Delete it or wire it up to `Metrics.WriteSuccessRate()`.

### 🟡 [Nitpick] `setup.yaml` references `deploy/job.yaml` in its header comment

```yaml
# After the cluster is healthy (~3-5 min), deploy the long-haul test:
#   kubectl apply -f deploy/rbac.yaml
#   kubectl apply -f deploy/job.yaml
```

There is no `job.yaml` — it was renamed to `deployment.yaml` (correctly, per the "Why a Deployment" comment in that file). Update the README/comments.

### 🟡 [Nitpick] `setup.yaml` uses `nodeCount: 1, instancesPerNode: 1`

Consistent with the CRD, but means **upgrade auto-skip can't fire** (it isn't implemented anyway, see Critical), AND **ScaleUp can't succeed** (nodeCount=1 max=1). After fixing the scale bug, set `instancesPerNode: 2` here so the canary actually exercises the HA path the design doc requires for FM4.

### 🟡 [Nitpick] `report.EmitAnnotation` calls `Round(1)` on a `time.Duration`

`s.Duration.Round(1)` rounds to 1 nanosecond — a no-op. You probably meant `Round(time.Second)` or `Round(time.Minute)`. The `report.go` `GenerateMarkdown` does it correctly (`Round(time.Second)`).

### 🟡 [Nitpick] `Writer.writeOne` payload includes `time.Now().UnixNano()`

The checksum is `SHA256(writerID:seq:payload)`. Since payload includes a timestamp, the checksum is non-reproducible — the verifier recomputes the same checksum from the *stored* payload, so it works, but the test cannot detect *partial-write* corruption (where the stored payload differs from what was sent). Consider using a content-addressable payload (e.g. `payload = sha256(writerID+seq)`) so the checksum becomes a true durability oracle.

### 🟢 [Nitpick] Workflow secret format

`longhaul-deploy.yml` and `longhaul-monitor.yaml` both `printf '%s' "${{ secrets.LONGHAUL_KUBECONFIG }}"`. Consider reusing the kubeconfig setup as a small composite action or shared script — it's identical in both workflows and the next workflow that needs it will copy-paste again.

---

## Questions

1. **Phase tracker vs CRD reality** — the design doc lists scale-up/scale-down/upgrade as Phase 1f complete. Given the `nodeCount` bug, has the live canary at `longhaul-aks` actually executed any scale operations successfully? `kubectl get cm longhaul-report -n documentdb-test-ns -o yaml` would tell us — happy to look if you can paste the latest report.

2. **`upgrade-if-needed` job in `longhaul-monitor.yaml`** — this auto-upgrades the operator (via `helm upgrade`) on every successful hourly run. Is that the intended cadence? An operator regression would propagate to the canary within ~1h with no human gate. The design doc's Principle 8 says *"Human in the Loop for Alerts"* — but this is auto-deploying upstream changes in addition to alerting on them. Worth at least documenting the policy.

3. **Why a separate Go module rather than a sub-package of `test/e2e`?** The PR body says it's "its own Go module" but doesn't motivate that choice. It costs reuse (see §"Alignment") and the build/CI plumbing is largely duplicated. Was it to avoid pulling Mongo client into the e2e module's deps, or something else?

---

## Positive Feedback

- **Design doc is excellent.** The eight failure-modes table, the principles-as-negative-constraints framing, and the layered architecture diagram do real work. Reviewers reading this for the first time can answer "why this not that?" without asking.
- **`Deployment` over `Job` is the right call** and the comment in `deployment.yaml` explains why succinctly. Future readers won't try to "fix" it back.
- **The journal/disruption-window oracle is the right primitive.** Tying outage-policy verdicts to journaled windows rather than wall-clock SLOs makes failures reproducible.
- **Distroless + nonroot Dockerfile**, sensible resource limits (100m/128Mi req, 500m/256Mi lim), `Recreate` strategy with single replica + PVC-stateful comment — container hygiene is right.
- **`config.go` has actual tests**, including the case-insensitive `IsEnabled` matrix and the "invalid duration" error path. This is the test-shape the other 5 packages should follow.
- **Hourly monitor's `kubectl apply -f operator/documentdb-helm-chart/crds/` before `helm upgrade`** correctly handles the Helm-doesn't-upgrade-CRDs gotcha, with a comment that explains why.
- **No hardcoded image tags in the workflow** — `:sha-<short>` is computed from the upstream build's `head_sha`. The `:main` rolling tag is documented as "convenience only; pin to :sha-<short>."

---

## Verdict

`/request-changes`

Blockers (`nodeCount` patch, env-var name mismatch, missing unit tests, missing `instancesPerNode` precondition) are all small fixes individually — the architecture and the design are sound. Once the scale field is corrected and 4 of 5 packages get a starter `_test.go`, this is mergeable.

`/needs-tests` — for `journal`, `monitor`, `report`, `operations`, `workload`.

WentingWu666666:
@xgerman thanks again for the detailed review  all items from your latest review are now addressed. Quick disposition summary:

**Critical**
1.  ScaleUp/Down now patch `spec.instancesPerNode` (not `nodeCount`).
2.  Env vars renamed to `LONGHAUL_MIN_INSTANCES` / `LONGHAUL_MAX_INSTANCES` everywhere (code, README, workflow).
3.  Unit tests added for all 6 packages (`config`, `journal`, `monitor`, `operations`, `report`, `workload`)  commit `463c0ae`.
4.  Upgrade auto-skips when `instancesPerNode < 2` (`operations/upgrade.go` Precondition).

**Major**
5.  Verifier resumes from `nextSeq` instead of full re-scan every cycle.
6.  Writer now treats `mongo.IsDuplicateKeyError` as success (idempotent retry).
7.  `test/longhaul/go.mod` aligned to `go 1.25.9` to match operator/e2e  needed downgrading k8s.io deps from v0.36.0  v0.35.0 since v0.36 requires Go 1.26+ (commit `e17fcf3`).
8.  `OutagePolicy.AllowedDowntime` now has a TODO + enforcement skeleton in `journal/policy.go`.

**Suggestions / Nitpicks**
9.  `waitForImage` now logs polling errors instead of swallowing them.
10. Acknowledged  scheduler poll cadence kept as-is.
11. Out of scope for this PR (cross-module API pin).
12.  Removed unused `ClusterHealth.WritesHealthy`.
13.  `setup.yaml` header now references `deployment.yaml`.
14.  `setup.yaml` defaults to `instancesPerNode: 2`.
15.  `EmitAnnotation` uses `Round(time.Second)`.
16. Declined  keeping timestamp-based payload to preserve write-time observability.
17. Suggestion only  workflow secret format unchanged.

Verified locally: `go build ./...` and `go test ./...` from `test/longhaul/` are green. Ready for re-review when you have a moment.

xgerman:
## Review: long-haul harness vs. existing `test/e2e/pkg/e2eutils/`

**Verdict:** the PR does not leverage the existing e2e helpers and ships meaningful duplication of the DocumentDB CR contract and Mongo connection plumbing.

### Root cause
`test/longhaul/go.mod` is a standalone module with no `replace github.com/documentdb/documentdb-operator => ../../operator/src` directive and no dependency on the operator or e2e module. As a result, longhaul cannot import `previewv1` or `test/e2e/pkg/e2eutils/`, and reimplements them via raw `dynamic.Interface` + `unstructured`. `test/e2e/go.mod:110` has the replace directive that makes this work for the e2e suite.

### Concrete duplications

**1. `test/longhaul/monitor/k8sclient.go` duplicates `test/e2e/pkg/e2eutils/documentdb/documentdb.go`**

| Longhaul (raw dynamic) | E2E (typed) |
|---|---|
| `K8sClusterClient.ScaleCluster` (k8sclient.go:178) | `documentdb.PatchInstances` (documentdb.go:299) |
| `K8sClusterClient.UpgradeDocumentDB` (k8sclient.go:221) | `documentdb.PatchSpec` — generic mutator (documentdb.go:325) |
| `K8sClusterClient.GetInstancesPerNode` (k8sclient.go:155) | `documentdb.Get` + field access (documentdb.go:428) |
| `K8sClusterClient.GetCurrentDocumentDBImageTag` (k8sclient.go:200) | `documentdb.Get` + field access |
| `K8sClusterClient.GetClusterHealth` — `status == "Cluster in healthy state"` hardcoded at k8sclient.go:148 | `documentdb.ReadyStatus` constant (documentdb.go:68) |
| Polling-for-healthy in `HealthMonitor.check` (health.go:93) | `documentdb.WaitHealthy` (documentdb.go:346) |

The hardcoded readiness string is the most concerning item: if the operator ever changes its `ReadyStatus`, e2e picks it up automatically; longhaul silently breaks.

**2. `cmd/longhaul/main.go` mongo connection duplicates `test/e2e/pkg/e2eutils/mongo/`**

- main.go does raw `mongo.Connect(options.Client().ApplyURI(cfg.MongoURI))` plus a manual ping.
- e2e already exports `mongo.NewClient`, `mongo.BuildURI`, and `mongo.NewFromDocumentDB` — the last reads the `documentdb-credentials` Secret, opens a port-forward when needed, applies TLS, and retries pings.
- Longhaul sidesteps the credentials Secret by requiring an externally-injected `LONGHAUL_MONGO_URI` env var via a separate `longhaul-mongo-credentials` Secret (`deploy/setup.yaml`). That's an operational workaround, not a code-reuse decision — `mongo.BuildURI` and the credentials-secret reader still apply even when the cluster is LB-exposed.

**3. GVR hardcoding**

`monitor/k8sclient.go:78-82` hardcodes `{Group: "documentdb.io", Version: "preview", Resource: "dbs"}`. The typed client encodes the same fact via `AddToScheme` and would never drift. (Note the resource is `dbs`, not the more obvious `documentdbs` — exactly the kind of detail you only get wrong once when you bypass the scheme.)

**4. Operator pod health monitoring**

`monitor/health.go` ad-hoc counts pod readiness. `test/e2e/pkg/e2eutils/operatorhealth/gate.go` already encapsulates "snapshot operator pod identity + restart count, detect churn" — the exact primitive needed to distinguish "operator died" from "operator was upgraded mid-run".

### Not duplicated (longhaul-specific, fine as-is)

`workload/verifier.go`, `workload/writer.go`, `monitor/leakdetect.go`, `journal/`, `operations/scheduler.go`, `report/report.go` are genuinely longhaul-only — keep them where they are.

### Recommendations (ranked)

1. **Add the operator replace directive and an e2eutils dependency** to `test/longhaul/go.mod`, *or* extract `e2eutils/{documentdb,mongo,operatorhealth,portforward}` into a `test/shared/` module that both `test/e2e/` and `test/longhaul/` consume. The shared-module option is cleaner long-term because it avoids importing Ginkgo into a non-Ginkgo binary.
2. Replace the CR operations in `monitor/k8sclient.go` with `documentdb.Get` / `PatchInstances` / `PatchSpec` / `WaitHealthy`, and reference `documentdb.ReadyStatus` instead of the literal string. `operations/scale.go` and `operations/upgrade.go` consume the `monitor.ClusterClient` interface, so this is an internal refactor and won't ripple.
3. Replace `mongo.Connect` in `cmd/longhaul/main.go` with `mongo.NewFromDocumentDB` (or `mongo.NewClient` + `mongo.BuildURI` when an external URI is genuinely required).
4. **Lower priority — even if (1) is rejected as too invasive:** at minimum expose `ReadyStatus` from a small shared constants package and import it. Do not duplicate the magic string.

The harness itself is good work; this is a maintenance-liability call, not a behavioural one. Worth resolving before merge so we don't end up with two source-of-truths for "how to talk to a DocumentDB CR".

PR #351, opened 2026-04-21
feat(monitoring): add sidecar telemetry and playground container metrics

udsmicrosoft:
## Summary

Adds DocumentDB pod-local telemetry through the injected OTel sidecar, and keeps kubelet-backed container metrics out of the production operator chart. The per-Cluster `spec.monitoring.kubeletstats` API and sidecar-based kubeletstats design were removed; clusters should use their existing node-level metrics collector, or the playground reference DaemonSet if they need one for demos/local validation.

## What changed

- keeps the per-pod OTel sidecar focused on pod-local signals: `sqlquery` for Postgres health and gateway OTLP metrics on `127.0.0.1:4317`
- removes `spec.monitoring.kubeletstats` from the preview API and regenerates CRDs
- deletes the operator flow that created and cleaned up per-cluster kubeletstats ClusterRoleBindings
- removes sidecar kubeletstats injection and the sidecar `K8S_NODE_NAME` env var
- removes the container-metrics DaemonSet and `containerMetrics` values from the operator Helm chart
- adds the same container-metrics collector as a playground/reference manifest under `documentdb-playground/telemetry/container-metrics/`
- updates the local telemetry playground to apply that reference manifest for demo clusters
- updates the Grafana dashboard, validation script, CHANGELOG, and monitoring docs to match the split architecture

## Why this design

- tenant DocumentDB pods no longer receive kubelet `nodes/stats` privilege
- the operator no longer manages kubeletstats-specific per-cluster RBAC
- the production operator chart no longer installs generic node-level platform telemetry
- clusters that already run kube-prometheus-stack, an OTel Collector DaemonSet, or a managed monitoring agent can use that existing collector
- the playground still provides a one-command demo path for clusters without existing container metrics

## Validation

- `go test ./internal/otel/...`
- `helm lint operator/documentdb-helm-chart`
- `helm template documentdb-operator operator/documentdb-helm-chart` and verified it no longer renders `documentdb-container-metrics`
- offline YAML parse for `documentdb-playground/telemetry/container-metrics/container-metrics.yaml`
- `bash -n` for the local deploy and validation scripts
- Grafana dashboard JSON parse


WentingWu666666 (changes requested):
Requesting changes  the **user-facing monitoring docs** in this PR lock in "use Prometheus + cAdvisor" as the canonical path for container resource metrics, which leaves OTLP-only users (Datadog, New Relic, Grafana Cloud OTLP, vendor-managed OTel backends) with **no container CPU/memory visibility** for DocumentDB pods. The operator already injects a per-pod OTel Collector sidecar for gateway metrics  adding a `kubeletstats` receiver to that same sidecar would surface container resource metrics through the existing OTLP pipeline, with no extra deployment burden on the user.

This was discussed on #286 (https://github.com/documentdb/documentdb-kubernetes-operator/pull/286#issuecomment-4290046215). The playground itself is fine  it ships Prometheus, so cAdvisor scraping works. The concern is specifically the public docs at `docs/operator-public-documentation/preview/monitoring/metrics.md` which present cAdvisor as *the* answer rather than *one* answer.

### What I would like changed before merge

**Add a `kubeletstats` receiver to the operator''s per-pod OTel Collector sidecar in this PR**, so container resource metrics (CPU, memory, network, filesystem for `postgres`, `documentdb-gateway`, `otel-collector`) flow through the existing OTLP pipeline alongside the gateway metrics. The collector config additions are small (one receiver block + a service-account token mount), and the docs in this PR can then present a single coherent OTel-first story instead of splitting users into "you have Prometheus" vs "you don''t."

Without this, the public monitoring docs commit users to a Prometheus-only assumption that contradicts the rest of the OTel-first telemetry story (#286, #350).

WentingWu666666:
Second concern  **gateway metrics aren''t actually emitted by the latest released DocumentDB image** (v0.110-0). The OTLP exporter code in the gateway lives on `main` but hasn''t shipped in a release yet. Anyone who deploys this playground today against the latest published image will see an empty Gateway dashboard, the OTLP receiver waiting for traffic that never arrives, and confusing "no data" panels that look like a deployment bug.

Two ways to ship cleanly:

1. **Pin the playground to a custom/dev image** that contains the gateway OTLP code, and document this clearly in the README ("requires DocumentDB image >= X, not yet in a public release"). Not great for a public playground.

2. **Remove the gateway metrics scope from this PR entirely**  drop the Gateway dashboard, the OTLP receiver wiring for gateway, and the gateway-specific sections of the metrics doc. Ship just what works against released images today (Postgres / CNPG metrics, alerts, the playground infra). File a follow-up issue to add the gateway dashboard + docs once the gateway image with OTLP support is released.

I prefer option 2: ship a working playground against released images now, and add gateway metrics as a follow-up the moment a gateway release with OTLP lands. This keeps the public docs honest about what users will actually see.

Combined with the kubeletstats request above, the scope of this PR becomes: **playground infra + Postgres/CNPG metrics + dashboards + container resource metrics via kubeletstats**. Gateway metrics get their own follow-up PR tied to the gateway release that ships OTLP.

WentingWu666666:
**Operator metrics: docs describe a path users can''t actually take  suggest splitting into a separate PR**

The new `metrics.md` documents `controller_runtime_*` / `workqueue_*` metrics with PromQL examples, and the playground''s Prometheus config has a `documentdb-operator` scrape job (commented out) with the note *"Requires the operator to be started with --metrics-bind-address=:8443"*.

The problem: **there is no way for a user to actually start the operator with that flag through the helm chart today.**

Looking at `operator/documentdb-helm-chart/templates/09_documentdb_operator.yaml`:

```yaml
args:
- --webhook-cert-path=/tmp/k8s-webhook-server/serving-certs   # only arg
ports:
- containerPort: 9443  # webhook
- containerPort: 8081  # health
# no metrics port, no --metrics-bind-address flag, no Service, no values toggle
```

And `cmd/main.go` defaults `--metrics-bind-address` to `"0"` (disabled). So a user who follows the docs and uncomments the scrape job gets an empty target  the operator pod isn''t listening on `:8443` and the chart exposes no toggle to make it listen. They''d have to manually patch the Deployment.

For comparison, CNPG''s helm chart wires this end-to-end: operator binds `:8080` by default and the chart ships a `PodMonitor` template gated by `monitoring.podMonitorEnabled`.

**Suggestion: split operator metrics into a separate PR.** That PR should:

1. Add a `metrics.enabled` (and `metrics.port`) value toggle in the helm chart
2. Pass `--metrics-bind-address=:<port>` to the operator container when enabled
3. Expose the metrics `containerPort` and add a `Service` (or `PodMonitor`) for scraping
4. Uncomment / update the `documentdb-operator` scrape job in the telemetry playground
5. Keep the `metrics.md` operator section accurate against what actually ships

For **this** PR, please either:
- Remove the operator-metrics content from `metrics.md` and the commented scrape job, or
- Reword the docs to say "operator metrics are not yet exposed  tracked in #<follow-up issue>"

so users aren''t pointed at a configuration they can''t reach.

WentingWu666666:
**CNPG instance metrics: scrape path is Prometheus-only and bespoke  suggest splitting into a separate PR via the OTel sidecar**

(Scoping note: this is about the **instance manager** metrics on `:9187` per Postgres pod  `cnpg_pg_replication_*`, `cnpg_collector_postgres_up`, `cnpg_pg_stat_*`, `cnpg_pg_wal_*`, plus any custom queries. CNPG operator-side metrics are out of scope here.)

### What the PR does today

`prometheus.yaml` has a `cnpg` scrape job that uses pod discovery + relabeling on `cnpg.io/cluster` and the `metrics` named port to scrape `:9187` directly. That works inside the playground but has two problems for real users:

1. **Same OTLP-blind-spot as container resource metrics.** Direct Prometheus scrape of `:9187` means OTLP-only backends (Datadog, NewRelic, Grafana Cloud OTLP) get **no** instance metrics  replication lag, postgres-up, WAL stats, all invisible. This is the same gap I raised for cAdvisor; the fix should be the same: **point the existing per-pod OTel collector sidecar at `localhost:9187` via `prometheusreceiver`** so everything fans out through one OTLP pipe.

2. **Nothing is wired into the helm chart.** The playground''s scrape config is bespoke  a user installing the DocumentDB helm chart with their own Prometheus gets `cnpg_*` automatically only if they copy these relabeling rules by hand. There''s no shipped `PodMonitor` and no DocumentDB CRD knob to opt in/out.

### Should we expose all of them to users?

Worth pausing on this  DocumentDB is a Mongo-compatible surface, and the gateway already emits `db_client_*` OTLP metrics that cover the user-visible operation view. The full `cnpg_pg_stat_*` firehose is Postgres-internals territory that most end-users won''t parse.

I''d suggest exposing a **curated subset** that maps to things users actually act on, and leaving the rest available but not dashboarded:

| Useful to expose | Why |
|---|---|
| `cnpg_pg_replication_lag_seconds` | RPO / replica freshness  directly user-visible |
| `cnpg_collector_postgres_up` | liveness  drives alerts |
| `cnpg_pg_replication` slot lag | replica falling behind / slot blocking WAL recycling |
| `cnpg_pg_wal_*` (size on disk, archived/failed) | capacity & backup health |
| `cnpg_pg_database_size_bytes` | capacity planning |
| Custom query: connection count vs `max_connections` | hits the gateway pool hard |

Skip (or expose without dashboards) everything else from `cnpg_pg_stat_*`  it''s noise for the typical DocumentDB user. The gateway''s `db_client_*` already covers operation rate / latency / errors at the right abstraction level.

### Suggestion

Move CNPG instance metrics to a **separate follow-up PR** that:

1. Wires `localhost:9187` into the existing OTel sidecar''s `prometheusreceiver` (single OTLP egress, parity with gateway + future kubeletstats)
2. Optionally adds a Prometheus exporter on the same sidecar so existing Prometheus users keep working, and ships a `PodMonitor` template in the helm chart gated by a `metrics.instance.enabled` value
3. Curates the dashboards/docs to the subset above instead of dumping the full `cnpg_*` namespace
4. Surfaces a `monitoring.customQueries` passthrough on `DocumentDBSpec` if/when we want DocumentDB-specific Postgres queries

For **this** PR, please remove the `cnpg` Prometheus scrape job + any `cnpg_*` references from `metrics.md` and the dashboards, and link the follow-up issue. That keeps PR #351 focused on what''s actually wired end-to-end (gateway via OTel sidecar) and avoids documenting a path that only works inside this playground.

This is the same pattern as the operator-metrics and cAdvisor comments: the PR documents a metric stream, but the only collection path is direct Prometheus scrape  which leaves OTLP-only users blind and doesn''t carry over to a real helm-chart deployment.

WentingWu666666:
### Concern: privilege model for `monitoring.kubeletstats`

I want to flag the RBAC story behind this field before it lands, because once it ships it''ll be hard to walk back.

**What the field actually grants**

Setting `monitoring.kubeletstats: {}` on a `DocumentDB` CR causes the operator to:

1. Create a `ClusterRoleBinding` granting `nodes/stats` to the CNPG per-Cluster `ServiceAccount` (the SA the Postgres container uses).
2. Mount the kubelet-bearer token from that SA into the OTel sidecar container.
3. Sidecar calls `https://<node>:10250/stats/summary`, which returns container resource stats **for every pod on that node, every namespace, every tenant** -- kubelet has no per-caller filter.

A few properties worth surfacing explicitly:

- `nodes/stats` is a **cluster-scoped subresource**. It cannot be bound via `Role` / `RoleBinding`, and `resourceNames` does not restrict the response payload. There is no way to scope this binding to "only my pods" or "only my namespace."
- The operator itself must therefore hold `rbac.authorization.k8s.io/clusterrolebindings: [create, update, delete]` to manage these per-Cluster bindings. That''s a **privilege-escalation primitive** in the operator''s own SA, present whether or not any user has actually opted in. Hardened-cluster reviewers (Gatekeeper / Kyverno / OPA policies) will flag the operator install over this even when no Cluster has the field set.
- The bearer token sits in `/var/run/secrets/kubernetes.io/serviceaccount/token` of the Postgres pod, with `defaultMode: 0644`, reachable by the Postgres process. A compromised Postgres process -- or any extension that can read pod files -- gets cluster-wide node-stats read.
- Per-Cluster CRBs accumulate. Each cluster lifecycle adds/removes one CRB; orphaned bindings on partial deletions are a real failure mode.

**Persona mismatch**

The CR field reads like a tenant opt-in, but the privilege required to satisfy it is platform-admin level. The tenant who edits their `DocumentDB` CR is not the actor evaluating the trust implication -- the operator pre-grants it on their behalf. That''s the inversion I''d most like to avoid encoding into the public API: tenant-shaped surface, platform-shaped trust.

**The asymmetry vs in-pod metrics**

Engine, gateway, sqlquery, and Postgres-internal metrics all stay inside the pod, on localhost, with no extra RBAC. They fit the sidecar shape cleanly. `kubeletstats` is the one collector being mixed in that breaks that property -- and breaks it for the whole feature, since enabling it requires elevating the entire workload''s identity.

**Alternatives that get equivalent data without cluster RBAC**

1. **Sidecar + read-only PVC mounts + `hostmetrics`/`filesystem` scraper** -> per-PVC disk and inode usage. CNPG''s sidecar plugin can mount the existing `pgdata`/`pg-wal` volumes read-only. No RBAC.
2. **Sidecar + `shareProcessNamespace: true` + `hostmetrics`/`process` scraper** -> per-container CPU/memory via shared `/proc`. No RBAC, no kubelet contact. Tradeoff is the documented one (containers see each other''s process names -- not a fit if anyone passes secrets in argv, which CNPG doesn''t).
3. **Chart-managed DaemonSet collector with its own SA** -> if the goal is fleet-wide container metrics for the DBA persona, a single chart-installed DaemonSet holds `nodes/stats` once, with an SA isolated from any Postgres pod. The operator no longer needs `clusterrolebindings` verbs, the per-Cluster CRD field goes away, and the privilege exists exactly once per chart release rather than once per Cluster.

(1)+(2) cover the tenant-scoped use case from inside the sidecar with zero new privilege. (3) covers the platform/DBA-scoped use case without coupling it to any tenant pod''s identity.

**Concrete asks**

- Either drop `monitoring.kubeletstats` from the per-Cluster CRD and ship container metrics via a chart-toggled DaemonSet, **or** document explicitly in the field''s godoc and in the user-facing docs that:
  - enabling it grants the Postgres pod''s SA cluster-wide node-stats read,
  - the operator carries `clusterrolebindings: create/update/delete` on its own SA whether or not any Cluster opts in,
  - the data exposed via the kubelet endpoint covers pods outside the tenant''s namespace.
- If keeping it, audit-log the binding creation events from the operator (controller-runtime event recorder) so cluster admins can see when a per-Cluster CRB is granted.
- Consider whether `shareProcessNamespace + hostmetrics` (no extra privilege) covers enough of the per-container resource use case to make `kubeletstats` unnecessary in the sidecar entirely.

Happy to put up an alternative PR sketching (3) if it would help compare side-by-side.

WentingWu666666:
### Concern: API shape -- `monitoring.kubeletstats: {}` as an empty-object toggle

Separate from the privilege discussion, the field shape itself has some design debt I''d like to raise while it''s still cheap to change.

**Empty-object as boolean is unobvious**

`monitoring.kubeletstats: {}` uses presence of the field as the on-switch. It works, but:

- `kubeletstats: {}` (enabled) and `kubeletstats:` / field absent (disabled) look identical to most readers and to several common YAML linters.
- Removing the field via `kubectl patch` requires a JSON-patch `remove`, not a `replace` -- harder to script than flipping a bool.
- Defaulting / SSA semantics: if a webhook or another controller writes the field as `{}` for any reason, it silently enables the feature. A bool with `omitempty` doesn''t have that footgun.
- The existing user-facing doc at `docs/operator-public-documentation/preview/monitoring/metrics.md` (around line 256) refers to `monitoring.kubeletstats.enabled` -- the docs already drifted toward the explicit-bool form, suggesting reviewers found the empty-object shape confusing too.

A plain `kubeletstats: { enabled: bool }` is self-documenting, scriptable, and leaves room for non-toggle config (collection interval, metric allow-list) without re-shaping the field later.

**Per-receiver flags don''t scale**

The current shape names a **receiver** (`kubeletstats`) in the public API. The next category of metrics will want the same treatment, and we''ll end up with:

```yaml
monitoring:
  enabled: true
  kubeletstats: { enabled: true }
  hostmetrics:   { enabled: true }
  prometheus:    { enabled: true }   # for kube-state-metrics, etc.
  sqlquery:      { enabled: true }
```

That couples our public CRD to the OTel Collector receiver namespace. If we ever swap the data source (e.g. move container metrics from `kubeletstats` to a DaemonSet, or from `prometheus` to `otlp` push), the field name lies.

A pattern that ages better is naming the **metric category** rather than the **collector**:

```yaml
monitoring:
  enabled: true
  inPodMetrics:        { enabled: true }   # sqlquery, engine, gateway -- no extra RBAC
  containerResources:  { enabled: false }  # CPU/mem/PVC -- may need elevated collection
  clusterTopology:     { enabled: false }  # KSM-style, deferred to platform
```

The implementation is free to pick whatever receiver / collection mode satisfies each category, including swapping it later, without breaking users'' CRs.

**Asymmetry with `monitoring.enabled`**

Right now `monitoring.enabled` is a bool but `monitoring.kubeletstats` is an empty-object toggle. Users will reasonably expect either both to be bools or both to follow the same idiom. The mix invites bug reports of the form "I set `kubeletstats: false` and it stayed on" -- which is technically correct (`false` != absent) but surprising.

**Concrete asks**

- Convert `monitoring.kubeletstats` to `monitoring.kubeletstats.enabled: bool` (or whatever name survives the privilege-concern thread), at least as the documented form. The empty-object shape can stay accepted for backward compat if anyone has already adopted it, but the docs and CRD validation should standardize on the bool form.
- Resolve the docs inconsistency at `docs/operator-public-documentation/preview/monitoring/metrics.md:256` so the field shape is stated identically in godoc, CRD validation, helm values, and user docs.
- Consider whether the public API should name **metric categories** rather than **OTel receivers**, so we can change collection mechanism in the future without a CRD migration.
- If both this and the privilege concern lead to dropping the field from the per-Cluster CRD entirely (in favor of a chart-managed DaemonSet), neither of these API-shape questions matters -- which is itself a useful signal about how to converge.

WentingWu666666:
**Suggestion: move the container-metrics DaemonSet out of the operator chart and into `documentdb-playground/`**

The new chart-managed DaemonSet design (from `templates/11_container_metrics.yaml`) is a big improvement on the per-Cluster CRB approach -- single SA, single CRB, no privilege flowing into tenant pods. Thanks for the redesign.

That said, I'd push to take one more step: ship the DaemonSet as a **playground reference example**, not as part of the operator chart. The sidecar belongs in the operator (it's tenant-scoped and only the operator knows where the gateway socket and Postgres are). The DaemonSet doesn't have that property.

**Why I think it's a platform concern, not an operator concern:**

1. **It's not DocumentDB-aware.** The DaemonSet scrapes every container on every node via the local kubelet -- it has no notion of "DocumentDB pod." Functionally it''s a generic node-level kubelet scraper that we happen to ship.

2. **Most production clusters already have one.** Anyone running DocumentDB at non-trivial scale almost certainly already runs `kube-prometheus-stack`, an existing OTel Collector DaemonSet, or a managed agent (Datadog, AKS Container Insights, GKE Cloud Monitoring, CloudWatch Agent, etc.). If our chart silently installs another one with `prometheus.io/scrape: "true"` annotations on `:8889`, we collide with theirs -- duplicate scrapes, duplicate cost, conflicting dashboards.

3. **It re-expands the operator''s RBAC footprint.** A platform team auditing "what does `helm install documentdb-operator` do to my cluster?" should get a small, honest answer focused on what the operator needs to reconcile DocumentDB. Adding a `nodes/stats` (and `nodes/proxy`) ClusterRole back in -- even scoped to one SA -- makes that answer larger than it needs to be, for a feature most clusters don''t actually need from us.

4. **Defaults are opinionated in a way chart defaults shouldn''t be.** Hard-coded `kubeletstats -> prometheus on :8889` works for users running Prometheus. It doesn''t work for users pushing OTLP to Tempo/Loki/Datadog/CloudWatch/etc. Playground examples are *expected* to be opinionated; chart defaults shouldn''t be.

5. **The `containerMetrics.enabled=true` flag suggests the chart is honest about it being optional, but the install model still bundles it.** If it's truly optional and cluster-platform-shaped, the more honest packaging is "here''s a reference manifest; deploy it if your cluster doesn't already have something equivalent."

**Concrete suggestion**

| Component | Current location | Proposed location | Why |
|---|---|---|---|
| Per-pod sidecar (sqlquery + gateway OTLP) | Operator (sidecar plugin) | Operator (unchanged) | Tenant-scoped, operator-specific, no cluster RBAC |
| Container/node metrics DaemonSet | Operator chart | `documentdb-playground/telemetry/container-metrics/` | Cluster-platform concern, often already provided |
| Internals Grafana dashboard JSON | playground | playground (unchanged) | Already correct |

In the operator docs, add a short "Container metrics" section: *"DocumentDB does not ship a node-level metrics collector. If your cluster already collects kubelet metrics (most do), you''re done. If it doesn't, see the reference example at `documentdb-playground/telemetry/container-metrics/` for a working DaemonSet you can deploy."*

**Migration impact**

For users who installed via the chart with `containerMetrics.enabled=true`: removing the templates and providing the same YAML in playground means a one-time `kubectl apply -f documentdb-playground/telemetry/container-metrics/` -- no functional change, same SA name, same metrics, same Prometheus annotations.

**Counterargument I''d want to address**

The strongest case for keeping it in the chart is the new-user demo path: someone evaluating DocumentDB without a Prometheus stack should get metrics with one command. I think that''s solved by **making the playground path easy** (a `make telemetry` target, or a `helm install documentdb-telemetry ./documentdb-playground/telemetry/`) rather than by the production-install chart silently provisioning cluster-platform infrastructure.

Happy to be wrong here -- curious how others see the operator-vs-playground line for this.

WentingWu666666 (changes requested):
## Monitoring docs: blocking on scope and exporter-neutrality

Re-reading `monitoring/overview.md` and `monitoring/metrics.md` together, both have the same structural issue. Requesting changes because the shape of these docs locks in user expectations for the preview, and the fixes are structural rather than wording.

### 1. Scope creep: cluster monitoring is being documented as DocumentDB monitoring

A large fraction of both docs covers things that aren't DocumentDB-specific:

- Container/pod CPU, memory, network, filesystem (kubelet `/stats/summary`)
- The kubeletstats DaemonSet pattern
- Pod-annotation Prometheus scrape configuration
- Grafana setup

These apply to every Kubernetes workload. Most users come in with kube-prometheus-stack, ADOT, GKE Managed Prometheus, the EKS observability addon, Splunk OTel, Datadog agent, etc. already running. The operator chart correctly does not install or assume any of this (good change in this PR -- DaemonSet moved out of the chart). The docs haven't caught up: they still treat container metrics as if the operator owns them.

The honest scope boundary is:

- **Operator owns:** the sidecar it injects, the `spec.monitoring` API, exporter choice (Prometheus/OTLP), DocumentDB-specific metrics
- **Cluster operator owns:** kubelet metrics, node metrics, scrape infrastructure, visualization

Suggest: drop the container-metrics tables and "Platform node-level collector" lane from both docs. Replace with a short paragraph: *"Pod/container resource metrics are not DocumentDB-specific -- use your existing platform collector. If you don't have one, the playground has a reference kubeletstats DaemonSet. Filter to DocumentDB pods with `k8s_container_name in (\"postgres\",\"documentdb-gateway\")`."*

### 2. OTLP exporter is a first-class API field but the docs only document Prometheus

`MonitoringSpec.Exporter` exposes both `OTLP` and `Prometheus`, and `otel/config.go` renders whichever (or both) is configured. But:

- `overview.md` title: *"... Prometheus, and Grafana"*
- The architecture diagram only shows the Prometheus path
- The "Sidecar pipeline" table lists `prometheus` as the exporter -- implies it's the only one
- The single `spec.monitoring` example sets `exporter.prometheus.port`
- Verification uses `curl http://localhost:9090/api/v1/query`
- `metrics.md` has every metric in two columns -- `Metric (OTel)` and `Metric (Prometheus form)` -- and every example query is PromQL

A user wanting to push to Datadog OTLP / New Relic OTLP / Honeycomb / Grafana Cloud OTLP / a vendor-managed OTel agent / their own central Collector gets no guidance from either doc, despite the API and the operator both supporting it.

Suggest: equal billing for both exporters in `overview.md` (Prometheus pull example + OTLP push example, side by side), and verification split per backend choice. In `metrics.md`, the OTel name is canonical (it's what flows through OTLP); the Prometheus form is one rendering of it.

### 3. Sidecar pipeline internals shouldn't be in user-facing docs

`overview.md` has a Receivers/Processors/Exporters table, mentions `127.0.0.1:4317`, `base_config.yaml` deep-merge mechanics, and ConfigMap content-hashing. `metrics.md` has a "Naming convention" callout explaining the OTel->Prometheus translator's prefix-dropping rules.

Users can't change any of this -- it's operator-internal and will drift as the operator evolves. Suggest moving these to a CONTRIBUTING-style doc or code comments, and replacing in user docs with: *"the operator-rendered sidecar pipeline is internal implementation."*

### 4. Visualization should be downstream choice, not part of the canonical narrative

Grafana shows up in the title, the architecture diagram, and the playground. Playground use is fine -- it's a demo stack. But the *overview* should not anchor on one visualization; users may pick Datadog dashboards, New Relic UI, kubectl top, k9s, Honeycomb, custom Looker, etc.

Suggest: drop "Grafana" from the title and remove it from the architecture diagram. Keep it in the Telemetry Playground section as one example.

### 5. `metrics.md` has the inverse imbalance

Today the doc is roughly:
- ~80% generic kubelet metrics with PromQL queries
- ~5% DocumentDB-specific (the two stub paragraphs for Gateway and CNPG)
- ~15% format/conventions

Once the kubelet-metric tables move out (point 1), what remains honestly reflects the preview: the DocumentDB-specific metrics are still landing. The right move is to lean into that -- list the planned metric groups, mark them as planned, and let the doc grow as instrumentation lands. That's more useful than the current 8-table reference for metrics that aren't ours.

### Recommended doc shapes

**`overview.md`:**

```
1. What you get (sidecar with Postgres health + gateway metrics, vendor-neutral output)
2. Enable (spec.monitoring.enabled + an exporter)
3. Route to your backend
   - Prometheus pull (example)
   - OTLP push (example)
   - Both
4. Pod/container resource metrics: not operator scope; pointer to playground if needed
5. Verify (sidecar logs; per-backend hint)
6. Reference: metrics.md, playground link
```

**`metrics.md`:**

```
1. DocumentDB-specific metrics (Postgres health - TBD, gateway - planned, CNPG - out of preview scope; documented as they ship)
2. Pod/container resource metrics: kubelet-level, not DocumentDB-specific; pointer to upstream kubeletstats receiver docs and playground reference DaemonSet
```

Both ~50% smaller, focused on what the operator actually contracts to deliver, and backend-neutral. Happy to iterate on wording -- main ask is the scope decision before the structure hardens for preview readers.

docs/operator-public-documentation/preview/monitoring/overview.md
```diff
@@ -0,0 +1,177 @@
...
+The operator chart does not install a node-level collector, a Prometheus stack, Grafana, or vendor dashboards. Those are downstream platform choices.
+
+## Architecture
+
+```mermaid
+graph TB
+    subgraph pod["DocumentDB pod"]
+        pg["postgres"]
+        gateway["documentdb-gateway"]
+        sidecar["OpenTelemetry Collector sidecar"]
```
WentingWu666666: Image

Nit: "OpenTelemetry Collector sidecar" and "OTLP backend or central Collector" are long enough that it gets truncated in the rendered preview. Could you please shorten them?

WentingWu666666 (approved):

CHANGELOG.md
```diff
@@ -8,8 +8,10 @@
...
 
 ### Breaking Changes
+- **Removed `spec.monitoring.kubeletstats` (preview API)**: The per-`DocumentDB` toggle for kubelet-sourced container metrics has been removed. Use your platform's existing node-level metrics collector, or apply the reference DaemonSet in `documentdb-playground/telemetry/container-metrics/` if your cluster does not already collect kubelet metrics. This removes the per-tenant `nodes/stats` privilege grant, removes the kubeletstats-specific `bind` permission added for the earlier sidecar design, and stops injecting `K8S_NODE_NAME` into the sidecar. Migration: drop `spec.monitoring.kubeletstats: {}` from your `DocumentDB` resources.
```
WentingWu666666: `spec.monitoring.kubeletstats` was never in the preview API on main.

docs/operator-public-documentation/preview/monitoring/overview.md
udsmicrosoft: Good catch — shortened to `OTel Collector sidecar` (matches the wording already used in the "What the operator owns" table) and `OTLP backend / Collector` in c822f39.

CHANGELOG.md
```diff
@@ -8,8 +8,10 @@
...
 - **Two-Phase Extension Upgrade**: New `spec.schemaVersion` field separates binary upgrades (`spec.documentDBVersion`) from irreversible schema migrations (`ALTER EXTENSION UPDATE`). The default behavior gives you a rollback-safe window — update the binary first, validate, then finalize the schema. Set `schemaVersion: "auto"` for single-step upgrades in development environments. See the [upgrade guide](docs/operator-public-documentation/preview/operations/upgrades.md) for details.
+- **Container metrics reference collector**: The telemetry playground now includes a reference OpenTelemetry Collector DaemonSet under `documentdb-playground/telemetry/container-metrics/` for clusters that do not already collect kubelet-backed container metrics. It scrapes each node's local kubelet for container, pod, and node CPU/memory/network/filesystem metrics and exposes them via Prometheus. The production operator chart does not install this platform-level collector; tenant DocumentDB clusters do not receive kubelet privileges.
```
WentingWu666666: Nit: this seems not major feature because it's a playground example, not part of the operator.

alaye-ms (approved):

xgerman (commented):
## Summary

Net-positive scope reduction for monitoring: removes per-Cluster `spec.monitoring.kubeletstats` API, deletes the operator-managed container-metrics DaemonSet from the production chart, and moves the equivalent collector into `documentdb-playground/telemetry/container-metrics/` as a reference manifest. Tenant DocumentDB pods no longer need kubelet `nodes/stats` privileges.

Operator-side Go changes are small and well-guarded:
- `internal/otel/config.go`: switches resource processor from `upsert` → `insert` (preserves per-datapoint `k8s.*` identity from receivers that emit their own) and enables `resource_to_telemetry_conversion` on the Prometheus exporter so resource attributes become Prometheus labels instead of being buried in `target_info`. Both changes have new regression tests.
- `zz_generated.deepcopy.go`: alphabetical reorder only (controller-gen output).
- `cnpg_sync.go`, `controller/documentdb_controller.go`, `utils/constants.go`, `utils/pv_recovery.go`: pure `gofmt` whitespace fixes (verified: `gofmt -l` on `main` flags exactly these four files).

The bulk of the diff (~2.2K lines) is new playground content + two new public docs pages.

**Local validation:** `go test ./internal/otel/...` ✓, `go vet ./...` ✓, `go build ./...` ✓.

**CI:** 34 success / 1 skipped / 3 failures. All 3 failures are `govulncheck` jobs on modules this PR doesn't touch (newly published `GO-2026-5037/5038/5039` advisories) — **pre-existing, not introduced by this PR**; needs its own dependency-bump PR.

## Critical Issues

None.

## Suggestions

### 🟠 [Suggestion] OTel Collector image drift in the demo path
`documentdb-playground/telemetry/container-metrics/container-metrics.yaml` pins `otel/opentelemetry-collector-contrib:0.114.0`, but `operator/src/internal/utils/constants.go` defines `DEFAULT_OTEL_COLLECTOR_IMAGE = "otel/opentelemetry-collector-contrib:0.149.0"`. After `deploy.sh` runs, the per-pod sidecars are on `0.149` and the reference DaemonSet is on `0.114` in the same kind cluster. Recommend bumping the playground manifest to `0.149.0` (or pulling from a single source-of-truth constant) so the demo doesn't bake in stale image refs.

### 🟠 [Suggestion] Unrelated `gofmt` cleanups bundled in
`cnpg_sync.go`, `documentdb_controller.go`, `utils/constants.go`, `utils/pv_recovery.go` only contain whitespace alignment that `gofmt` would produce. Not blocking, but mixing four unrelated drive-by formatting changes into a monitoring-feature PR makes the change history harder to bisect. Ideally split into a precursor `chore: gofmt` PR next time.

### 🟡 [Minor] `validate.sh` `query()` helper doesn't URL-encode PromQL
Copilot-bot's earlier feedback was not addressed:
```bash
wget -qO- "http://localhost:9090/api/v1/query?query=$1"
```
Works in practice for the specific queries used today (Prometheus tolerates `{`/`}`/`=`/`"` in query params and `wget` doesn't glob), but it will silently corrupt future PromQL containing spaces, `+`, `#`, or `%`. A small `url_encode` helper would future-proof it. Follow-up.

### 🟡 [Minor] CHANGELOG categorisation
The CHANGELOG entry sits under preview release notes as "Playground & Examples". Wenting's earlier nit that this isn't really a major feature still has merit — but the operator chart **also no longer ships `documentdb-container-metrics`**, which *is* a behavior change worth a release note. Consider splitting:
- Release notes: one line noting the chart no longer installs platform-level node collection and platform metrics are user-owned.
- Examples/docs: the playground reference manifest mention.

### 🟡 [Nitpick] No-op refactor in `internal/otel/config.go`
```go
receiverNames := []string{"sqlquery", "otlp"}
```
…has a single use site with a hardcoded literal — adds indirection without adding flexibility. Either inline it back, or wire up the conditional appends that would justify the extraction (e.g. a future spec-driven receiver list).

### 🟡 [Nitpick] Reference DaemonSet has no probes / priority class
Fine for a playground reference, but the README explicitly says "Treat it as a starting point for … platform-owned deployments." A stock `livenessProbe` on the collector's health-check extension (`:13133`) would be a kinder default, or at least a comment in the YAML pointing users at the omission.

## Questions

### 🟡 [Question] `kindest/node:v1.35.0` availability
`setup-kind.sh` defaults `K8S_VERSION="v1.35.0"` and the troubleshooting tip references that exact tag. Please confirm `kindest/node:v1.35.0` is published at the time of merge — kind release cadence sometimes lags upstream Kubernetes by a few weeks.

## Positive Feedback

🟢 The behavior changes in `config.go` are exactly the kind that deserve a regression test, and the test file documents *why* each assertion exists (`Must not contain 'upsert' — that would clobber per-datapoint resource attrs…`). Exemplary.

🟢 The `cluster.yaml` port comment ("CNPG's instance manager uses 9187 for its own /metrics endpoint, so we use 9188 here") is the kind of "why, not what" comment the project benefits from.

🟢 Tight RBAC scope on the reference DaemonSet (`get` on `nodes/stats` + `nodes/proxy` only) and a solid container security context (non-root UID 10001, read-only rootfs, drop ALL capabilities, no privilege escalation).

🟢 Demo manifests carry visible `⚠️ DEMO/PLAYGROUND ONLY` banners — important because the Grafana config (`GF_AUTH_ANONYMOUS_ORG_ROLE: Admin` + plaintext admin password) and the `DemoPassword100` cluster password would otherwise be at risk of drifting into production via copy-paste.

🟢 Earlier review feedback was visibly incorporated (CNPG `cnpg.io/cluster=…` label selectors, correct `documentdb.io/preview` apiVersion, exporter-neutral language in the public docs).

---

**Verdict:** No blockers. Two 🟠 items worth addressing here or in a quick follow-up (OTel image alignment + splitting the gofmt drive-bys); the 🟡/🟢 items are non-blocking. The govulncheck CI failures are unrelated to this PR and need their own dependency-bump PR.

_Review generated by Copilot CLI on behalf of @xgerman; not a binding approval._

xgerman (approved):

PR #354, opened 2026-04-22
Build DocumentDB images from custom gitHub sources

hossain-rayhan:
This is needed for personal testing.

### DIFF
--- a/.github/workflows/build_documentdb_images.yml
+++ b/.github/workflows/build_documentdb_images.yml
@@ -14,13 +14,13 @@ on:
         required: false
         default: '0.110.0'
       documentdb_repo:
-        description: 'GitHub repo for DocumentDB extension releases (owner/repo)'
+        description: 'GitHub repo name for DocumentDB extension releases (repo only, owner derived from current repo)'
         required: false
-        default: 'documentdb/documentdb'
-      gateway_source_repo:
-        description: 'GHCR repo for gateway source image (without ghcr.io/ prefix)'
+        default: 'documentdb'
+      gateway_source_image:
+        description: 'Gateway source image path under owner (e.g., documentdb/documentdb-local)'
         required: false
-        default: 'documentdb/documentdb/documentdb-local'
+        default: 'documentdb/documentdb-local'
 
   repository_dispatch:
     types: [documentdb-release]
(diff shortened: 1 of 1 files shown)


guanzhousongmicrosoft (commented):
the PR is fine, but please set up the github repo so only admin/maintainers can start the build, so approve the build started by external contributors, to protect malicious PR and workflow build

hossain-rayhan:
> the PR is fine, but please set up the github repo so only admin/maintainers can start the build, so approve the build started by external contributors, to protect malicious PR and workflow build

Allowed to modify only the repo/image name, not the owner. Now fork owner can only run workflow with their fork image and same for documentdb official repo.

xgerman (approved):

WentingWu666666 (commented):
**Suggest taking full owner/repo paths as inputs instead of coupling to `github.repository_owner`**

The current design always prepends `${{ github.repository_owner }}` to user-supplied values, so the user can override the **repo name** but never the **owner**. That has a few problems:

1. **The defaults are only correct on the upstream repo.** On any fork  say `acme/documentdb-kubernetes-operator`  the default `documentdb_repo: documentdb` resolves to `acme/documentdb`, which probably doesn''t exist. The input has a "default" that''s effectively wrong everywhere except `documentdb/documentdb-kubernetes-operator`.

2. **Common dev workflows break.** A developer who forks *only* the operator and wants to build/test against the upstream `documentdb/documentdb` extension can''t  they''re forced to also fork `documentdb/documentdb` into their org just to satisfy the owner coupling.

3. **It conflates three independent things into one knob:** the operator repo owner, the extension repo owner, and the gateway image registry path. They aren''t the same concept and shouldn''t be tied together.

4. **The `gateway_source_image` framing is confusing.** It''s described as "path under owner" but its default is `documentdb/documentdb-local` (itself a slash-separated path). When concatenated with owner you get a triple-segment ghcr path like `ghcr.io/owner/documentdb/documentdb-local`, which doesn''t match the "path under owner" mental model.

5. **The security framing for restricting owner is weak.** This is a `workflow_dispatch` workflow  only people with write access to the repo can run it. Letting them point at an arbitrary `org/repo` isn''t meaningfully more dangerous than letting them edit the workflow file.

### Suggested change  take full paths

```yaml
inputs:
  documentdb_release_repo:
    description: ''GitHub owner/repo for DocumentDB extension releases''
    required: false
    default: ''documentdb/documentdb''
  gateway_source_image_repo:
    description: ''Full container image repo for the gateway source image (registry/owner/path)''
    required: false
    default: ''ghcr.io/documentdb/documentdb/documentdb-local''

env:
  DOCUMENTDB_RELEASE_REPO: ${{ github.event.inputs.documentdb_release_repo || ''documentdb/documentdb'' }}
  GATEWAY_SOURCE_IMAGE_REPO: ${{ github.event.inputs.gateway_source_image_repo || ''ghcr.io/documentdb/documentdb/documentdb-local'' }}
```

Benefits:

- Defaults are correct on upstream **and** on any fork
- Forks of just the operator can still build against upstream extension releases
- The input names match what they actually contain (no hidden owner concatenation)
- One knob per independent concept; no implicit coupling to `github.repository_owner`
- The `repository_dispatch` trigger keeps working unchanged because the env defaults handle the case when no inputs are provided

hossain-rayhan:
@guanzhousongmicrosoft / @WentingWu666666, Lets go back to original full path repo/image. Its cleaner. I will work with @xgerman to setup strict rules on who can run the workflow. I don't have admin permission. 
Thanks.

WentingWu666666 (approved):

xgerman (approved):

PR #357, opened 2026-04-22
fix(security): remove Disabled TLS gateway mode to prevent plaintex

### LINKED ISSUE #356: security: gateway serves plaintext Mongo wire protocol with tls.gateway.mode=Disabled, contradicting docs
## Summary

`spec.tls.gateway.mode: Disabled` causes the DocumentDB gateway sidecar to serve the Mongo wire protocol **in plaintext**, contradicting the public docs which promise that TLS is always on. This is a silent footgun — clients who trust `status.connectionString` (which always contains `tls=true`) are safe, but any attacker with on-cluster pod-network access can bypass TLS entirely by connecting with `tls=false`.

Recommend removing `Disabled` from the `GatewayTLS.Mode` enum so unencrypted traffic is impossible by construction.

## Evidence of the contradiction

**Docs promise encryption on all modes** — `docs/operator-public-documentation/preview/configuration/tls.md:43`:

> Disabled mode means the operator does not manage TLS certificates. However, the gateway still encrypts all connections using an internally generated self-signed certificate. Clients must connect with `tls=true&tlsAllowInvalidCertificates=true`.

**E2E test proves plaintext is actually served** — `test/e2e/tests/tls/tls_disabled_test.go:17-19, 41-54`:

```go
// The gateway still listens but accepts plain-text mongo wire
// protocol. This spec verifies the happy-path: a freshly-created
// DocumentDB with TLS disabled accepts an unencrypted connection
// from the mongo driver.
...
client, err := mongohelper.NewClient(connectCtx, mongohelper.ClientOptions{
    Host:     host,
    Port:     port,
    User:     tlsCredentialUser,
    Password: tlsCredentialPassword,
    TLS:      false,
})
...
Eventually(func() error {
    return mongohelper.Ping(connectCtx, client)
}, ...).Should(Succeed(), "plaintext ping should succeed when TLS is disabled")
```

Ping succeeds, so the gateway is genuinely listening without TLS on port 10260.

## Root cause

`operator/cnpg-plugins/sidecar-injector/internal/lifecycle/lifecycle.go:188-232`:

- When `gatewayTLSSecret` parameter is absent (which is what `Mode=Disabled` produces — `operator/src/internal/controller/certificate_controller.go:64-74` short-circuits the cert reconciler and leaves `status.TLS.Ready=false`), the sidecar is started **without** `--cert-path`/`--key-file` CLI args and **without** the `CERT_PATH`/`KEY_FILE`/`TLS_CERT_DIR` env vars.
- The upstream gateway binary then falls back to plaintext, not to self-generated self-signed. The docs' promise is wrong.

## Why this matters

1. **Defense in depth violated.** Mongo credentials travel on the pod network in the clear. Any compromised workload in the same cluster / same VNet can trivially credential-harvest.
2. **Docs encourage the belief that TLS is always on**, so users selecting `Mode=Disabled` for "dev" likely think they're getting a self-signed cert they can skip-verify. They're actually getting plaintext.
3. **`status.connectionString` lies.** It contains `tls=true` unconditionally (`operator/src/internal/utils/util.go:423`) regardless of `Mode=Disabled`. Clients pasting the published string work; clients who read the CR spec do not. The two contracts diverge silently.
4. **SCRAM-SHA-256 authentication over plaintext** still leaks enough for offline brute force given a modern wordlist.

## Proposed fix

**Option A (recommended): remove unencrypted traffic as a possibility.**

1. Drop `Disabled` from the `GatewayTLS.Mode` enum in `operator/src/api/preview/documentdb_types.go:237` — change validation to `Enum=SelfSigned;CertManager;Provided`.
2. Default an unset `Mode` to `SelfSigned` so existing CRs that omit the field keep working (and keep encryption on).
3. Update the cert controller's "empty or Disabled" branch in `operator/src/internal/controller/certificate_controller.go:64` to treat empty as SelfSigned.
4. Remove `test/e2e/tests/tls/tls_disabled_test.go` and `test/e2e/manifests/mixins/tls_disabled.yaml.template`.
5. Remove the "Disabled" tab from `docs/operator-public-documentation/preview/configuration/tls.md`.
6. Note in CHANGELOG as a breaking change for pre-GA; migration path is "remove `mode: Disabled` (or the whole `tls:` block) to get SelfSigned behavior."

**Option B (fallback if Disabled must remain for some out-of-tree user): make it mean what the docs claim.**

1. Have the sidecar injector generate a self-signed cert in-cluster when `Mode=Disabled` (or omitted) and mount it. This is effectively what `Mode=SelfSigned` already does, so at that point Disabled and SelfSigned are synonyms and Option A is cleaner.

Option A is strictly better because it removes the attack surface instead of relying on correct configuration.

## Migration impact

- API change pre-GA (`preview` apiVersion), so the usual "no breaking changes" bar doesn't apply.
- Users with `mode: Disabled` get a behavior change (plaintext → self-signed TLS). Since the documented contract was *already* "TLS is always on", this aligns behavior with the documented contract.

## Out of scope

- Same audit should confirm the Postgres wire protocol inside the pod is TLS'd between the gateway sidecar and the Postgres container. The gateway is launched with `--pg-port 5432` against `127.0.0.1`, so same-pod IPC over loopback — lower risk but worth a follow-up issue.

## Companion doc bug

Independent of this fix, `docs/operator-public-documentation/preview/configuration/tls.md:43` must be corrected to match reality. If Option A is taken, the section is removed entirely.

hossain-rayhan:
## Summary

Fixes #356: gateway no longer serves plaintext Mongo wire protocol. The `Disabled` value for `spec.tls.gateway.mode` has been removed, and TLS is now always enabled — defaulting to operator-managed self-signed certificates whenever `tls`, `tls.gateway`, or `tls.gateway.mode` is unset.

## Problem

`spec.tls.gateway.mode: Disabled` caused the gateway to listen on plaintext, contradicting the docs (which describe Disabled as "self-signed bootstrap"). This left clusters silently insecure when users set what they reasonably believed was a safe value.

### Breaking change — CRD enum
- Removed `Disabled` from `GatewayTLS.Mode` enum in [`documentdb_types.go`](operator/src/api/preview/documentdb_types.go).
- Added `+kubebuilder:default=SelfSigned`.
- Valid values are now: `SelfSigned`, `CertManager`, `Provided`.
- Existing resources with `mode: Disabled` will fail validation on apply — users must set an explicit valid mode or remove the field.

### Controller — fail-safe defaults
- [`certificate_controller.go`](operator/src/internal/controller/certificate_controller.go) now defaults to `SelfSigned` when:
  - `spec.tls` is nil (block omitted entirely),
  - `spec.tls.gateway` is nil, or
  - `spec.tls.gateway.mode` is empty.
- A managed certificate is always provisioned — no code path leaves the gateway plaintext.

### Tests
Added unit tests in [`certificate_controller_test.go`](operator/src/internal/controller/certificate_controller_test.go):
- `TestEmptyModeDefaultsToSelfSigned` — empty string mode → SelfSigned cert created
- `TestEmptyModeWithNilStatus` — empty mode with nil status → SelfSigned cert created
- `TestNilTLSDefaultsToSelfSigned` — `spec.tls` omitted entirely → SelfSigned cert created
- `TestDisabledModeNotSupported` — `Disabled` falls through switch default (defense in depth alongside CRD validation)

All pre-existing TLS tests (`TestSelfSigned*`, `TestProvided*`, `TestCertManager*`) continue to pass.

### Documentation
- [`tls.md`](docs/operator-public-documentation/preview/tls.md) — removed `Disabled` mode references, documented SelfSigned as the default.
- [`connecting-to-documentdb.md`](docs/operator-public-documentation/preview/connecting-to-documentdb.md) — updated connection examples.
- [`api-reference.md`](docs/operator-public-documentation/preview/api-reference.md) — auto-regenerated from CRD.
- [`CHANGELOG.md`](CHANGELOG.md) — breaking change entry under Unreleased.

### Generated artifacts
- `make manifests generate` regenerated CRDs and DeepCopy methods.

### Unit tests
```bash
cd operator/src
go test ./internal/controller/... -run "TestEmptyMode|TestDisabled|TestNilTLS|TestSelfSigned" -count=1 -v
# All pass

Fixes #356

### CODE REVIEW

**Review by xgerman — CHANGES_REQUESTED** [review]
## Review — PR #357 (fix: remove Disabled TLS gateway mode)

Context: I filed #356 calling out that `mode: Disabled` served plaintext traffic despite docs claiming TLS was always active. Thanks for the fast turnaround on this.

The direction is right and the fresh-install case is fully closed. I found **one correctness gap on the upgrade path** that I think needs addressing before merge, plus a few smaller items.

### 🔴 Critical: upgrade path with stored `mode: Disabled` still lands in plaintext

The fix relies on two layers — schema enum + controller switch — but the upgrade path threads between them:

1. CRD validation (`+kubebuilder:validation:Enum=SelfSigned;CertManager;Provided`) rejects new/updated resources with `mode: Disabled`. ✅
2. `reconcileCertificates` maps empty / nil / missing into `SelfSigned`. ✅
3. But when the controller sees a **previously-stored** resource with `mode: Disabled` (apiserver does not re-validate on read), the switch falls through `default:` and returns `nil, nil`.

What happens next:
- `ddb.Status.TLS.Ready` stays `false` (set by the pre-switch "init" block).
- In `operator/src/internal/cnpg/cnpg_cluster.go:87`, `gatewayTLSSecret` is set **only when** `Status.TLS.Ready && SecretName != ""`. It therefore remains unset.
- In `operator/cnpg-plugins/sidecar-injector/internal/lifecycle/lifecycle.go:185,226`, no `gatewayTLSSecret` → no volume mount, no `--cert-path`/`--key-file` args, no `CERT_PATH` env → **gateway sidecar starts plaintext again**, which is the exact bug #356 reported.

This is a real regression path for anyone upgrading from a release that shipped `Disabled` as valid. The CHANGELOG entry says "users must remove this setting", but the operator should fail safe rather than rely on operator action.

**Suggested fix** (the simplest one that matches the PR's stated "no plaintext path" invariant):

```go
switch mode {
case "SelfSigned":
    return r.ensureSelfSignedCert(ctx, ddb)
case "Provided":
    return r.ensureProvidedSecret(ctx, ddb)
case "CertManager":
    return r.ensureCertManagerManagedCert(ctx, ddb)
default:
    // Unknown/legacy mode (e.g. "Disabled" from a pre-#357 resource still in etcd).
    // Fail-safe to SelfSigned so the gateway never serves plaintext.
    logger.Info("unknown TLS mode, defaulting to SelfSigned", "requestedMode", mode)
    r.Recorder.Eventf(ddb, corev1.EventTypeWarning, "LegacyTLSMode",
        "TLS mode %q is no longer supported; provisioning a self-signed certificate. Update spec.tls.gateway.mode to SelfSigned/CertManager/Provided.", mode)
    return r.ensureSelfSignedCert(ctx, ddb)
}
```

Then flip `TestDisabledModeNotSupported` from "asserts no cert is created" to "asserts a SelfSigned cert *is* created for legacy Disabled values". The current test actually encodes the regression as expected behavior — worth calling out.

### 🟠 Major: no e2e proof that the gateway actually refuses plaintext

All four new tests sit at the certificate-reconciler layer. They verify a `cmapi.Certificate` object is created, not that the gateway listens only on TLS. The whole point of #356 was an end-to-end behavior gap. Suggest adding one of:

- A test in `test/e2e/tests/tls/` that deploys with default settings and asserts a plaintext `mongosh --tls=false` (or raw TCP handshake) is rejected/timed out.
- Or at minimum a sidecar-injector unit test that asserts: for a DocumentDB with `Status.TLS.Ready=false` + stored `mode: Disabled`, the injected sidecar has `CERT_PATH` env / `--cert-path` args (given the fail-safe above).

Without this, the next time someone edits this code path the same regression can reappear silently.

### 🟡 Minor

**M1. `TestDisabledModeNotSupported` assertion style**
```go
require.True(t, err != nil, "...")
```
should be `require.Error(t, err)` (or, more precisely, `require.True(t, errors.IsNotFound(err))` so a *different* error doesn't falsely pass).

**M2. `zz_generated.deepcopy.go` reorder noise**
The diff includes a re-ordering of `PrometheusExporterSpec` / `PVRecoveryConfiguration` / `PostgresTLS` deepcopy blocks that isn't related to the TLS change — looks like `make generate` was run with a slightly different controller-gen. Not harmful, but adds merge-conflict surface. Consider bumping controller-gen in a separate PR or reverting the ordering hunk.

**M3. CHANGELOG wording slightly off**
"contradicting documentation that stated TLS was always active" — the *old* docs tab for Disabled actually said the gateway *still encrypts using an internally generated self-signed certificate*, which is what users were relying on. That's a self-inconsistency inside the docs (tab promised TLS, implementation gave plaintext), not docs saying one thing everywhere. Consider: "…contradicting the `Disabled` tab in `tls.md`, which described the mode as a self-signed bootstrap."

**M4. Test comment is inaccurate**
`TestDisabledModeNotSupported` comment line 223: "In production, the webhook validation prevents this invalid value." There's no webhook involved — it's the CRD's built-in `enum` validation at admission time. Update the comment to match the mechanism.

### 🟢 Docs quality (Microsoft Writing Style Guide)

`tls.md` is good overall — front matter is complete, sentence-case works for most headings, and the new line 16 clarification ("always encrypts connections — TLS is active regardless of the mode you choose") directly addresses the confusion that caused #356. A few items to tighten:

- **Heading case**: `## Certificate Rotation`, `### Monitoring Certificate Expiration`, `## Additional Resources` → use sentence case ("Certificate rotation", "Monitor certificate expiration", "Additional resources") per MS Style.
- **"e.g."** on line 149 annotation ("(e.g., `my-ca-issuer` from the prerequisite example above)") → use "for example" for parity with line 119 which already uses "(for example, a corporate CA)".
- Consider adding a short migration admonition at the top of `tls.md`:
  ```markdown
  !!! warning "Breaking change in this release"
      The `Disabled` TLS mode has been removed. If you previously set `spec.tls.gateway.mode: Disabled`, update it to `SelfSigned` (or remove the field — `SelfSigned` is now the default). See [CHANGELOG](../../../CHANGELOG.md) for details.
  ```
  Makes the breaking change discoverable from the docs, not just the changelog.
- `api-reference.md`: looks correct and auto-regenerated — default `SelfSigned`, enum trimmed. ✅

### Positive

- CRD-level enforcement (enum trimmed + `+kubebuilder:default=SelfSigned`) is the right control for new resources.
- Nil-safe defaulting covers the three unset cases (`spec.tls` nil, `gateway` nil, `mode` empty) with a single normalized `mode` local — clean, easy to read.
- Three separate tests for the unset paths (`TestEmptyMode…`, `TestNilTLS…`) — good coverage of the happy-path.
- CHANGELOG entry is under `Breaking Changes` (correct) and links to the originating issue.
- API reference regeneration committed in the same PR — follows the generator-artifact rule.

### Verdict

`/request-changes` — the 🔴 upgrade-path fall-through is the blocker. Once the `default:` case fail-safes to `SelfSigned` (and the test flips to assert that), this becomes a clean merge candidate. The 🟠/🟡/🟢 items are follow-ups.

xgerman (approved):

alaye-ms (approved):

PR #358, opened 2026-04-24
Add testing guidelines with fork repos

hossain-rayhan:


WentingWu666666 (changes requested):
Thanks for adding this guide -- the Helm install/CR sections are accurate and useful. Two blockers and a few small items below.

### :x: 1. Step 1 inputs depend on PR #354 (currently unmerged)

The doc instructs users to set `documentdb_extension_github_repo` and `documentdb_gateway_image_repo` when dispatching `build_documentdb_images.yml`, but on `main` that workflow only declares one input (`version`). Those two inputs are added by **#354** ("Build DocumentDB images from custom gitHub sources"), which is still open.

As written, anyone following this doc today will get a workflow-dispatch error. Please either:
- Land #354 first and rebase this PR on top, or
- Merge them together,

so the doc is correct the moment it ships. A "requires #354" note in the doc itself ages badly.

### :x: 2. The bare `:<version>` tag is not produced by `build_documentdb_images.yml`

Step 1.4 says:
> *"The workflow also retags it with the bare `<version>` for convenience; check the run summary."*

That is not what the workflow does. `build_documentdb_images.yml` only publishes the candidate tag `<version>-build-<run_id>-<attempt>-<sha>` (and per-arch `-amd64` / `-arm64` variants). Its own summary explicitly directs the user to run `release_documentdb_images.yml` next:

```
- candidate_version: 0.110.0-build-12345-1-abc1234
```

The bare-version retag actually happens in `release_documentdb_images.yml`:

```yaml
- name: Retag existing manifest
  env:
    SOURCE_TAG: ${{ inputs.candidate_version }}
    TARGET_TAG: ${{ inputs.version }}
```

So after only running the build workflow, `ghcr.io/<fork>/documentdb-kubernetes-operator/documentdb:0.110.0` does **not** exist on the user's fork -- and the example in Step 3 will fail to pull.

Please pick one fix:
- **(A) Use the long candidate tag in the example.** Replace the `:0.110.0` references in Step 3's CR snippet with the long candidate tag the build workflow actually produces, and explain how to read it from the run summary.
- **(B) Add a "Step 1.5: Promote candidate to release tag"** that runs `release_documentdb_images.yml` with `candidate_version=<long tag>` and `version=0.110.0`, then keep the `:0.110.0` examples.

(B) is closer to how the operator track works and matches what the build workflow's summary tells users to do.

### :warning: Minor

- **Private GHCR + imagePullSecrets**: the doc explains how to make the fork's packages public, but if a reviewer skips that step, the install will silently fail at pod startup. Either add an `imagePullSecrets` example for the `documentdb-operator-system` namespace, or call out that the packages **must** be public for the helm install to succeed.
- **cert-manager `v1.13.2+`**: `AGENTS.md`'s version compatibility matrix lists cert-manager `1.19.2` as the version the project tests against. Worth aligning so users don't pick an older version we don't validate.

Happy to re-review once #354 is in and the bare-tag step is reconciled.

WentingWu666666 (approved):

xgerman (approved):

PR #359, opened 2026-04-28
feat: consolidate Helm chart distribution to OCI-only and drop GitHub Pages

### LINKED ISSUE #326: Consider consolidating Helm chart distribution to OCI-only (drop GitHub Pages)
## Summary

Simplify the Helm chart release pipeline by consolidating distribution to OCI-only and dropping the GitHub Pages-based Helm repository.

## Current State

The release workflow (`release_images.yml`) publishes the Helm chart to **two** distribution channels:

1. **OCI registry (GHCR)** via `helm push ... oci://ghcr.io/documentdb/...`
2. **GitHub Pages** via the `repair_helm_pages_release.yml` reusable workflow (publishes to the `gh-pages` branch with `index.yaml`)

## Problem

Maintaining two distribution channels adds complexity:

- `repair_helm_pages_release.yml` is a **300+ line workflow** dedicated to rebuilding/merging `index.yaml` and publishing to `gh-pages`
- The `gh-pages` branch must be maintained alongside the OCI registry
- Any issues with `index.yaml` require manual repair workflows
- OCI is the modern standard for Helm chart distribution (supported since **Helm 3.8**, which is 3+ years old)

## Proposal

- Drop the GitHub Pages-based Helm repository and consolidate to **OCI-only** distribution
- Remove `repair_helm_pages_release.yml` and the `publish-helm-pages` job from `release_images.yml`
- Update documentation (README, install guides) to use OCI-based install commands exclusively:
  `
  helm install documentdb oci://ghcr.io/documentdb/documentdb-operator --version <version>
  `

## Tradeoff

- Users lose `helm repo add` + `helm search repo` discoverability (must know the OCI URL)
- Requires Helm 3.8+ (released March 2022, widely adopted)
- Significantly simplifies the release pipeline

## Note

The OCI URL and install command should be prominently documented in the README to compensate for the loss of `helm search` discoverability.

udsmicrosoft:
Closes #326.

## Summary

Removes the GitHub Pages-based Helm repository (`https://documentdb.github.io/documentdb-kubernetes-operator`) and consolidates Helm chart distribution to the OCI registry already published to GHCR (`oci://ghcr.io/documentdb/documentdb-operator`). Updates workflows, actions, scripts, and docs that still consumed the chart from GH Pages.

### Release pipeline (no more GH Pages publish)
- Deleted `.github/workflows/repair_helm_pages_release.yml` (the 300+ line reusable workflow that rebuilt `index.yaml` and committed it to the `gh-pages` branch).
- Removed the `publish-helm-pages` job from `release_operator.yml` and from the deprecated `release_images.yml`. OCI publication via `helm push` is unchanged.

### CI consumers migrated to OCI
- `.github/actions/setup-test-environment/action.yml`: the `released-chart-version` install path now runs `helm install … oci://ghcr.io/documentdb/documentdb-operator --version <X>`.
- `.github/workflows/test-upgrade-and-rollback.yml`: dropped `helm repo add documentdb` / `helm repo update`.
- "latest" version resolution: OCI registries don't support `helm search repo`, so both call sites now resolve `latest` by **listing GHCR tags directly** (`/v2/.../tags/list`) and picking the highest stable SemVer. This better matches the prior `helm search repo … | .[0].version` semantics than `gh api releases/latest` would.
- GHCR Helm registry login is best-effort for public chart installs, and GHCR tag-list requests only send an Authorization header when a token was resolved.

### Playground scripts
- `documentdb-playground/aws-setup/scripts/create-cluster.sh`: collapsed the "try GH Pages → fall back to OCI" structure into a single OCI install. `GITHUB_USERNAME`/`GITHUB_TOKEN` are optional for authenticated GHCR pulls.
- `documentdb-playground/aks-fleet-deployment/install-documentdb-operator.sh`: replaced GH Pages install with OCI; requires explicit `CHART_VERSION` env var when not building a local chart.

### Docs
- Quickstart-style install snippets now use the OCI chart without hardcoding `DOCUMENTDB_VERSION=0.2.0`:
  ```bash
  helm install documentdb-operator oci://ghcr.io/documentdb/documentdb-operator \
    --namespace documentdb-operator --create-namespace --wait
  ```
- Production/repeatable-install guidance tells users to add `--version <release-version>` from GitHub Releases.
- Upgrade and production-oriented docs use `<release-version>` placeholders instead of literal release numbers.
- `.github/workflows/deploy_docs.yml` stamps `<release-version>` with the GitHub release tag before `mike deploy`, so versioned docs snapshots do not require per-release manual doc bumps.
- Files updated: `README.md`, `index.md`, `getting-started/quickstart-kind.md`, `getting-started/quickstart-k3s.md`, `getting-started/deploy-on-eks.md`, `multi-region-deployment/setup.md`, `operations/upgrades.md`, `faq.md`, `documentdb-playground/aws-setup/README.md`.
- Stale `documentdb.github.io/...` link in `aws-setup/README.md` repointed to `documentdb.io/...`.

### Root README
- New `�� Install` section between `📚 Documentation` and `🚀 Quick Start`, with a copy-paste OCI install snippet, version-pinning guidance, link to GitHub Releases for version discovery, and the Helm 3.8+ requirement.

## Tradeoffs (per the issue)

- Users lose `helm repo add` + `helm search repo` discoverability — must consult [GitHub Releases](https://github.com/documentdb/documentdb-kubernetes-operator/releases) for available versions.
- Requires Helm 3.8+ (March 2022) for OCI support — already widely adopted.
- Significantly simpler release pipeline; no more `gh-pages` Helm-repo branch maintenance or `index.yaml` repair workflow.

## Validation

- `yaml.safe_load` parses all modified workflows / composite action.
- `bash -n` parses both modified shell scripts.
- Final grep sweep for `helm repo add documentdb`, `documentdb.github.io/documentdb-kubernetes-operator`, `helm search repo documentdb`, `helm repo update documentdb`, `repair_helm_pages`, `publish-helm-pages` returns zero hits across the repo.
- Grep sweep for hardcoded `DOCUMENTDB_VERSION=0.2.0` / `TARGET_VERSION=0.2.0` returns zero hits in the public docs and README.
- Smoke-tested the GHCR tag-list resolver locally — returns `0.2.0`.
- Confirmed `oci://ghcr.io/documentdb/documentdb-operator` is anonymously pullable with an isolated Helm registry config, so the quickstart snippets work without `helm registry login`.
- Smoke-tested the docs release-stamping command against a temporary copy of `docs/operator-public-documentation`.

## Out of scope (intentionally untouched)

- The `gh-pages` branch used by `deploy_docs.yml` for **mike-based docs hosting** is left alone — it serves a different purpose than the deprecated Helm repository and is not what issue #326 targets.


alaye-ms (approved):

xgerman (approved):

WentingWu666666:
**Question on docs versioning + chart pin: avoid per-release manual bumps?**

The PR pins `DOCUMENTDB_VERSION=0.2.0` literally in several quickstart/operations pages. That works today, but combined with `mike` (which snapshots `docs/` into `/{release}/` on every release) it means **every release needs a docs PR** to bump the literal -- otherwise the `/0.3.0/` snapshot will still tell users to install `0.2.0`, and `/latest/` (which aliases the new release) will be wrong.

Three patterns that get the same "docs match deployed chart" property without the per-release toil:

**1. Placeholder + release-time substitution (lowest churn).** Use a sentinel in markdown:

```bash
DOCUMENTDB_VERSION=__DOCUMENTDB_VERSION__
helm install documentdb-operator oci://ghcr.io/documentdb/documentdb-operator \
  --version ${DOCUMENTDB_VERSION} ...
```

And in `deploy_docs.yml` before `mike deploy ${VERSION}`:

```yaml
- name: Stamp release version into docs
  if: github.event_name == 'release'
  run: |
    VERSION=${{ github.event.release.tag_name }}
    find docs -name '*.md' -exec sed -i "s/__DOCUMENTDB_VERSION__/${VERSION}/g" {} +
```

Main branch source stays evergreen, every release snapshot is correct, zero release-time work.

**2. Drop the pin from quickstarts entirely.** OCI charts support `helm install ... oci://...` without `--version`; Helm picks the highest non-prerelease SemVer tag. Argo CD and cert-manager both use this pattern in their getting-started pages, with a callout box for production:

> **Production:** pin `--version` to a specific release for upgrade safety. See the release notes for available versions.

The version-snapshot problem then disappears -- `/0.3.0/`'s quickstart says "install latest," which is `0.3.0` at the time of that snapshot. Self-consistent. The CRD-coupling concern that motivates pinning is really an *upgrade* concern, not a *fresh install* concern, so quickstarts can lean toward "just give me the latest."

**3. Hybrid (probably the right answer here).**
- **Quickstart pages** -- drop the pin (pattern 2). Add a note pointing to the upgrade/operations page for production.
- **Operations / upgrade pages** -- keep an explicit `--version` example, using the placeholder pattern (pattern 1) so each snapshot stamps itself.
- **`documentdb-playground/*.sh`** -- keep these pinned to a literal version. They're branch-versioned by git, not mike-versioned, and reproducibility of example scripts matters more than freshness.

This way: zero per-release docs bumps, each snapshot is internally consistent, new users get the lowest-friction install, and production users get the explicit pinning guidance where it actually matters.

Verified locally that `helm install ... oci://ghcr.io/documentdb/documentdb-operator` with no `--version` resolves to `0.2.0` (highest non-prerelease tag; correctly skipped `0.1.2-test` and `0.1.3-test`).

Happy to send a follow-up PR with the placeholder + sed change if there's interest in pattern 1 or 3.

udsmicrosoft:
Addressed in `54394ec` using the hybrid approach you suggested:

- Quickstart-style install snippets no longer hardcode `DOCUMENTDB_VERSION=0.2.0`; they install the latest stable OCI chart and include a tip to add `--version <release-version>` for production/repeatable installs.
- Upgrade / production-oriented examples now use `<release-version>` instead of a literal version.
- `deploy_docs.yml` stamps `<release-version>` with the GitHub release tag before `mike deploy`, so release snapshots do not require a manual docs bump for the chart version.
- Kept the playground scripts pinned/explicit as branch-versioned examples.

I also updated the PR description to reflect this so it no longer says every snippet is pinned to a literal version.

WentingWu666666 (approved):

WentingWu666666 (approved):

xgerman:
Summary

  This PR consolidates Helm chart distribution from GitHub Pages + OCI to OCI-only (oci://ghcr.io/documentdb/documentdb-operator), closes #326. It's a well-structured cleanup: deletes 348 lines of the repair_helm_pages_release.yml workflow, simplifies the release pipeline, and updates all consuming workflows/docs/scripts consistently.

  Overall assessment: ✅ Approve with minor suggestions

  Critical Issues

  None. The changes are coherent and all CI passes.

  🟠 Major — Worth addressing

   1. GHCR tag-list resolution is fragile for future pre-release tags
   In setup-test-environment/action.yml, the "latest" resolver filters with grep -E '^[0-9]+\.[0-9]+\.[0-9]+$' — this correctly excludes pre-release tags (e.g. 0.3.0-rc.1), but if someone publishes 0.3.0 then 0.3.0-post1, the sort -V | tail -1 would pick 0.3.0-post1 which might not exist as a chart. Consider documenting that only strict X.Y.Z tags should be pushed, or add a helm show chart probe to validate before installing.
   2. test-upgrade-and-rollback.yml OCI chart path fix
   Line in test-e2e.yml changes from oci://ghcr.io/documentdb/charts/documentdb-operator to oci://ghcr.io/documentdb/documentdb-operator. Confirm this is the correct OCI path — the existing main branch uses /charts/ in some places. If the chart was historically pushed to the /charts/ path, older versions won't be installable at the new path for upgrade tests.

  🟡 Minor — Suggestions

   1. deploy_docs.yml indentation changed from 2-space to 6-space — looks like an inadvertent reformat. The YAML is valid but the style differs from the rest of the file. [Nitpick]
   2. install-documentdb-operator.sh now hard-fails without CHART_VERSION — this is a breaking change for anyone who previously ran the script without it. The error message is clear, but consider defaulting to "latest" with the same GHCR tag-list resolution used elsewhere, for parity. [Suggestion]
   3. <release-version> placeholder in docs — The deploy_docs.yml stamps this with sed at release time, which is clever. But if someone reads the docs from main branch (dev docs) they'll see literal <release-version>. Consider using a value like latest or 0.x.y with a comment "(replace with your version)" for the dev/preview docs path. [Suggestion]

  Questions

   - Is the gh-pages branch going to be cleaned up / deleted in a follow-up? The PR description mentions the branch stays for mike-based docs, which makes sense — just confirming it's intentional.
   - The aks-fleet-deployment script now requires CHART_VERSION — was this previously defaulting to something? If existing CI/automation calls this script, they'll break. Is there an upstream consumer?

  🟢 Positive Feedback

   - Excellent PR description with validation steps, tradeoff analysis, and explicit out-of-scope boundaries.
   - The grep sweep verification gives confidence nothing was missed.
   - Clean separation: OCI publication is unchanged, only the Pages consumer path is removed.
   - The !!! tip "Pin production installs" admonitions are a good UX pattern for quickstart docs.
   - Error messages throughout are actionable — they tell users exactly what to set and where to look.

udsmicrosoft:
Thanks for the careful read @xgerman! Going through each point:

### Major

**1. GHCR tag-list resolution / pre-release tags** — I think this is already covered. The regex `^[0-9]+\.[0-9]+\.[0-9]+$` is anchored on both ends, so `0.3.0-post1`, `0.3.0-rc.1`, `0.1.2-test`, etc. don't match and get dropped before `sort -V`. You can see it working today — `0.1.2-test` and `0.1.3-test` are already published to GHCR alongside the stable tags, and `latest` resolves cleanly to `0.2.0`:

```console
$ curl -sSL -H "Authorization: Bearer $TOK" \
    https://ghcr.io/v2/documentdb/documentdb-operator/tags/list | jq '.tags'
["0.1.1","0.1.2-test","0.1.2","0.1.3-test","0.1.3","0.2.0"]
```

A `helm show chart` probe would be belt-and-suspenders, but the source of truth here *is* the GHCR tag list, so I'd rather not add a second round-trip per CI run.

**2. OCI chart path (`/charts/` vs no `/charts/`)** — Confirmed the new path is correct, and the old `/charts/` path never existed:

```console
$ curl -sSL https://ghcr.io/v2/documentdb/charts/documentdb-operator/tags/list
{"errors":[{"code":"NAME_UNKNOWN","message":"repository name not known to registry"}]}
```

`release_operator.yml` pushes via `helm push "$CHART_FILE" oci://${GHCR_REPO}` where `GHCR_REPO=ghcr.io/${{ github.repository_owner }}` (so `oci://ghcr.io/documentdb`), and `helm push` appends the chart name from `Chart.yaml`. Every historical release (0.1.1, 0.1.2, 0.1.3, 0.2.0) is installable from `oci://ghcr.io/documentdb/documentdb-operator`, so the upgrade-test path works for all prior versions.

### Minor

**1. `deploy_docs.yml` indentation** — Not a reformat — the workflow was rewritten end-to-end (mkdocs + helm-pages publisher → `mike`-based versioned docs). Every step is new, so the indentation is just what the new file uses, not a drift from the old one.

**2. `install-documentdb-operator.sh` hard-failing without `CHART_VERSION`** — Intentional. The default path (`BUILD_CHART=true`) is unchanged; the new requirement only kicks in on the explicit `BUILD_CHART=false` OCI install branch. For a fleet-deployment script I'd rather fail loudly than silently resolve `latest` and have an upgrade roll out somewhere unexpected. The `setup-test-environment` action keeps the "latest" resolver because it's nightly CI verification, which is the opposite trade-off.

**3. `<release-version>` placeholder in docs** — I'd lean on the `<…>` convention here. The surrounding prose always pairs it with a Releases link ("…add `--version <release-version>` using a version from <https://github.com/.../releases>"), and pasting the literal into a shell would fail loudly because `<` is a redirect. Replacing it with `latest`/`0.x.y` in dev docs felt like it could mislead readers into thinking that's the actual current version.

### Questions

- **gh-pages branch** — Yes, intentional. `mike` uses gh-pages as its publishing branch, so the branch stays but its contents are now mike-managed versioned docs rather than the old per-commit mkdocs build + helm repo index.
- **aks-fleet script consumer** — No upstream automation calls this script with `BUILD_CHART=false` today; the default (`BUILD_CHART=true`) builds and installs the local chart and is unchanged. The new `CHART_VERSION` requirement only applies on the OCI branch.

Thanks again for the review!

PR #364, opened 2026-05-05
chore: bump default DocumentDB version to 0.110.0

WentingWu666666:
## Summary
Bumps the default DocumentDB database image version (extension + gateway) from **0.109.0  0.110.0** across the Helm chart, operator constants, sidecar-injector config, CI workflows, design docs, and tests.

### 1. Default version bump
- `operator/documentdb-helm-chart/values.yaml`  `documentDbVersion: 0.110.0`
- `operator/src/internal/utils/constants.go` and sidecar-injector `internal/config` defaults updated
- Test fixtures, design docs, and operator-track examples aligned

### 2. CI: pin documentdb build to the chart''s pinned version
`test-build-and-package.yml` previously built whatever upstream `main` had at run time, which caused drift between what the chart advertised and what the e2e suite actually exercised. The build is now pinned to `documentDbVersion` from the chart values, so CI always tests the version the chart will ship.

### 3. CI: skip the upstream source build when the registry already has it
New `probe-images` job queries GitHub Packages for the documentdb + gateway images at `documentDbVersion`. Two execution modes:
- **registry mode** (both images already published): downstream jobs `docker pull` and re-tag  ~15s per image. `build-packages` skipped entirely.
- **build mode** (either image missing, or caller passed `inputs.documentdb_ref`): falls back to building from upstream source as before.

The probe uses `gh api /orgs/{owner}/packages/container/.../versions` (with `/users/...` fallback) authenticated by the workflow''s built-in `GITHUB_TOKEN`  no anonymous registry token dance, no extra secrets.

Net effect for typical PRs that don''t change the version: documentdb/gateway image jobs drop from ~25 min to ~15 s, and `build-packages` is skipped (saving another big chunk).

### 4. Govulncheck / build-toolchain fixes
- `golang.org/x/net` bumped to `v0.53.0` to clear `GO-2025-3956`
- Go directive bumped to `1.25.10` across operator/src, sidecar-injector, kubectl plugin, and test/e2e modules; toolchain workaround dropped
- Builder images allowed to auto-download the Go toolchain (was blocked by the MS golang base image config)

### 5. Gateway package build aligned with current upstream layout
- Gateway packages now built with PG 18 (`--pg 18`)
- Workflow handles both old and new `build_gateway_packages.sh` script paths

## Verification
- Unit tests: `go test` on `internal/utils`, `internal/cnpg`, sidecar-injector `internal/config`  all pass
- All 38 CI checks pass on the latest run
- `Build DocumentDB Packages` skipped (correct), `Build DocumentDB/Gateway Images` finished in 15-17 s via registry pull, all 16 E2E suites pass on amd64 and arm64


alaye-ms (approved):

guanzhousongmicrosoft (approved):
better to have a config file so only need to update one place

xgerman (approved):

hossain-rayhan:
Maybe we should release v0.2.1 as we are updating core component default versions.

xgerman:
@copilot resolve the merge conflicts in this pull request

WentingWu666666:
**Update:** the first attempt (`DEB_BUILD_OPTIONS=nocheck` env var on the job) didn't work  upstream's `debian/rules override_dh_auto_test` unconditionally runs `make check` and ignores `nocheck`. The build also runs inside a Docker container that doesn't inherit host env, so the var was doubly useless.

The new fix in commit `50b4169` rewrites `packaging/deb/common/rules` after checkout (before `build_packages.sh`) to neutralize the `override_dh_auto_test` target. Verified working: both `Build DocumentDB Packages (amd64)` and `(arm64)` jobs now pass on commit `50b4169` (previously both failed at the `check-basic` regression suite on commit `678663c`).

Original failures: [run 25511078036](https://github.com/documentdb/documentdb-kubernetes-operator/actions/runs/25511078036/job/74869696017?pr=364), [run 25511375099](https://github.com/documentdb/documentdb-kubernetes-operator/actions/runs/25511375099/job/74870747221?pr=364)  both red on `setUnion` / `setDifference` projection diffs caused by missing `ORDER BY` in upstream `pg_documentdb/src/test/regress/expected/*.out` files.

@team please review:
- The skip is **scoped to the operator-repo CI build only**  it does not change what gets shipped or what runs in the upstream documentdb repo's own CI.
- The runtime correctness of the produced `.deb` (extension + gateway) is still validated end-to-end by the `TEST - E2E` workflow, which provisions a kind cluster and exercises the operator + DocumentDB + gateway path against a real Mongo client.
- We will need to file an issue against `documentdb/documentdb` to add `ORDER BY` to those regress queries (or otherwise stabilize the output ordering across ICU/planner versions) so we can re-enable the check here. I'll open that follow-up issue once this PR merges.

Anything else you want me to add to this guard (e.g. echo a clearer banner, tag the skip in the build artifact name, or only skip on PR builds and keep it on for nightly main)?

guanzhousongmicrosoft (approved):

Issue #365, opened 2026-05-06
deploy.sh: pinned Istio 1.24.0 is EOL and breaks on Kubernetes 1.31+ (image volume rejection)


markjbrown:
## Summary

`documentdb-playground/multi-cloud-deployment/deploy.sh` currently pins:

```bash
ISTIO_VERSION="1.24.0"
```

This release is **end-of-life** ([Istio supported releases](https://istio.io/latest/docs/releases/supported-releases/)) and its bundled Helm charts no longer render cleanly on current `istioctl` versions. A fresh `istioctl install` against a Kubernetes 1.33+ cluster fails with:

```
WARNING: Istio 1.24.0 may be out of support (EOL) already
Error: generate config: helm render: load chart: component does not exist
```

In addition, the Istio 1.23.x sidecar webhook (which is what gets installed if a user already has `istioctl` on their PATH from before this script was written) does not understand the Kubernetes [OCI image volume source](https://kubernetes.io/docs/concepts/storage/volumes/#image) (alpha 1.31, beta 1.33, **GA 1.35**). When the DocumentDB cnpg-i sidecar plugin attaches the extension binaries via an `image:` volume, the older Istio injector mutates the spec into something the API server rejects:

```
spec.volumes[3].image: Forbidden: may not specify more than 1 volume type
spec.containers[1].volumeMounts[5].name: Not found: "istio-envoy"
```

This silently breaks the whole multi-cloud deployment on K8s 1.31+ ΓÇö CNPG sits in `Setting up primary` forever because init/join Job pods never get created, and the operator log just loops on `Selected PVC is not ready yet`.

## Repro

1. Provision a multi-cluster setup on Kubernetes 1.33 or newer (e.g. AKS + EKS, both on K8s 1.35).
2. Run `deploy.sh` end-to-end.
3. Apply `documentdb-cluster.yaml`.
4. Observe that `kubectl get documentdb` stays in `Setting up primary` indefinitely; the underlying `cluster.postgresql.cnpg.io` reaches `Cluster is unrecoverable`; events on the cluster show the `Forbidden: may not specify more than 1 volume type` rejection.

## Suggested fix

[...]

PR #366, opened 2026-05-06
docs(multi-cloud): troubleshoot Istio sidecar webhook + K8s 1.31+ image volume rejection

markjbrown:
## What

Adds a troubleshooting entry to `documentdb-playground/multi-cloud-deployment/README.md` for a failure mode that silently breaks the multi-cloud deployment on Kubernetes 1.31+: the Istio sidecar webhook on older releases (`<= 1.23.x`) does not understand the Kubernetes [OCI image volume source](https://kubernetes.io/docs/concepts/storage/volumes/#image), and produces a malformed pod spec when it mutates the init/join Job pods that the DocumentDB cnpg-i sidecar plugin attaches `image:` volumes to.

The user-visible symptom is that `kubectl get documentdb` sits in `Setting up primary` forever and the operator log loops on `Selected PVC is not ready yet`. The actual API-server rejection is on the CNPG cluster events:

```
spec.volumes[3].image: Forbidden: may not specify more than 1 volume type
spec.containers[1].volumeMounts[5].name: Not found: "istio-envoy"
```

This is non-obvious because none of the symptoms point at Istio.

The new section:

- Names the symptom and the underlying cause
- **Explicitly warns against** "disable Istio injection on the data namespace" as a workaround — Istio is required for `crossCloudNetworkingStrategy: Istio`, the demo's whole cross-cloud transport
- Documents the fix: install a current `istioctl`, re-apply the canonical `IstioOperator` spec on each member, restart east-west gateways, recreate the CNPG cluster CR
- Notes that `deploy.sh`'s `ISTIO_VERSION=1.24.0` pin is now EOL and its bundled charts fail to render with `helm render: load chart: component does not exist`

Verified end-to-end on AKS 1.35 (eastus2) + EKS 1.35 (us-west-2) with Istio 1.29.2 — both clusters reach `Cluster in healthy state` after applying the documented procedure.

## Why this PR (and not just bumping `deploy.sh`)

Filed as a separate issue for the `deploy.sh` bump: #365. Keeping that change separate because it touches the install path and probably wants more eyes / its own version-compat conversation, whereas this is purely additive documentation that helps anyone hitting the symptom today.

## Test plan

- Reproduced the failure on two clusters (AKS + EKS, both K8s 1.35) running Istio 1.23.4
- Followed the documented fix verbatim → both clusters healthy, cross-cloud replication working
- Working tree contains only the README diff (49 added lines)

### DIFF
--- a/documentdb-playground/multi-cloud-deployment/README.md
+++ b/documentdb-playground/multi-cloud-deployment/README.md
@@ -485,9 +485,15 @@ ISTIO_VERSION=1.29.2
 curl -L https://istio.io/downloadIstio | ISTIO_VERSION=$ISTIO_VERSION sh -
 export PATH="$PWD/istio-$ISTIO_VERSION/bin:$PATH"
 
-# 2) For each member cluster, re-apply the canonical IstioOperator spec
-#    (matches the spec installed by deploy.sh, with the new control plane version)
-for cluster in azure-documentdb aws-documentdb; do
+# 2) For *each* member cluster, re-apply the canonical IstioOperator spec
+#    (matches the spec installed by deploy.sh, with the new control plane version).
+#    Replace MEMBER_CLUSTERS with the kubectl contexts for ALL of your member
+#    clusters — for the default 3-cloud demo this is azure-documentdb,
+#    gcp-documentdb, and aws-documentdb. Leaving any member on the stale
+#    sidecar-injector will keep the failure mode on that cluster.
+MEMBER_CLUSTERS=(azure-documentdb gcp-documentdb aws-documentdb)
+index=0
+for cluster in "${MEMBER_CLUSTERS[@]}"; do
   index=$(( index + 1 ))
   istioctl --context "$cluster" x precheck
   cat <<EOF | istioctl --context "$cluster" install -y -f -
@@ -505,8 +511,14 @@ EOF
   kubectl --context "$cluster" -n istio-system rollout restart deploy istio-eastwestgateway
 done
 
-# 4) Recreate the CNPG cluster CR (operator will recreate it from the DocumentDB CR)
-kubectl --context azure-documentdb -n documentdb-preview-ns delete cluster.postgresql.cnpg.io --all
+# 4) Recreate the CNPG cluster CR on the *primary* cluster (the operator will
+#    then recreate it from the DocumentDB CR). Discover which member is
+#    primary from the DocumentDB spec rather than hard-coding it:
+PRIMARY_CLUSTER=$(kubectl --context "$HUB_CONTEXT" -n documentdb-preview-ns \
+  get documentdb documentdb-preview \
+  -o jsonpath='{.spec.clusterReplication.primary}')
+kubectl --context "$PRIMARY_CLUSTER" -n documentdb-preview-ns \
+  delete cluster.postgresql.cnpg.io --all
 ```
 
 **Verification:** Within ~30s the operator recreates the CNPG `Cluster`. The `*-initdb` Job pod should now reach `Completed`, the primary pod should run `3/3`, and the cross-cloud replica's `*-join` Job should follow shortly. `kubectl get documentdb.documentdb.io` will report `Cluster in healthy state`.
--- a/documentdb-playground/multi-cloud-deployment/README.md
+++ b/documentdb-playground/multi-cloud-deployment/README.md
@@ -464,6 +464,55 @@ kubectl --context <cluster-name> get svc -n istio-system istio-eastwestgateway
 kubectl --context <cluster-name> get secrets -n istio-system | grep istio-remote-secret
 ```
 
+### CNPG init/join Job pods rejected: `volumes[*].image: Forbidden: may not specify more than 1 volume type`
+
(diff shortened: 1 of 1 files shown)


markjbrown:
Thanks for the review — both points were spot on. Pushed `53cbf9c` addressing them:

1. **Member clusters:** the loop now uses an explicit `MEMBER_CLUSTERS=(azure-documentdb gcp-documentdb aws-documentdb)` array matching the default 3-cloud demo, with a comment calling out that skipping a member would leave the failure mode in place on that cluster.
2. **Primary discovery:** the delete step now discovers the primary from `.spec.clusterReplication.primary` on the DocumentDB CR (verified the field name against the live operator) instead of hard-coding `azure-documentdb`, so it targets the correct cluster regardless of which member is currently primary or if the CR has been failed over.

Ready for another look.

documentdb-playground/multi-cloud-deployment/README.md
```diff
@@ -464,6 +464,67 @@ kubectl --context <cluster-name> get svc -n istio-system istio-eastwestgateway
 kubectl --context <cluster-name> get secrets -n istio-system | grep istio-remote-secret
 ```
 
+### CNPG init/join Job pods rejected: `volumes[*].image: Forbidden: may not specify more than 1 volume type`
```
hossain-rayhan, line 467: This should go in the prerequisite section rather than like a troubleshooting step.

alaye-ms (approved):

xgerman:
🟠 [Required] Avoid --all when deleting the CNPG Cluster

  In the new remediation snippet:

   kubectl --context "$PRIMARY_CLUSTER" -n documentdb-preview-ns \
     delete cluster.postgresql.cnpg.io --all

  This is too broad for troubleshooting documentation. If a reader has more than one CNPG Cluster in documentdb-preview-ns, this deletes all of them. It
  also makes the “fix” look more destructive than necessary.

  Please target the backing CNPG Cluster for the affected DocumentDB CR, or derive it explicitly. For the default playground deployment, something like
  this is safer:

   DOCUMENTDB_NAMESPACE="${DOCUMENTDB_NAMESPACE:-documentdb-preview-ns}"
   DOCUMENTDB_NAME="${DOCUMENTDB_NAME:-documentdb-preview}"
   
   PRIMARY_CLUSTER=$(kubectl --context "$HUB_CONTEXT" -n "$DOCUMENTDB_NAMESPACE" \
     get documentdb "$DOCUMENTDB_NAME" \
     -o jsonpath='{.spec.clusterReplication.primary}')
   
   kubectl --context "$PRIMARY_CLUSTER" -n "$DOCUMENTDB_NAMESPACE" \
     delete cluster.postgresql.cnpg.io "$DOCUMENTDB_NAME" --ignore-not-found

  If the CNPG Cluster name can differ from the DocumentDB CR name, the doc should show how to discover the exact backing CNPG resource instead of using 
  --all.

  🟠 [Required] Define required snippet variables locally

  The snippet uses $HUB_CONTEXT but doesn’t define it in the troubleshooting section. Readers who jump directly to this section can copy-paste into a shell
  where HUB_CONTEXT is empty, which makes the primary discovery command fail.

  Please add defaults near the top of the snippet:

   HUB_CONTEXT="${HUB_CONTEXT:-hub}"
   DOCUMENTDB_NAMESPACE="${DOCUMENTDB_NAMESPACE:-documentdb-preview-ns}"
   DOCUMENTDB_NAME="${DOCUMENTDB_NAME:-documentdb-preview}"

  Then use those variables consistently instead of hard-coded documentdb-preview-ns and documentdb-preview.

  Suggestions

  🟡 [Required] Qualify “cluster” in the new section

  The project documentation standard says to avoid using “cluster” alone. The added section has several unqualified uses, for example:

   - “the cluster sits in Setting up primary”
   - “every member cluster”
   - “failure mode on that cluster”
   - “the primary cluster”

  Please qualify these as “DocumentDB cluster,” “Kubernetes member cluster,” “primary Kubernetes member cluster,” or “CNPG Cluster resource,” depending on
  the meaning.

  🟡 [Suggestion] Use the same Istio command shape as deploy.sh, or explain the difference

  deploy.sh currently applies the control plane with:

   istioctl --context "$cluster" apply -y -f -

  The new troubleshooting snippet uses:

   istioctl --context "$cluster" install -y -f -

  Both may work depending on Istio version, but because the prose says “re-apply the canonical IstioOperator spec,” matching deploy.sh would reduce reader
  uncertainty. If install is intentional for the upgrade path, add a short note explaining why it differs.

  🟡 [Suggestion] Clarify the gateway restart statement

  The comment says:

   # Restart the east-west gateway so its sidecar is re-injected with the new version

  East-west gateways are Istio gateway workloads, not ordinary application pods with injected sidecars. If the goal is to pick up the new
  gateway/control-plane version or refresh gateway pods, reword to avoid implying sidecar injection is the mechanism.

  Questions

   - Is deleting and recreating the CNPG Cluster resource guaranteed to preserve the underlying data PVCs in this playground setup? If yes, it would help 
  to add one sentence saying the DocumentDB CR remains the source of truth and the operator recreates only the backing CNPG resource.
   - Should issue #365 be linked from this README section so readers know the script-level ISTIO_VERSION bump is tracked separately?

WentingWu666666:
Hi @markjbrown  circling back on this.

@xgerman left a detailed review on May 7 (3 Required items, 2 suggestions, 2 questions) and @hossain-rayhan suggested moving the section to the prerequisites instead of troubleshooting. Are you still planning to address these, or would you prefer one of us take over the remaining edits?

Either is fine  just want to keep this moving since the multi-cloud troubleshooting content is useful as-is. I'll give it about a week and then take over if I don't hear back.

PR #367, opened 2026-05-06
FEAT: add kubefleet debugging skill

alaye-ms:


xgerman (commented):
# Skill review: `fleet-debugging`

_Run via the [skill-reviewer agent](https://github.com/xgerman/documentdb-agent-kit/blob/main/.github/agents/skill-reviewer.agent.md) — deterministic check + Anthropic Agent Skills rubric (Level 1 / 2 / 3 progressive disclosure)._

**Architecture cost:** PASS
**Antipatterns violated:** none (#5 informational only — no evals, soft signal for a technical skill)
**Description quality:** **FAIL**

## Token cost (tiktoken/cl100k_base, 200k context)

| Level | What loads | Tokens | % of 200k |
|---|---|---|---|
| L1 — always loaded | frontmatter every turn | 24 | 0.012% |
| L2 — on invocation | SKILL.md body | 815 | 0.419% |
| L3 — on demand | references | 0 (none) | 0% |
| **Full load** | everything | **839** | **0.419%** |

Architecture is well within budget (Anthropic's L2 ceiling is ~5,000 tokens; this skill is at ~16% of that).

## Findings

### 🔴 Blockers
- **`SKILL.md` line 2 — description fails the routing test.** `"Skill for debugging Kubefleet and fleet networking in DocumentDB Kubernetes Operator."` says *what* the skill is, never *when* to invoke it. Rubric §A: *"A description that says only what the skill is, not when to invoke it, fails this check."* This is the highest-leverage fix — the description is loaded every turn and is the only signal the routing model has. Rewrite as `Use when …` with 2–3 concrete trigger situations, e.g. *"a DocumentDB primary fails to join the fleet, a ServiceImport is missing on a member cluster, MemberCluster heartbeat goes stale, MultiClusterService doesn't surface a backing endpoint."*

### 🟡 Improvements
- **JQ pin to `.items[0]` across every command.** The Caveats section asks the agent to "adjust the index value", but when the agent runs commands verbatim in a multi-cluster environment it silently inspects the wrong cluster (Antipattern #3 in spirit — environmental coupling masquerading as a default). Consider hoisting selection into a single discovery step at the top — *"if more than one DocumentDB CR exists, prompt for `${DDB}`"* — and have all subsequent commands use `${DDB}` instead of `[0]`.
- **References section uses prose URLs, not markdown link syntax.** Both the deterministic checker and downstream agent link extractors look for `[label](url)` form. Convert to `[kubefleet](https://github.com/kubefleet-dev/kubefleet)` and `[fleet-networking](https://github.com/Azure/fleet-networking)`.
- **No newline at end of file.** Cosmetic.
- **(info) No evals.** Soft signal — debugging/diagnostic skills don't need a Golden Set; mention only because the checker reports it.

### 🟢 Strengths
- **Right shape.** 75 lines, single-file spine, no unnecessary references — exactly the *target* shape per kit conventions for a small operational skill. Don't split it.
- **Has a `# Caveats` section.** Antipattern #4 (missing Gotchas) does not apply — the regex matches the heading and the content (hub-is-also-member edge case, `[0]` index assumption) is the right kind of *"your environment isn't average"* material.
- **No hardcoded user paths.** Uses `{HUB_CLUSTER}` / `{MEMBER_CLUSTER}` placeholders and `$cluster` loops — portable across reviewers' machines.
- **Discovery-first commands.** The opening "get all documentdb names" + "get cluster names from the CR" pattern is exactly the fix Antipattern #3 prescribes (*"ask the agent to discover the right path rather than declare it"*).

## Recommended next step

Fix the description (Blocker) — rewrite as `Use when …` with 2–3 concrete trigger situations — and ship; the other findings are non-blocking improvements.

WentingWu666666:
**Review: skill is a solid command reference, but not yet a debugging procedure**

Thanks for putting this together -- the commands are accurate (verified against `documentdb_types.go`, `physical_replication.go`, and `deploy-fleet-bicep.sh`) and the Caveats section calls out real footguns. Below is structural feedback aimed at making this usable end-to-end by an AI agent rather than as a command palette.

### What''s missing for an agent to actually debug an incident

**1. Applicability gate.** First check should be: does this DocumentDB even use fleet?
```bash
kubectl get documentdb <name> -n <ns> -o jsonpath=''{.spec.clusterReplication.crossCloudNetworkingStrategy}''
# AzureFleet  -> use this skill
# Istio/None  -> skill does not apply
```

**2. Symptom -> path mapping.** The current skill is a flat list. An agent following it runs every command instead of branching. Even 3-4 named symptoms would help, e.g.:
- *Replica not catching up* -> check operator log -> `.status.clusterReplication` -> CNPG cluster status
- *ServiceImport never appears on member* -> check `MemberCluster` join -> check `ServiceExport` on source -> mcs-controller-manager logs
- *Cross-cluster connection refused* -> DNS resolution -> `<svc>.fleet-system.svc` reachability -> `MultiClusterService` status

**3. The two highest-signal sources aren''t mentioned:**
- **DocumentDB operator log** (`-n documentdb-operator`) -- grep for `Failed to cleanup ServiceImports`, `Failed to force reconcile InternalServiceExports`, `Annotated InternalServiceExports`. This is usually where the root cause shows up.
- **`DocumentDB.status.clusterReplication`** including `TargetPrimary`/`LocalPrimary` (`documentdb_types.go:352-353`) -- authoritative replication state from the operator''s perspective.

**4. Log retrieval commands.** The skill says where logs live but never shows the kubectl invocation. Without explicit `-c <container>`, an agent will hit the refresh-token sidecar by accident on `mcs-controller-manager` / `member-net-controller-manager` (this is in Caveats but easy to miss):
```bash
kubectl --context <member> logs -n fleet-system -l app=mcs-controller-manager -c manager --tail=200
```

**5. `InternalServiceExport` CRD.** Referenced by the controller (`documentdb_controller.go:299`) but missing from the resource list -- it''s the intermediate between `ServiceExport` and `ServiceImport`.

**6. Layer awareness / handoff.** Replication actually flows: DocumentDB CR -> operator -> fleet-networking -> CNPG -> Postgres replication slot. The skill only covers the middle layer. A short "if symptom X, switch to CNPG debugging" pointer would prevent agents from chasing leads in the wrong layer.

**7. Output interpretation.** For each command, one line on what healthy vs unhealthy looks like (e.g., *"a healthy `ServiceImport` has `Conditions[type=Ready]=True`; `Pending` for >2min usually means stale `MemberCluster` heartbeat"*) makes the difference between an agent finding a symptom and an agent diagnosing a cause.

### Smaller items
- `MemberCluster` is cluster-scoped; the `-n fleet-system-hub` flag in the example is silently ignored (cosmetic).
- Consider adding a one-shot "dump everything" command for quick triage.
- Adding a "common root causes" section (e.g., kubeconfig token expiry on member, MCS not enabled on AKS, DNS not resolving `*.fleet-system.svc`) would close the loop from "found the symptom" to "fixed it".

### Suggested minimum additions before merge
1. Applicability gate at the top
2. DocumentDB operator log + `.status.clusterReplication` as primary signals
3. Concrete `kubectl logs` invocation with `-c manager` for the dual-container fleet pods

Happy to help restructure if useful.

PR #374, opened 2026-05-08
feat(api)!: domain-grouped DocumentDBSpec with image/postgres/plugins stanzas + imagePullSecrets

xgerman:
## Summary

Reworked per @wenting-an's review (May-18 comment): the **combined-image mode is removed**. The PR is now a clean domain-grouped restructure of `DocumentDBSpec` plus `spec.imagePullSecrets` passthrough — the two parts of the original PR that were uncontested.

## Why drop combined-image

Wenting's review pointed out — convincingly — that every cited combined-image use case is already covered by the layered/ImageVolume path:

- Custom Postgres image → `spec.image.postgres`
- Mirrored extension image in air-gapped clusters → `spec.image.documentDB`
- Private registries → `spec.imagePullSecrets` (this PR)
- Run-as identity → `spec.postgres.{uid, gid}` (this PR)

Keeping combined alongside layered would:

- Re-fragment the deployment model that 0.2.0 just consolidated when ImageVolume went GA in K8s 1.35.
- Regress the two-phase upgrade contract (operator owns the extension binary lifecycle in layered mode; combined mode hands that to the image builder).
- Forfeit independent CVE patching (Postgres vs extension).
- Double the test/support matrix without delivering a capability that isn't already there.

### CRD restructure (`DocumentDBSpec` 0.2.0 → grouped stanzas)

| New | Replaces |
| --- | --- |
| `spec.image.documentDB` | `spec.documentDBImage` |
| `spec.image.gateway` | `spec.gatewayImage` |
| `spec.image.postgres` | _(new — was hardcoded)_ |
| `spec.plugins.sidecarInjectorName` | _(new — was hardcoded; immutable)_ |
| `spec.plugins.walReplicaName` | _(new — was hardcoded; immutable)_ |
| `spec.postgres.uid` / `spec.postgres.gid` | _(new; XValidation enforces both-or-neither)_ |
| `spec.postgres.postInitSQL` | `spec.postInitApplicationSQL` |

### `spec.imagePullSecrets` passthrough

Top-level `[]corev1.LocalObjectReference` field propagated to `cnpgv1.ClusterSpec.ImagePullSecrets` via `toCNPGImagePullSecrets`. This wires every pod CNPG spawns from the per-CR cluster — engine, gateway, postgres — to honor the secret list.

> Note: the operator/sidecar/wal-replica deployment images use the Helm-chart-level `.Values.imagePullSecrets` introduced in #382. The two knobs are complementary: chart-level covers the operator itself, CRD-level covers the database pods spawned per CR.

## What's _not_ in this PR (anymore)

- ❌ `spec.image.mode` enum (layered/combined) — removed; layered is the only path.
- ❌ `spec.postgres.preloadLibraries` — removed; the operator owns `shared_preload_libraries` (`pg_cron`, `pg_documentdb_core`, `pg_documentdb`) and there's no user-facing reason to expose it now that combined mode is gone.
- ❌ `isCombinedImageMode()` controller helper and the combined branch of `buildPostgresConfiguration()`.
- ❌ "combined image mode" `Context` in `cnpg_cluster_test.go`.
- ❌ `spec.advanced` stanza (was dropped earlier in the PR's history; only mentioned here for completeness).

## Files changed in the rework commit

```
operator/src/api/preview/documentdb_types.go            -- removed Mode const + field, removed PreloadLibraries field
operator/src/internal/cnpg/cnpg_cluster.go              -- removed isCombinedImageMode, collapsed buildPostgresConfiguration
operator/src/internal/cnpg/cnpg_cluster_test.go         -- removed combined-mode Context
operator/src/api/preview/zz_generated.deepcopy.go       -- regenerated
operator/src/config/crd/bases/documentdb.io_dbs.yaml    -- regenerated
operator/documentdb-helm-chart/crds/documentdb.io_dbs.yaml -- regenerated copy
docs/operator-public-documentation/preview/api-reference.md -- regenerated via `make api-docs`
CHANGELOG.md                                            -- restructure bullet updated (drops mode/preloadLibraries mentions)
```

## Validation

- `go build ./...` ✓
- `go vet ./...` ✓
- `go test ./...` (all 5 packages) ✓
- Fresh-checkout simulation (`git archive HEAD | tar -x` then `go vet` + compile-only test) for both `operator/src` and `test/e2e` ✓
- `make manifests generate` + `make api-docs` clean ✓
- `grep -rE 'ImageMode|preloadLibraries|spec\.image\.mode|isCombinedImageMode'` returns no matches ✓

## Migration notes for users

Same as before — anyone using the 0.2.0 spec shape needs to move to the grouped stanzas. The CHANGELOG entry under `[Unreleased] → Breaking Changes` is current.

The only thing additionally worth calling out: there is **no** `spec.image.mode` field — if you carried a draft manifest with `mode: layered` from earlier iterations of this PR, drop the line.


operator/src/api/preview/documentdb_types.go
```diff
@@ -148,7 +185,7 @@ type BootstrapConfiguration struct {
 }
 
 // RecoveryConfiguration defines recovery settings for bootstrapping a DocumentDB cluster.
-// +kubebuilder:validation:XValidation:rule="!(has(self.backup) && self.backup.name != '' && has(self.persistentVolume) && self.persistentVolume.name != '')",message="cannot specify both backup and persistentVolume recovery at the same time"
+// +kubebuilder:validation:XValidation:rule="!(has(self.backup) && self.backup.name != ” && has(self.persistentVolume) && self.persistentVolume.name != ”)",message="cannot specify both backup and persistentVolume recovery at the same time"
```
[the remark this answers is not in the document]
alaye-ms: This change gets added by gofmt, and I agree should be removed. Changing to smart quotes breaks this check

WentingWu666666:
**Suggestion: organize by domain instead of by audience tier**

Thanks for taking on the spec restructure -- the underlying motivation (smaller, scannable root surface; mark fields that aren''t part of the supported default path) is real and worth solving. But I''d like to push on the specific axis chosen.

### The concern with `spec.advanced`

`advanced` is an **audience-tier** group: "fields most users shouldn''t touch." Two issues:

1. **Audience drifts; domain doesn''t.** `affinity` is "rarely touched" today, but as multi-zone HA becomes the default DocumentDB story it becomes a day-one field. We can''t move it back to root without another breaking change. By contrast, "this field configures scheduling" is true today and forever.

2. **K8s ecosystem convention is consistently domain-grouped, not tier-grouped.** PodSpec, CNPG (the operator we wrap), cert-manager, Argo CD, Flux -- all keep `imagePullSecrets`, `affinity`, image refs at root or in domain sub-objects. I can''t find a major CNCF operator using an `advanced` tier. CNPG users reaching for `affinity` will hunt for it at root by reflex.

### Proposed alternative

Group by **what the field configures**, not **who configures it**:

```yaml
spec:
  # Capacity / versioning / security / networking / topology / lifecycle / observability
  # -- all current root fields, unchanged
  ...

  # PodSpec-equivalent primitives: promote to root (K8s convention)
  imagePullSecrets: [...]
  affinity: {...}

  # Image overrides -- domain group of 3
  image:
    documentDB
    gateway
    postgres

  # Postgres process tuning -- domain group of 4 (already nested in PR)
  postgres:
    uid, gid
    preloadLibraries
    postInitSQL

  # Operator plugin selection -- domain group of 2 (borderline; could stay flat)
  plugins:
    sidecarInjectorName
    walReplicaName

  # Operational tuning -- already a group
  timeouts: {...}
```

Result: ~16 root fields (vs ~10 in the PR, ~18 fully flat). Each nested group has a permanent semantic name; nothing graduates out of `image` or `postgres` over time.

### Risk signaling without nesting

The "you''re leaving the supported path" warning the PR achieves via `spec.advanced` can move to per-field godoc + the existing `preview` API version. That''s how K8s itself handles it (KEP-3299 stability tags, alpha/beta/stable in field godoc, separate API versions for whole-resource maturity). It''s per-field precise -- `postgres.postInitSQL` (high risk) and `affinity` (zero risk) don''t have to share one warning level.

### Why now

We''re already paying a breaking-change cost in this PR. If we ship `spec.advanced` and later realize domain grouping is better, that''s a second breaking change. Doing the domain restructure now -- before GA -- costs the same as `advanced` and converges on a structure that holds up long-term.

### Concrete minimum if a full restructure isn''t on the table

If we want to land most of this PR as-is, the highest-leverage two changes are:
1. **Promote `imagePullSecrets` and `affinity` to root** -- they''re standard K8s primitives, not expert knobs.
2. **Rename `spec.advanced` -> `spec.image`** for the image fields, keep `postgres`, `plugins`, `timeouts` as siblings.

Happy to help sketch the type definitions if useful.

xgerman:
### Rewrite per review feedback — domain-grouped spec

Thanks @WentingWu1 (and the rest of the team for the Teams thread). The point that **Postgres config should be operator-agnostic** and that a **k8s operator's spec should be domain-grouped, not audience-tiered** landed. I scrapped the `spec.advanced` umbrella and rebuilt the branch from `documentdb/main` as 5 atomic commits with the structure below.

Force-pushed `feat/combined-image-and-image-pull-secrets` (`12d1db6...f231a96`). New base is `aa1fb07`.

#### New shape

```yaml
spec:
  imagePullSecrets: []                # root — []corev1.LocalObjectReference, promoted
  affinity, timeouts, documentDBVersion, documentDbCredentialSecret  # unchanged
  image:                              # NEW — image overrides (pointer, optional)
    documentDB: ""
    gateway: ""
    postgres: ""
    mode: layered | combined          # default: layered; was inferred before
  postgres:                           # NEW — operator-agnostic process/config knobs
    uid, gid: *int64                  # CEL: has(self.uid) == has(self.gid)
    preloadLibraries: []string
    postInitSQL: []string             # APPENDED after operator's mandatory bootstrap
  plugins:                            # NEW — CNPG plugin selection
    sidecarInjectorName: ""           # CEL: self == oldSelf (immutable)
    walReplicaName: ""
```

Three stanzas (`image`, `postgres`, `plugins`) are pointer types so partial CRs stay valid — all consumers nil-check before assignment.

#### Migration mapping

| Old (`spec.*` or `spec.advanced.*`) | New |
|---|---|
| `spec.documentDBImage` | `spec.image.documentDB` |
| `spec.gatewayImage` | `spec.image.gateway` |
| `spec.postgresImage` | `spec.image.postgres` |
| (empty `documentDBImage` → combined) | `spec.image.mode: combined` (explicit) |
| `spec.postInitSQL` | `spec.postgres.postInitSQL` |
| `spec.preloadLibraries` | `spec.postgres.preloadLibraries` |
| `spec.postgresUID` / `postgresGID` | `spec.postgres.uid` / `gid` |
| `spec.sidecarInjectorPluginName` | `spec.plugins.sidecarInjectorName` |
| `spec.walReplicaPluginName` | `spec.plugins.walReplicaName` |
| `spec.imagePullSecrets` (new) | `spec.imagePullSecrets` (root) |

`DocumentDBStatus.DocumentDBImage` / `GatewayImage` (status-side observability) are **not** renamed.

#### Bug fixes folded in from review

1. **PostInitSQL ordering** — operator's mandatory bootstrap runs first, user `spec.postgres.postInitSQL` is appended after. The previous version let the user list shadow operator bootstrap. Godoc on the field calls this out.
2. **`isCombinedImageMode`** now checks `spec.image.mode == "combined"` explicitly instead of inferring from an empty image string (which was ambiguous when a user set only `gateway`).
3. **UID/GID pairing** — `XValidation` rule `has(self.uid) == has(self.gid)` prevents specifying one without the other.
4. CEL rules use straight quotes only (the prior smart-quote U+201D in one rule would have failed apiserver validation).

#### Commits

- `e3503fa` `feat(api)!: domain-grouped DocumentDBSpec with image, postgres, plugins stanzas`
- `378bc38` `feat(controller): plumb domain-grouped spec through CNPG translator and controllers`
- `2b5dd32` `chore: regenerate CRDs, webhook manifests, and api-reference docs`
- `8993e59` `docs: update playground, design docs, and operator guides for domain-grouped spec`
- `f231a96` `test(e2e): migrate e2e suite to domain-grouped DocumentDBSpec`

#### Validation

- `go test ./...` in `operator/src/` — all 60+ unit tests pass.
- Fresh-checkout simulation (`git archive HEAD | tar -x -C $TMPDIR && cd $TMPDIR/test/e2e && go vet ./... && go test -count=0 -run='^$' ./...`) — all 27 e2e packages compile clean.
- CRDs / webhook manifests / api-reference doc regenerated via `make manifests generate` and committed.

Ready for another pass. @WentingWu1 @hossain-rayhan PTAL.

hossain-rayhan:
I see one ` mode: layered | combined ` item under `spec.image`. Can this be image specific? like layered for gateway and combined for postgres?

Maybe I am wrong but I think it should go under `spec.image.postgres.mode` as it only applies to postgres image.

WentingWu666666:
The domain-grouped restructure (`spec.image` / `spec.postgres` / `spec.plugins`) and `spec.imagePullSecrets` passthrough look great. **My main pushback: I don't think we need combined-image mode at all — the existing two-image flow already covers every use case the PR cites for it.**

Concretely, with the other additions already in this PR you can do all of the following today without combined-image:

| Use case from PR description | Already covered by |
|---|---|
| Custom / hardened / enterprise Postgres image | `spec.image.postgres` |
| Mirrored or re-signed DocumentDB extension | `spec.image.documentDB` |
| Private registry / air-gapped | `spec.imagePullSecrets` (this PR) → applies to both images |
| Custom `shared_preload_libraries` | `spec.postgres.preloadLibraries` (mode-independent) |
| Custom Postgres UID/GID | `spec.postgres.uid` / `spec.postgres.gid` (mode-independent) |
| Custom post-init SQL | `spec.postgres.postInitSQL` (mode-independent) |

So combined-image isn't unlocking new capability — it's offering a second, parallel way to do the same things, and we'd pay real cost for that:

1. **Two-phase upgrade contract regresses.** `documentDBVersion` + `schemaVersion` assumes the operator owns the extension binary lifecycle. In combined-image the extension is baked into the Postgres image, so `documentDBVersion` becomes advisory and the rollback-safe upgrade window we just shipped no longer really applies. How do we want to reconcile that?
2. **Independent CVE patching is lost.** Today users can rebuild Postgres without touching the extension, or bump the extension without rebuilding Postgres. Combined-image collapses these into one rebuild owned by the user.
3. **Support triage gets harder.** "Is it our extension or your image?" stops being answerable when the two are fused.
4. **Test/support matrix doubles.** Every feature (backup/restore, PV recovery, physical replication, two-phase upgrade, TLS, monitoring) has to be validated in both modes. The PR adds three unit-test contexts but no combined-image e2e for backup/replication/upgrade.
5. **We just removed this** for the K8s 1.35+ / ImageVolume baseline (per CHANGELOG). Re-adding it shortly after re-fragments the deployment model we deliberately consolidated.

**Could we step back and answer:**

- Is there a named user/scenario where two custom images + `imagePullSecrets` + `spec.postgres.*` is genuinely insufficient? If it's "ImageVolume is disabled on my platform," is that a platform fix rather than a permanent API surface we maintain?
- If we do keep combined-image, what's the written support contract — specifically how `documentDBVersion`/`schemaVersion` behave when the operator doesn't own the extension binary, and which features are supported vs. best-effort in this mode?

**Suggestion:** split the PR. The restructure and `imagePullSecrets` are clear wins and I'd merge them now. Hold combined-image until we have a concrete user it unblocks and a documented support contract — otherwise we're paying ongoing maintenance and upgrade-story cost for a mode the existing two-image flow already covers.

(Minor: the PR description still refers to a `spec.advanced` wrapper, but the code uses domain-specific blocks — worth updating so reviewers aren't misled.)

WentingWu666666 (approved):

operator/documentdb-helm-chart/crds/documentdb.io_dbs.yaml
```diff
@@ -1198,12 +1192,55 @@ spec:
...
+                      This image is mounted into the PostgreSQL container via CNPG's
+                      ImageVolumeSource so that the extension files are available alongside
+                      an upstream PostgreSQL image.
+                    type: string
+                  gateway:
+                    description: Gateway is the container image for the DocumentDB
+                      Gateway sidecar.
+                    type: string
+                  postgres:
+                    default: ghcr.io/cloudnative-pg/postgresql:18-minimal-trixie
```
hossain-rayhan, line 1213: So, we have default image only for postgres but not for documentDB and gateway?

operator/documentdb-helm-chart/crds/documentdb.io_dbs.yaml
```diff
@@ -1255,13 +1292,57 @@ spec:
...
+              plugins:
                 description: |-
-                  PostgresImage is the container image to use for the PostgreSQL server.
-                  If not specified, defaults to the last stable PostgreSQL version compatible with DocumentDB.
-                  Must use trixie (Debian 13) base to match the extension's GLIBC requirements.
-                type: string
+                  Plugins groups CNPG plugin configuration (sidecar injector name, WAL replica name).
+                  All fields are optional; defaults are preserved when omitted.
+                properties:
+                  sidecarInjectorName:
```
hossain-rayhan, line 1300: We have a default value for this which will never change in ideal case. Do we want to include it like other default values?

operator/src/internal/cnpg/cnpg_cluster.go
```diff
@@ -235,3 +213,104 @@ func parsePullPolicy(value string) corev1.PullPolicy {
...
+	}
+	// TODO: once DocumentDB supports change streams natively, additional GUC parameters may be needed here.
+	if dbpreview.IsFeatureGateEnabled(documentdb, dbpreview.FeatureGateChangeStreams) {
+		params["wal_level"] = "logical"
+	}
+
+	pgHBA := []string{
+		"host all all 0.0.0.0/0 trust",
+		"host all all ::0/0 trust",
+		"host replication all all trust",
```
hossain-rayhan, line 299: Maybe we need to open a separate issue to make this more restricted?

operator/src/internal/cnpg/cnpg_cluster.go
```diff
@@ -235,3 +213,104 @@ func parsePullPolicy(value string) corev1.PullPolicy {
...
+//
+// The operator declares the DocumentDB extension via CNPG's Extensions
+// stanza (mounted from spec.image.documentDB as an ImageVolumeSource),
+// sets a fixed AdditionalLibraries list, and applies a small set of
+// operator-managed GUCs.
+func buildPostgresConfiguration(documentdb *dbpreview.DocumentDB, extensionImageSource corev1.ImageVolumeSource) cnpgv1.PostgresConfiguration {
+	params := map[string]string{
+		"cron.database_name":    "postgres",
+		"max_replication_slots": "10",
+		"max_wal_senders":       "10",
```
hossain-rayhan, line 289: Should we expose these with default value 10?

hossain-rayhan (approved):

PR #379, opened 2026-05-18
feat: allow using an existing Cloudnative PG operator

AntonOfTheWoods:
- flag for cnpg chart install
- use a template for cnpg namespace rather than hard-code
- various other potential failure ignores

No change to default behaviour

### DIFF
--- a/operator/documentdb-helm-chart/Chart.yaml
+++ b/operator/documentdb-helm-chart/Chart.yaml
@@ -9,3 +9,5 @@ dependencies:
   - name: cloudnative-pg
     version: "0.28.1"
     repository: "https://cloudnative-pg.github.io/charts/"
+    alias: cnpg
+    condition: cnpg.install
--- a/operator/documentdb-helm-chart/values.yaml
+++ b/operator/documentdb-helm-chart/values.yaml
@@ -23,9 +23,9 @@ serviceAccount:
   automount: true
   annotations: {}
   name: "documentdb-operator"
-  
+
 # WAL Replica feature flag
-walReplica: false  # Set to true to deploy the WAL replica plugin
+walReplica: false # Set to true to deploy the WAL replica plugin
 
 image:
   documentdbk8soperator:
@@ -37,8 +37,10 @@ image:
   walreplica:
     repository: ghcr.io/documentdb/documentdb-kubernetes-operator/wal-replica
     pullPolicy: Always
-cloudnative-pg:
+cnpg:
+  install: true
   namespaceOverride: cnpg-system
+  webhookFailurePolicy: Fail
   additionalEnv:
     - name: ENABLE_INSTANCE_MANAGER_INPLACE_UPDATES
       value: "true"
--- a/operator/documentdb-helm-chart/templates/01_namespace.yaml
+++ b/operator/documentdb-helm-chart/templates/01_namespace.yaml
@@ -1,7 +1,9 @@
+{{- if .Values.cnpg.install }}
 apiVersion: v1
 kind: Namespace
 metadata:
-  name: cnpg-system
+  name: {{ include "documentdb-chart.cnpgNamespace" . }}
   labels:
     app.kubernetes.io/name: {{ include "documentdb-chart.name" . }}
     app.kubernetes.io/managed-by: "Helm"
+{{- end }}
--- a/operator/documentdb-helm-chart/templates/02_documentdb_sidecar_injector.yaml
+++ b/operator/documentdb-helm-chart/templates/02_documentdb_sidecar_injector.yaml
@@ -10,7 +10,7 @@ metadata:
     cnpg.io/pluginName: cnpg-i-sidecar-injector.documentdb.io
     app.kubernetes.io/name: {{ include "documentdb-chart.name" . }}
   name: sidecar-injector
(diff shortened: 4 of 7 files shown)


operator/documentdb-helm-chart/Chart.yaml
```diff
@@ -8,4 +8,6 @@ appVersion: "0.2.0"
 dependencies:
   - name: cloudnative-pg
     version: "0.28.1"
-    repository: "https://cloudnative-pg.github.io/charts/"
\ No newline at end of file
+    repository: "https://cloudnative-pg.github.io/charts/"
+    alias: cnpg
```
WentingWu666666, line 12: Concern with `alias: cnpg`: this is a silent breaking change — any user who has overridden `cloudnative-pg.*` values today will find their overrides become no-ops after upgrading, since the subchart now reads from `cnpg.*` instead.
 
The benefit also feels small: it just shortens the key name. We can get the same BYO toggle without the rename by pointing condition: at a value under the existing `cloudnative-pg:` block. That keeps the chart backward compatible.

operator/documentdb-helm-chart/values.yaml
```diff
@@ -37,8 +37,10 @@ image:
   walreplica:
     repository: ghcr.io/documentdb/documentdb-kubernetes-operator/wal-replica
     pullPolicy: Always
-cloudnative-pg:
+cnpg:
+  install: true
   namespaceOverride: cnpg-system
+  webhookFailurePolicy: Fail
```
WentingWu666666, line 43: `webhookFailurePolicy` controls the DocumentDB operator's webhook, not CNPG's — worth moving it out from under the CNPG block.
 
This value is consumed in `10_documentdb_webhook.yaml`, which is the `ValidatingWebhookConfiguration` for `dbs.documentdb.io` — the DocumentDB operator's own webhook. It has no relationship to CNPG.
 
 Two reasons the current placement is confusing:
 1. The upstream CNPG chart already exposes its own webhook failure policies under `cloudnative-pg.webhook.mutating.failurePolicy` and `cloudnative-pg.webhook.validating.failurePolicy`. Users seeing `cnpg.webhookFailurePolicy` next to those will reasonably assume it's a third knob for CNPG when it isn't.
 2. Grouping by use case ("things I added for the BYO PR") rather than by what the value configures makes the values schema harder to reason about over time.
 
Suggestion: move it to a top-level `webhook:` block, mirroring the upstream CNPG schema so the convention is consistent:
 ```
  webhook:
    validating:
      failurePolicy: Fail
 
  # templates/10_documentdb_webhook.yaml
  failurePolicy: {{ .Values.webhook.validating.failurePolicy | default "Fail" }}
```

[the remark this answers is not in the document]
Comment (hossain-rayhan) [reply operator/documentdb-helm-chart/values.yaml:43]: good suggestion. Let's keep the old full form.

File: operator/documentdb-helm-chart/values.yaml [hunk 3]
```diff
@@ -37,8 +37,10 @@ image:
   walreplica:
     repository: ghcr.io/documentdb/documentdb-kubernetes-operator/wal-replica
     pullPolicy: Always
-cloudnative-pg:
+cnpg:
+  install: true
```
WentingWu666666, line 41: Naming nit: consider bundled over install.
 
`install: true|false` reads as an action, which is a bit awkward in the `false` direction — does it mean "don't install now", "uninstall", or "it's already installed"? The state being expressed is really "is CNPG bundled with this chart, or provided externally?", and naming the state directly reads more cleanly:
 ```
  cloudnative-pg:
    bundled: true   # default: chart installs and manages CNPG
                    # false: assume CNPG is already installed (BYO)
 ```

Symmetric, names the ownership relationship directly, and unlikely to ever collide with a key the upstream CNPG chart might add. Non-blocking — `install` or `enabled` are also defensible — but bundled is worth considering since this is the first user-facing API for the BYO toggle and changing it later requires a deprecation cycle.

hossain-rayhan:
Thanks for the contribution. 

A good e2e test, playground/ proper documentation update with guideline would be a great addition to this PR.

WentingWu666666:
**Non-blocking from my side, just flagging for a follow-up issue.**

Separate from the comments I've already left on this PR: once users can opt into BYO mode, we implicitly take on a version-compatibility surface we didn't have before. The operator will run against whatever CNPG version the user already has installed, which could be older or newer than what we test against. Today there's no declared supported range, no detection, and no enforcement  so out-of-range combinations will fail in subtle ways (silent backup skips, replication issues) that look like operator bugs.

I don't want to expand the scope of this PR  happy to address that as a follow-up. Filed #380 to track the compatibility work (declared range, runtime detection, CI matrix) so it doesn't get lost.

AntonOfTheWoods:
Thanks for your careful comments. Unfortunately, given the current state of the chart and the level of quality you are requiring, it's not something I am going to be able to commit to. Quite frankly, the chart does not, shall we say, represent what is broadly considered best practice. Like at all. It's really more like a bunch of resource files from an alpha test put into a templates directory than an actual helm chart. 

I think maybe someone with commit rights should ask claude to turn it into a proper chart (with e2e tests maybe). It could also probably do what this PR helps to do.

Issue #381, opened 2026-05-19
Helm chart improvements for GA readiness


WentingWu666666:
## Summary

Pre-GA audit of `operator/documentdb-helm-chart`. The chart works for the current demo/preview use cases, but several gaps will block enterprise adoption (multi-install collisions, Pod Security Admission rejection, broad RBAC, no CRD upgrade story, missing scheduling/security knobs).

This is a tracking issue. Individual fixes can be split into sub-PRs; some items may have better designs than what is suggested here.

Related: #380 (CNPG version compatibility for BYO mode).

---

##  Critical (GA blockers)

- [x] **C1. Cluster-scoped resources have hardcoded names**  chart cannot be installed twice in the same cluster. `ClusterRole/documentdb-operator-cluster-role`, `ClusterRoleBinding/...`, `ValidatingWebhookConfiguration/documentdb-validating-webhook`, `Service/documentdb-webhook-service`, all three Deployments, and the wal-replica RBAC are all hardcoded. Needs a proper `documentdb-operator.fullname` helper and consistent use across every cluster-scoped resource. `templates/10_documentdb_webhook.yaml` even has a comment admitting this.
- [x] **C2. No CRD upgrade story.** CRDs live in `crds/` (Helm 3 special dir)  installed on first install, never updated by `helm upgrade`. Pick a strategy: hook-based install in `templates/`, separate `documentdb-operator-crds` chart (Strimzi/CNPG pattern), or documented manual `kubectl apply` step in release notes.
- [x] **C3. `Deployment.spec.selector.matchLabels: { app: {{ .Release.Name }} }`**  selectors are immutable; coupling to release name risks `helm upgrade` failures. Standardize on `app.kubernetes.io/*` labels with `selectorLabels` / `labels` helpers (what `helm create` ships).
- [x] **C4. No `resources:` requests/limits** on operator, sidecar-injector, or wal-replica deployments. LimitRange/ResourceQuota enforcement will reject the install. Expose via `values.yaml` with sane defaults.
- [x] **C5. ClusterRole is over-broad.** `templates/05_clusterrole.yaml` grants cluster-wide CRUD on `secrets`, `namespaces`, `rolebindings`, `clusterrolebindings`, `serviceaccounts`, `configmaps`. Several are not justifiable for a DocumentDB operator and will fail security review. Audit each rule against actual controller code.
[...]

PR #386, opened 2026-05-19
chore(helm): trim over-broad ClusterRole permissions

WentingWu666666:
## Summary

Trim the operator's and wal-replica's ClusterRoles to the set actually consumed by the controllers, narrowing the blast radius of a ServiceAccount-token compromise.

## Correction (commit 5bcd2e5)

E2E run [#26109274996](https://github.com/documentdb/documentdb-kubernetes-operator/actions/runs/26109274996/job/76784067844?pr=386) surfaced an RBAC escalation failure that my initial trim caused: the operator's `EnsureServiceAccountRoleAndRoleBinding` (`documentdb_controller.go:452-458`) creates a per-DocumentDB namespaced Role granting `pods, services, endpoints` CRUD to the CNPG cluster's ServiceAccount. K8s RBAC escalation rules require the grantor to already hold every verb being granted, so the operator's ClusterRole must keep these as full CRUD.

Restored in `5bcd2e5`:
- operator ClusterRole: `""/pods` full CRUD, `""/endpoints` full CRUD
- wal-replica ClusterRole: `""/endpoints` full CRUD (defensive)

**Methodology gap:** my grep audit only catches `&<group>v1.<Kind>{}` instantiated for client calls; it missed resources the operator only *grants* via `rbacv1.PolicyRule` string literals. Future RBAC audits must also grep `PolicyRule{` and `Resources: []string{...}` construction sites. Tracked in #390.

All other trims from `a6e7015` remain  none of them appear in any PolicyRule grant.

## How the trim was decided

Two-source audit:
1. **`+kubebuilder:rbac` markers** in `operator/src/internal/`. Found markers in only 4 files; backup, scheduledbackup, physical_replication, util, and cnpg_sync have **none**. So `make manifests` produces a `role.yaml` that's missing ~70% of the operator's real permission needs. The Helm chart has therefore been hand-maintained and drifted. **Fixing the markers so the chart round-trips through `make manifests` is the right long-term fix and is tracked separately (see #390).**
2. **Actual API call sites + PolicyRule construction sites**: grep across `operator/src/internal/` for every `&<group>v1.<Kind>{}` passed to `client.Get/List/Create/Update/Patch/Delete/Watch`, **plus** every `rbacv1.PolicyRule{...}` and `Resources: []string{...}` literal where the operator grants permissions to other SAs. This is the empirical floor of "what the controllers actually touch or grant."

Anything in the chart that is granted but **never referenced by code (as a client call OR a grant)** is removed.

## What changes  operator ClusterRole (`templates/05_clusterrole.yaml`)

| Removed | Why it's dead |
|---|---|
| `""/namespaces` (full CRUD) | No `corev1.Namespace` references in any controller; user creates target namespaces. |
| `rbac/clusterrolebindings` (full CRUD) | `util.go` `CreateRole` / `CreateRoleBinding` create *namespaced* RBAC only. |
| `apps/statefulsets`, `apps/deployments` (full CRUD) | Never instantiated; CNPG owns StatefulSet management for database pods. |
| `batch/jobs` (full CRUD) | No `&batchv1.Job{}` anywhere in `operator/src`. |
| `metrics.k8s.io/pods` (entirely) | No controller calls the metrics API. |
| `""/leases` | Wrong API group. Leader election uses `coordination.k8s.io/leases` (granted separately); core `leases` was a historical pre-1.14 location. |
| `cert-manager.io/clusterissuers` | CR exposes `IssuerRef.Kind: ClusterIssuer` but the operator only *references* it by name in the Certificate spec it creates  cert-manager itself resolves it. Operator never Get/Lists ClusterIssuer. |
| `cert-manager.io/certificates/finalizers` | Certificate controller never sets finalizers on cert-manager objects (only on its own DocumentDB CR). |

| Narrowed | From | To | Why |
|---|---|---|---|
| `""/secrets` | full CRUD | `get,list,watch` | Only `certificate_controller.go:115` reads TLS Secrets; no controller writes them; no PolicyRule grants `secrets`. |

`""/pods` and `""/endpoints` remain full CRUD  see Correction above.

## What changes  wal-replica ClusterRole (`templates/03_documentdb_wal_replica.yaml`)

Same dead grants removed where unambiguous (`""/leases`, `metrics.k8s.io/pods`). `""/endpoints` restored (see Correction). The wal-replica plugin's Go source lives in the [`cnpg-i-wal-replica`](https://github.com/cnpg-i/cnpg-i-wal-replica) repo, so the remaining rules are kept as-is until the plugin ships its own RBAC markers.

## What is intentionally NOT changed (yet)

These look trimmable but need per-controller audit-log validation to be safe:
- `""/configmaps`  physical_replication writes them; verbs probably correct.
- `""/services`, `""/serviceaccounts`, `rbac/roles`, `rbac/rolebindings`  `util.go` creates them, but `delete` verbs need exec-path audit to confirm.
- `""/persistentvolumeclaims`  current code uses `get,list,watch,create,delete`, no `update/patch`; conservative to leave as-is.
- `coordination.k8s.io/leases`  required for leader election, but should ideally be a `Role` in the operator namespace, not a `ClusterRole`. (Migration step, not just a verb change.)

These belong in a separate PR alongside the marker fix.

## Tests

- `helm lint`
- `helm upgrade` against a live kind cluster with an existing DocumentDB CR (`documentdb-cluster` in `documentdb-preview-ns`)
- Restarted operator post-fix: per-CR Role `documentdb-cluster` re-created successfully with `pods,services,endpoints` CRUD; no `forbidden` log entries.
- Operator pod stays `1/1 Running`, no restart, no `forbidden` / `cannot list` / `cannot get` / `cannot create` / `denied` / `RBAC` log entries over 2 minutes of normal reconciliation.
- pv-controller, documentdb-controller, certificate-controller, backup-controller, scheduledbackup-controller all confirmed reconciling post-upgrade.

## Follow-ups

- #390: fix `+kubebuilder:rbac` markers across all controllers so `make manifests` produces a `role.yaml` that matches the chart, **and** extend audit methodology to grep `PolicyRule{` construction sites, then add CI assertion.
- Future PR: convert leader-election `leases` from ClusterRole to namespace-scoped Role.

## Refs

#381


WentingWu666666:
Filed #390 for the marker-drift follow-up referenced in the description.

hossain-rayhan (approved):

alaye-ms (approved):

PR #389, opened 2026-05-19
fix(helm): standardize labels and selectors with app.kubernetes.io/* (#381 C3)

WentingWu666666:
## Summary

The chart''s three Deployments (operator, sidecar-injector, wal-replica) use ad-hoc selectors and don''t emit the standard `app.kubernetes.io/*` labels on pods. This PR fixes both with a new `_helpers.tpl` that follows the CNPG convention.

## Why this matters

The operator Deployment''s previous selector was:

```yaml
selector:
  matchLabels:
    app: {{ .Release.Name }}
```

Two bugs flowed from this:

1. **Renaming a release breaks `helm upgrade`**  `Deployment.spec.selector.matchLabels` is immutable; renaming a release changes the selector value; upgrade fails with `field is immutable`.
2. **Label-based pod lookup doesn''t work.** The chart never emitted `app.kubernetes.io/name`, `instance`, or `component` on pods, so `kubectl get pod -l app.kubernetes.io/...`, Lens, k9s, Prometheus relabel rules, and ArgoCD/Flux ownership matchers all fail to identify the operator''s pods.

The two cnpg-system plugins had the same gap (`app: sidecar-injector` / `app: wal-replica`).

## What changes

- New `templates/_helpers.tpl` defining `name`, `chart`, `namespace`, `labels`, `selectorLabels`, `serviceAccountName`. `selectorLabels` takes a `component` argument so each Deployment gets a distinct, immutable selector.
- All three Deployments and their fronting Services switched to the new helpers.
- All resources now carry the full `app.kubernetes.io/{name,instance,component,version,managed-by}` + `helm.sh/chart` set.
- Cleanups along the way (no behavior change):
  - Removed `creationTimestamp: null` / `strategy: {}` artifacts left from kustomize.
  - Replaced `app.kubernetes.io/managed-by: kustomize` / `"Helm"` literals with `.Release.Service`.

## What does NOT change

**Resource names are intentionally unchanged.** An earlier draft of this PR also templatized the ClusterRole / ClusterRoleBinding / ServiceAccount names with the CNPG-style `<release>-` prefix to enable multiple installs in one cluster. On review we dropped that: the operator is a **cluster-singleton** by design  its CRDs and `ValidatingWebhookConfiguration` are cluster-scoped, and CNPG itself accepts the same single-install limitation (they hardcode `cnpg-validating-webhook-configuration` / `cnpg-webhook-service`). Templating the cluster-scoped RBAC names would have:

- forced every existing install through an uninstall/reinstall migration
- not actually enabled multi-install (CRDs and webhook still collide)

So we keep stable names. The RBAC layer is upgrade-clean across this change.

##  Breaking change for existing installs (smaller scope than the previous draft)

Selectors are immutable, so `helm upgrade` from current `main` will fail with `field is immutable` on each of the three Deployments. The fix is to **delete only the Deployments**, then `helm upgrade`. RBAC, ServiceAccount, webhook Service, and ValidatingWebhookConfiguration are unaffected.

```bash
kubectl -n documentdb-operator delete deploy documentdb-operator
kubectl -n cnpg-system delete deploy sidecar-injector wal-replica   # if walReplica=true
helm upgrade documentdb-operator <chart> -n documentdb-operator
```

This is acceptable pre-GA (chart is in preview). User-managed `DocumentDB` CRs and the underlying CNPG `Cluster`s are unaffected (different controller scope).

## Local verification

- `helm lint` clean.
- `helm template` (default and `--set walReplica=true`) renders the **same resource names as on `main`** (zero rename); every Deployment selector resolves to:
  ```yaml
  app.kubernetes.io/name: documentdb-operator
  app.kubernetes.io/instance: <release>
  app.kubernetes.io/component: <operator|sidecar-injector|wal-replica>
  ```
- Fresh `helm install` on kind: operator, sidecar-injector, and CNPG operator pods reach Ready in <30s.
- Validating webhook still reachable: a `DocumentDB` CR with `spec.nodeCount: -1` is rejected with the expected schema validation error.
- `kubectl get pod -l app.kubernetes.io/component=operator` and `=sidecar-injector` resolve to the expected pods (confirms labels reach pod templates).

## Related

- Tracks **#381 C3** (selector standardization)
- **Not included** (and intentionally so): #381 C1 (templatizing cluster-scoped RBAC names). See discussion above for why.

### DIFF
--- a/operator/documentdb-helm-chart/templates/01_namespace.yaml
+++ b/operator/documentdb-helm-chart/templates/01_namespace.yaml
@@ -3,5 +3,5 @@ kind: Namespace
 metadata:
   name: cnpg-system
   labels:
-    app.kubernetes.io/name: {{ include "documentdb-chart.name" . }}
-    app.kubernetes.io/managed-by: "Helm"
+    app.kubernetes.io/name: {{ include "documentdb-operator.name" . }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
--- a/operator/documentdb-helm-chart/templates/02_documentdb_sidecar_injector.yaml
+++ b/operator/documentdb-helm-chart/templates/02_documentdb_sidecar_injector.yaml
@@ -6,9 +6,10 @@ metadata:
     cnpg.io/pluginPort: "9090"
     cnpg.io/pluginServerSecret: sidecarinjector-server-tls
   labels:
-    app: sidecar-injector
     cnpg.io/pluginName: cnpg-i-sidecar-injector.documentdb.io
-    app.kubernetes.io/name: {{ include "documentdb-chart.name" . }}
+    {{- include "documentdb-operator.labels" (dict "root" $ "component" "sidecar-injector") | nindent 4 }}
+  # Name hardcoded: CNPG-I plugin Services in cnpg-system are shared infrastructure
+  # discovered by name; templating the name would require coordinated config in CNPG.
   name: sidecar-injector
   namespace: cnpg-system
 spec:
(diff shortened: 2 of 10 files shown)


WentingWu666666:
Reworked: dropped the cluster-scoped name templating (C1) on review. Operator is a cluster-singleton by design (CRDs + VWC), so the \<release>-\ prefix was solving a problem we don't have, while forcing every existing install through a rename. Kept just the selector + label standardization (C3). Resource names now match \main\ exactly; migration shrinks to deleting only the three Deployments.

Issue #375, opened 2026-05-09
Multi-cloud failover: WAL timeline divergence after promotion leaves replica unrecoverable


markjbrown:
## Summary

Rapid back-to-back failovers between members of a multi-cloud `DocumentDB` resource leave the demoted primary in a state where its WAL receiver cannot reconnect to the new primary, because the two clusters end up on incompatible PostgreSQL **timelines**. The demoted primary is unrecoverable via log replay and requires a full `pg_basebackup` rebuild to re-attach as a replica.

The DocumentDB operator does not appear to:

1. `pg_rewind` the demoted primary before re-attaching it as a replica, **or**
2. Refuse / serialize a second promotion until cross-cloud replication is confirmed streaming again, **or**
3. Auto-bootstrap a replica via `pg_basebackup` when its WAL receiver fails with a timeline-mismatch FATAL.

The result is a manual recovery path (delete the CNPG `Cluster` resource so the operator re-creates it) that is easy to get wrong, especially in a demo / live-failover scenario.

## Repro

Multi-cloud setup from <https://github.com/markjbrown/documentdb-local-to-multicloud>:

- Hub cluster + 2 member clusters (Azure AKS + AWS EKS), `instancesPerNode: 3` on both
- `DocumentDB` resource with `clusterReplication.primary: azure-documentdb` and `crossCloudNetworkingStrategy: serviceLoadBalancer`
- DocumentDB operator preview build

Steps:

1. Promote AWS: `kubectl --context hub -n documentdb-preview-ns patch documentdb documentdb-preview --type=merge -p '{"spec":{"clusterReplication":{"primary":"aws-documentdb"}}}'`
2. Wait until both clusters report Healthy in their CNPG status (~30-60s).
3. Promote back to Azure with the same patch (target=`azure-documentdb`).
4. Within 30s of the second promotion completing, observe the replica.

## Observed

On the new replica (AWS in this repro):

```
$ kubectl --context aws-documentdb -n documentdb-preview-ns logs <pod>-1 -c postgres | tail
FATAL: requested starting point 3/88000000 on timeline 8 is not in this server's history
DETAIL: This server's history forked from timeline 8 at 3/8704C8C0
```

```
$ kubectl --context aws-documentdb -n documentdb-preview-ns exec <pod>-1 -c postgres -- \
    psql -U postgres -tA -c "select status, sender_host from pg_stat_wal_receiver"
(0 rows)
```

[...]

PR #392, opened 2026-05-28
fix: serve fresh promotion token across back-to-back failovers

WentingWu666666:
## Summary

Fixes two bugs in `ensureTokenServiceResources` that caused stale demotion tokens to be served to the new primary during rapid back-to-back failovers, contributing to silent WAL timeline divergence (issue #375).

### 1. ConfigMap update silently rejected with `Conflict`
The previous code built a brand-new in-memory `ConfigMap` (no `resourceVersion`) and called `Update` directly. After the first failover the API server rejected the second `Update` with HTTP 409 Conflict; the error was logged but ignored, so the **old token kept being served** to subsequent failovers.

**Fix:** use `controllerutil.CreateOrUpdate`, which does `Get` → mutate → `Update` and preserves the current `resourceVersion`.

### 2. nginx Pod served stale `index.html` due to slow ConfigMap volume sync
Even when the ConfigMap eventually carried the new token, the nginx Pod mounts it as a volume — and kubelet's ConfigMap volume sync is eventually-consistent (typically 30–60s). During back-to-back failovers the cross-cluster HTTP `GET` from the new primary therefore read the **previous** failover's token. CNPG then fast-forwarded WAL to the wrong LSN, triggering timeline divergence.

**Fix:** compute `sha256(token)` and annotate both the ConfigMap and the Pod with `documentdb.io/promotion-token-revision`. When they diverge the Pod is deleted; the recreated Pod mounts the freshly-updated ConfigMap immediately. Same pattern as `kubectl rollout restart`'s pod-template-hash annotation.

### Bonus
Added `OwnerReferences` (CNPG cluster) to the ConfigMap, Pod and Service so they are GC'd with the parent cluster (previously only the `ServiceExport` had owner refs).

## Tests

3 new unit tests in `physical_replication_test.go`:
- **`updates the ConfigMap with the latest token across consecutive failovers`** — regression test for bug 1; would have caught the silent `Conflict` on the second pass.
- **`deletes the stale nginx Pod when the token revision changes`** — regression test for bug 2.
- **`returns false without error when the demotion token is not yet available`** — guards the early-exit path.

```
$ go test ./internal/controller/...
ok      github.com/documentdb/documentdb-operator/internal/controller   7.819s
```

Full 156-spec controller suite is green; `go vet` clean.

## Validation status

- [x] Unit tests
- [ ] End-to-end validation on a real KubeFleet OSS deployment (in progress, follow-up commit will note completion)

## Risk

Low. The change is local to `ensureTokenServiceResources` and preserves all existing call sites. The owner-references addition means deleting a `DocumentDB` / CNPG `Cluster` will now also remove the promotion-token ConfigMap/Pod/Service — desired behaviour.

Related: #375

### DIFF
--- a/operator/src/internal/controller/physical_replication.go
+++ b/operator/src/internal/controller/physical_replication.go
@@ -5,6 +5,8 @@ package controller
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
 	"io"
 	"net/http"
@@ -24,7 +26,19 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+const (
+	// promotionTokenRevisionAnnotation stores a content hash of the promotion
+	// token currently expected by the operator. It is applied to the ConfigMap
+	// holding the token AND the nginx Pod that serves it; when the two diverge
+	// the Pod is force-recreated so it picks up the freshly-mounted ConfigMap
+	// content without waiting for kubelet's eventually-consistent ConfigMap
+	// volume sync (which can be 30-60s — long enough to make a back-to-back
+	// failover read a STALE token and trigger WAL timeline divergence).
+	promotionTokenRevisionAnnotation = "documentdb.io/promotion-token-revision"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


WentingWu666666:
## End-to-end validation update

Ran the fix against a real Azure environment (2 AKS clusters in westus3 + uksouth, KubeFleet OSS, AzureFleet networking strategy, custom operator image `acrwwdocdbfix1133.azurecr.io/documentdb-k8s-operator:fix-stale-token-5daf7af`).

### Positive signals
- Failover #1 (westus3 → uksouth) succeeded; post-failover writes on the new primary replicated back to the demoted side within ~0s.
- Promotion-token resources were recreated with **distinct revision hashes** on each side across failovers (the new `documentdb.io/promotion-token-revision` annotation is working).
- **No WAL timeline divergence signatures** were observed in Postgres logs across either rapid failover — none of the `requested WAL position ... has not been promoted` errors from #375.

### Inconclusive
- Failover #2 (uksouth → westus3, ~34s after #1 completed) **stalled before the fix path could be exercised** — the operator on both clusters got stuck on `Get http://documentdb-preview-ns-promotion-token.fleet-system.svc: no such host` and the promotion-token MCS state was `Unknown` on westus3.
- The diagnostic log `Promotion-token Pod has stale revision` was never emitted, so the new stale-Pod-deletion remediation path was not directly observed firing.

### Root cause of the stall is upstream of this fix
The second-failover stall is the fleet-networking MultiClusterService propagation race we've seen before (separate from this PR). The `promotion-token` MCS condition stayed `Unknown` on the side that needed to consume it, so the operator's HTTP GET to nginx never even got a TCP socket — it failed DNS. That happens before the operator can read the (possibly stale) token, so this run can neither confirm nor refute that the fix prevents stale-token reads on the second rapid failover.

### Recommendation
Keep the PR in draft. The fix is sound by code review and unit tests; the two known stale-token failure modes are demonstrably gone in code. To get a clean end-to-end signal we need a fleet environment where MCS health is verified `True` on both clusters before each failover. I'll file a separate issue for the fleet-networking MCS instability as it's a recurring blocker for failover validation.

Full report and artifacts in the session workspace (`v4-validation-report.md`).

PR #393, opened 2026-05-28
test(e2e): add expired backup cleanup coverage

### LINKED ISSUE #370: test(e2e): add expired backup cleanup coverage
## Summary

PR #346 migrates the legacy backup workflow into the unified Go/Ginkgo E2E suite, but it does not yet carry over the legacy expired-backup cleanup scenario from `.github/workflows/test-backup-and-restore.yml`.

This issue tracks restoring that coverage before retiring the legacy workflow.

Origin: https://github.com/documentdb/documentdb-kubernetes-operator/pull/346#discussion_r4393216469

## Legacy behavior to preserve

The legacy workflow:

1. Creates or reuses a completed `Backup` CR.
2. Patches `backup.status.expiredAt` to a timestamp in the past, e.g. `2000-01-01T00:00:00Z`, using the Kubernetes **status subresource**.
3. Polls until the operator deletes the expired `Backup` CR.

The important implementation detail is that this must patch status, not spec:

```bash
kubectl patch backup <name> \
  --type='merge' \
  --patch='{"status":{"expiredAt":"2000-01-01T00:00:00Z"}}' \
  --subresource=status
```

## Proposed E2E shape

Add `test/e2e/tests/backup/expiration_cleanup_test.go`:

- Create a DocumentDB cluster using the existing backup area helpers.
- Create an on-demand `Backup`.
- Wait for `Backup` phase `completed`.
- Patch `status.expiredAt` to a past timestamp via `c.Status().Patch(...)`.
- `Eventually` assert `apierrors.IsNotFound` for that Backup CR.

Add helpers in `test/e2e/pkg/e2eutils/backup/backup.go`:

- `MarkExpired(ctx, c, backup, when)` using `c.Status().Patch(...)`.
- `WaitForBackupDeleted(ctx, c, ns, name, timeout)` mirroring the existing `WaitForPVCDeleted` helper style.

## Definition of Done

- [ ] New Ginkgo spec exercises expired backup garbage collection.
- [ ] Status is patched through `client.Status().Patch`, not `client.Patch`.
- [ ] Helper unit tests cover `WaitForBackupDeleted` success and timeout paths.
- [ ] The spec is labeled consistently with backup tests, including `backup` and `needs-csi-snapshots`.
- [ ] Legacy workflow retirement is not considered complete until this coverage exists in the unified E2E suite.

xgerman:
Closes #370.

## What

Adds an e2e spec verifying the operator garbage-collects `Backup` CRs once `status.expiredAt` is in the past, plus two reusable helpers in `test/e2e/pkg/e2eutils/backup`:

- **`MarkExpired`** — patches `status.expiredAt` via the **status subresource** to fast-forward expiration without waiting out the real `retentionDays` window. Routing through the status subresource matters because the Backup type declares `+kubebuilder:subresource:status`, so a plain `c.Patch` on the same field is a silent no-op.
- **`WaitForBackupDeleted`** — polls until the Backup CR is gone, treating `NotFound` as success.

## Why

Issue #370 — the operator's retention path (`backup_controller.go` `IsExpired()` → `r.Delete`) had no e2e coverage. The new helpers also unblock similar cleanup-style specs later (scheduled-backup retention, restore-source pruning, etc.).

## How the spec works

1. Provision a 1-node DocumentDB.
2. Create an on-demand `Backup` and wait for `Completed` — marking an in-flight backup expired would race the snapshot controller and the `retentionDays` contract is defined against completed backups.
3. `MarkExpired(now - 1h)` via status patch.
4. Assert the Backup CR is gone within 3 minutes (status patch fires a watch event, so deletion is event-driven, not bound to the 1-minute periodic requeue).

Labels: `[backup, needs-csi-snapshots, level:medium]`.

## Validation

- `go vet ./...` clean across `test/e2e`
- `go test ./pkg/e2eutils/backup/...` 11/11 pass (5.6s) — 4 new unit tests cover both status-subresource patch path and Get/NotFound semantics
- `ginkgo --focus 'expired backup cleanup' ./tests/backup` ✅ **PASS in 67s** on local kind (CSI hostpath + external-snapshotter)
- Fresh-checkout simulation (`git archive HEAD | tar -x; go vet + test`) clean

## Out of scope

- No `CHANGELOG.md` entry — test-only change, borderline per `CONTRIBUTING.md` examples.
- No `retry.RetryOnConflict` wrapper on `MarkExpired` — single in-flight test writer, conflict surface is near-zero. Easy to add later if other specs adopt the helper concurrently.

### DIFF
--- a/test/e2e/pkg/e2eutils/backup/backup.go
+++ b/test/e2e/pkg/e2eutils/backup/backup.go
@@ -23,6 +23,7 @@ import (
 	cnpgv1 "github.com/cloudnative-pg/cloudnative-pg/api/v1"
 	"github.com/cloudnative-pg/cloudnative-pg/tests/utils/envsubst"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/yaml"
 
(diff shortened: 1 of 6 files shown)


WentingWu666666 (commented):
Two small suggestions:

**1. Downstream cleanup isn't asserted**. The spec confirms the Backup CR disappears, but Reconcile also drives cleanup of the underlying VolumeSnapshot (and any CNPG-side artifacts) when a backup expires. Today the test would still pass if the controller deleted the CR but leaked the snapshot — which is the more user-visible failure mode (storage cost, orphaned objects). A quick Eventually that the associated VolumeSnapshot is also gone would close that loop.

**2. No negative case.** The spec only proves "expired → deleted." It doesn't prove "not-yet-expired → NOT deleted," which is the property that actually protects user data. A sibling spec that creates a backup, waits a reconcile interval or two without marking it expired, and asserts it's still present would guard against a regression where the controller becomes overly aggressive (e.g., a future refactor inverts the IsExpired check). Cheap to add — same fixture, no MarkExpired call.

WentingWu666666 (approved):

hossain-rayhan (approved):

PR #394, opened 2026-06-01
fix: clear stale promotionToken after failover and add cluster-replication e2e tests

### LINKED ISSUE #375: Multi-cloud failover: WAL timeline divergence after promotion leaves replica unrecoverable

WentingWu666666:
## Summary

Fix the stale `promotionToken` bug reported in #375 and add comprehensive end-to-end tests for cluster replication (multi-region / cross-cluster replication) running within a single Kind cluster.

## Bug fix

**Stale `promotionToken` after failover (#375 sub-issue 2):**
After a successful primary swap, the operator left `spec.replica.promotionToken` populated on the promoted cluster's CNPG resource. CNPG then reported "Cluster is unrecoverable / Promotion token content is not correct", blocking subsequent failovers and causing WAL timeline divergence.

**Root cause:** `syncReplicationChanges()` in `physical_replication.go` only set the token during promotion but never cleared it once promotion succeeded — the `primaryChanged` flag was `false` on subsequent reconciliations.

**Fix:** Added logic in `syncReplicationChanges()` to detect when `current.PromotionToken != desired.PromotionToken` (desired is always empty) and issue a ReplicaCluster replace patch to clear the stale token.

**Also fixed:** `backup_controller.go` was using the DocumentDB CR name as the CNPG cluster name instead of `ReplicationContext.CNPGClusterName`, causing backup creation to fail for replicated clusters.

### 1. Cluster replication tests (`cluster_replication_test.go`) — 9 ordered specs:
- **Data replication:** CNPG config verification, insert propagation, content fidelity, update propagation
- **Failover:** seed data, promote replica, verify data on new primary, write to new primary, verify reverse replication

### 2. Issue #375 regression tests (`timeline_divergence_test.go`) — 7 specs:
- Sub-issue 1: Rapid back-to-back failover (A→B→A) and WAL timeline recovery
- Sub-issue 2: `promotionToken` is cleared after successful promotion
- Sub-issue 3: `instancesPerNode` applied correctly to both primary and replica

### 3. Backup tests (`backup_test.go`) — 3 ordered specs:
- Backup created on primary, not on replica
- After failover, backup created on promoted primary

### Test infrastructure
- Mixin template, ExternalName bridge services, helper functions, labels & timeouts
- CI integration: `cluster-replication` added to E2E matrix (amd64 + arm64)

## Fixes
- Fixes: https://github.com/documentdb/documentdb-kubernetes-operator/issues/375


test/e2e/tests/cluster_replication/helpers_test.go
```diff
@@ -0,0 +1,182 @@
...
+		primaryExpected+"-rw",
+		fmt.Sprintf("%s-rw.%s.svc.cluster.local", replicaActual, ns),
+	)
+}
+
+// findCNPGCluster discovers the CNPG Cluster backing a DocumentDB CR.
+// In replication mode, the CNPG cluster name is a hash-based derivative
+// of the DocumentDB name (via generateCNPGClusterName), so we list all
+// CNPG Clusters in the namespace and match by the documentdb ownership
+// label.
```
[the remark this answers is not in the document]
WentingWu666666: Good catch — fixed. Updated the comment to say "match by OwnerReferences (Kind=="DocumentDB" && Name==ddName)" which accurately describes the implementation.

test/e2e/tests/cluster_replication/helpers_test.go
```diff
@@ -0,0 +1,182 @@
...
+	return dd
+}
+
+// cnpgClusterName mirrors the operator's generateCNPGClusterName logic:
+// "<docdbName>-<fnv64a(memberName) hex>", capped at 50 characters.
+func cnpgClusterName(docdbName, memberName string) string {
+	const maxLen = 50
+	h := fnv.New64a()
+	h.Write([]byte(memberName))
+	maxPrefix := maxLen - 9 // dash + 16-char hex
```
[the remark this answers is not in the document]
WentingWu666666: Good catch — the subtraction `- 9` reserves 1 dash + up to 8 hex chars (since the result is capped at maxLen=50). Updated the comment to: "dash + 8-char hex (truncated by maxLen cap)".

test/e2e/pkg/e2eutils/timeouts/timeouts.go
```diff
@@ -94,10 +113,10 @@ func For(op Op) time.Duration {
 // 10-second poll to reduce API-server churn during long waits.
 func PollInterval(op Op) time.Duration {
 	switch op {
-	case MongoConnect, ServiceReady:
+	case MongoConnect, ServiceReady, ClusterReplicationDataSync:
 		return 2 * time.Second
 	case DocumentDBReady, DocumentDBUpgrade, InstanceScale,
-		PVCResize, BackupComplete, RestoreComplete:
+		PVCResize, BackupComplete, RestoreComplete, ClusterReplicationFailover:
 		return 10 * time.Second
```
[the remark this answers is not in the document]
WentingWu666666: Good point — ClusterReplicationReady is a 10-minute wait so a 10s poll is appropriate. Moved it into the 10-second poll-interval case alongside the other slow cluster-level operations.

test/e2e/tests/cluster_replication/timeline_divergence_test.go
```diff
@@ -0,0 +1,425 @@
...
+
+		// Check if the cluster is in a healthy state
+		for _, cond := range cnpg.Status.Conditions {
+			if cond.Type == string(cnpgv1.ConditionContinuousArchiving) {
+				g.Expect(cond.Status).To(Equal("True"),
+					"CNPG cluster %s should have healthy continuous archiving", ddName)
+			}
+		}
+	}
+}
```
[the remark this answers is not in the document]
WentingWu666666: Valid — renamed to `assertCNPGContinuousArchivingHealthy` and updated the doc comment to clearly state it only checks the ContinuousArchiving condition, not streaming replication directly.

test/e2e/pkg/e2eutils/timeouts/timeouts.go
```diff
@@ -57,6 +73,9 @@ var documentDBDefaults = map[Op]time.Duration{
 	RestoreComplete:   15 * time.Minute,
 	MongoConnect:      30 * time.Second,
 	ServiceReady:      2 * time.Minute,
+	ClusterReplicationReady:  10 * time.Minute,
+	ClusterReplicationDataSync:     3 * time.Minute,
+	ClusterReplicationFailover:     10 * time.Minute,
```
hossain-rayhan, line 78: These timeouts are in minutes. Can we add a one-liner comment for all of them to make easier to understand.

hossain-rayhan: ignore. I see the comments above

hossain-rayhan (approved):

test/e2e/tests/cluster_replication/cluster_replication_test.go
```diff
@@ -0,0 +1,440 @@
...
+			Expect(err).ToNot(HaveOccurred(), "creating primary DocumentDB")
+			DeferCleanup(func(ctx SpecContext) {
+				_ = documentdb.Delete(ctx, c, primaryDD, 3*time.Minute)
+			})
+
+			By("waiting for the primary to become Ready")
+			primaryKey := types.NamespacedName{Namespace: ns, Name: primaryName}
+			Eventually(assertions.AssertDocumentDBReady(ctx, c, primaryKey),
+				timeouts.For(timeouts.DocumentDBReady),
+				timeouts.PollInterval(timeouts.DocumentDBReady),
```
alaye-ms, line 74: Could we poll more frequently than 5 minutes? Might be able to save some time even just reducing it to 1 or 2.5

alaye-ms: Same comment below as well

test/e2e/tests/cluster_replication/helpers_test.go
```diff
@@ -0,0 +1,196 @@
...
+//
+//	<replicaCRName>-<hash(primaryMemberName)>-rw
+//
+// but the actual primary service is named
+//
+//	<primaryCRName>-<hash(primaryMemberName)>-rw
+//
+// This function creates ExternalName services that CNAME the expected
+// names to the actual service FQDNs.
+func createReplicationBridgeServices(ctx context.Context, c client.Client, ns, primaryCR, replicaCR string) {
```
alaye-ms, line 122: This is really cool, and a very clean way to test the multi-cluster setup without actually needing to use multiple clusters

alaye-ms (approved):

test/e2e/tests/cluster_replication/cluster_replication_test.go
WentingWu666666: Thanks for the review! Just to clarify — the 5 minutes here is the timeout (max wait), not the poll frequency. The actual polling interval is **10 seconds**, set by `timeouts.PollInterval(timeouts.DocumentDBReady)`.

So `Eventually(check, 5min, 10s)` means: "retry every 10s, fail after 5min."

PR #399, opened 2026-06-09
DOCS: Adds documentation for multi-cloud deployments

alaye-ms:


docs/operator-public-documentation/preview/multi-region-deployment/multi-cloud.md
```diff
@@ -0,0 +1,170 @@
...
+
+The playground uses Istio as the reference approach because it works across AKS,
+GKE, and EKS without requiring one provider's private networking model to cover
+all member Kubernetes clusters. For implementation details, see the
+[multi-cloud deployment playground](https://github.com/documentdb/documentdb-kubernetes-operator/tree/main/documentdb-playground/multi-cloud-deployment)
+and the upstream [Istio multi-primary multi-network documentation](https://istio.io/latest/docs/setup/install/multicluster/multi-primary_multi-network/).
+
+For operator-level networking configuration, including the
+`crossCloudNetworkingStrategy` field, see the
+[multi-region setup guide](setup.md#networking-management).
```
WentingWu666666, line 97: Quick question on this section — I might be reading the API wrong, but my understanding is that `crossCloudNetworkingStrategy` only accepts Istio, AzureFleet, or None. The 4 approaches listed here seem to mix two different things:
 - Istio multi-cluster mesh — maps directly to strategy: Istio
 - Site-to-site VPN / cloud interconnect / public LBs — read more like underlying network transports rather than operator strategies?

If that's right, then for multi-cloud Istio would be the only operator-supported strategy, and the other three would still need strategy: Istio set on top to actually do cross-cluster service discovery — otherwise the operator would fall back to None and not know how to route replication. A reader picking "Site-to-site VPN" alone might end up in that state.

Could you clarify whether the section is meant to list operator strategies, network transports, or both? Happy to be wrong here — want to make sure I understand before suggesting changes.

alaye-ms: The multi-region docs already talk extensively about the operator modes for inter-connectivity, so this section is meant to talk about how to use those or other strategies for the inter-kubernetes cluster networking. For the crossCloudNetworkingStrategy I link here to the multi-region docs, which talks about the operator side. If it makes it clearer, I can fold the Istio point into the public load balancer point, as that's what Istio does.

docs/operator-public-documentation/preview/multi-region-deployment/multi-cloud.md
```diff
@@ -0,0 +1,170 @@
...
+- **Site-to-site VPNs:** Connect cloud networks with VPN tunnels when your
+  organization needs private IP connectivity but doesn't have one native
+  provider network that spans every Kubernetes cluster.
+- **Cloud interconnect or private WAN:** Use dedicated connectivity when latency,
+  bandwidth, or compliance requirements exceed what internet-routed VPN or mesh
+  gateways can provide.
+- **Public load balancers with strict controls:** Use provider load balancers
+  only when private connectivity isn't available, and pair them with TLS,
+  firewall restrictions, and tightly scoped access.
+
```
xgerman, line 86: would add mesh network like Tailscale, Teleport, and such... not sure if we need to go into the OSS solutions like headcsale or nebula (see https://github.com/slackhq/nebula) or similar

docs/operator-public-documentation/preview/multi-region-deployment/multi-cloud.md
```diff
@@ -0,0 +1,170 @@
...
+The playground uses Istio as the reference approach because it works across AKS,
+GKE, and EKS without requiring one provider's private networking model to cover
+all member Kubernetes clusters. For implementation details, see the
+[multi-cloud deployment playground](https://github.com/documentdb/documentdb-kubernetes-operator/tree/main/documentdb-playground/multi-cloud-deployment)
+and the upstream [Istio multi-primary multi-network documentation](https://istio.io/latest/docs/setup/install/multicluster/multi-primary_multi-network/).
+
```
xgerman, line 94: We might need a note that Istio was chosen for simoicity but for best performance Cx should consult witj networking professionals on the best approach.

docs/operator-public-documentation/preview/multi-region-deployment/multi-cloud.md
```diff
@@ -0,0 +1,170 @@
...
+in the other Kubernetes clusters.
+
+If you use Istio, remote secrets and east-west gateways provide the discovery
+and routing path. If you use site-to-site VPNs or private WAN connectivity, make
+sure DNS zones, conditional forwarding, firewall rules, and route tables are
+configured consistently across providers.
+
+### Identity and permissions are provider-specific
+
+Multi-region deployments within one provider often share one IAM model. In a
```
xgerman: Please write IAM (Idendity and Sccess Managment) - explain abbreviations

docs/operator-public-documentation/preview/multi-region-deployment/multi-cloud.md
```diff
@@ -0,0 +1,170 @@
...
+DNS resolution, route changes, packet loss, or certificate trust issues. In
+addition to monitoring [replication lag](overview.md#replication-lag) and
+DocumentDB cluster health, monitor:
+
+- Istio control plane and east-west gateway health.
+- VPN or interconnect tunnel status.
+- Cross-provider latency and packet loss.
+- Provider load balancer health.
+- DNS record propagation and client resolution.
+
```
xgerman, line 155: basic connectivity between database nodes and application layer
Monitor which gateways an application conencts to to avoid potential cross-cloud latency

xgerman (approved):

WentingWu666666 (approved):

PR #400, opened 2026-06-10
docs(longhaul): add long-haul test design document

WentingWu666666:
Part 1/5 of #348 split.

## Scope
Adds `docs/designs/long-haul-test-design.md` (367 lines, new file).

## Content
- Goals and non-goals for the long-haul test
- Architecture diagram (writer/verifier loop, operations scheduler, monitor, journal, report)
- Data plane invariants: majority writes, gap detection, checksum validation
- Failure modes and disruption window policy
- HA-gated upgrade scenario, `spec.instancesPerNode` scaling
- Relationship to `test/e2e`

## Verification
Docs-only; no build/test impact.

## Related
Splits #348 into 5 focused PRs:
- **1/5 (this PR)**: design doc
- **5/5 (#401)**: `test/shared` module extraction + e2e migration (can land in parallel)
- 2/5: long-haul driver code (depends on 5/5)
- 3/5: CI/CD workflows (depends on 2/5)
- 4/5: auto-upgrade + verification (depends on 2/5 + 3/5)

### DIFF
--- a/docs/designs/long-haul-test-design.md
+++ b/docs/designs/long-haul-test-design.md
@@ -124,6 +124,8 @@ Losing an acknowledged write or observing a checksum mismatch is a Tier-1 failur
 | **Degraded** | Operator pod restart, write timeout inside an expected disruption window | Log and continue if recovery within budget |
 | **Warning** | Memory trending up, reconcile latency rising | Log only |
 
+A Fatal failure does **not** auto-recreate the cluster — the preserved state is what makes post-mortem possible. Recovery is manual: a maintainer reviews the journal/logs (alerted via the monitor described above), files the bug, and re-provisions the long-haul cluster as a separate operation.
+
 ---
 
 ## Learnings from Other Projects
--- a/docs/designs/long-haul-test-design.md
+++ b/docs/designs/long-haul-test-design.md
@@ -59,6 +59,8 @@ The driver supports a **two-cluster** topology so signals become attributable: a
 | Both degrade | Operator-level bug — leak in the shared operator |
 | Baseline degrades, Primary stable | Infrastructure noise — dismiss |
 
+**Realistic cluster configuration.** Both clusters are deployed with production-style settings: `podAntiAffinity` (one replica per node) and a `PodDisruptionBudget` that keeps a majority of replicas available during voluntary disruptions. Without these, chaos and upgrade operations would surface misconfiguration failures (e.g. quorum lost because all replicas landed on the drained node) instead of the operator/DB bugs we are trying to catch.
+
 ---
 
 ## Lifecycle
--- a/docs/designs/long-haul-test-design.md
+++ b/docs/designs/long-haul-test-design.md
@@ -140,6 +140,8 @@ A Fatal failure does **not** auto-recreate the cluster — the preserved state i
 | **CloudNative-PG** | Failover via pod delete + SIGSTOP; pod-level resource sampling | Ginkgo framework (we use a long-lived `Deployment` instead) |
 | **CockroachDB** | Chaos runner; separate workload from disruption; roachstress | Custom roachtest framework (too heavy) |
 | **Vitess** | Background stress goroutine; per-query tracking | No fault injection (we need disruptive ops) |
+| **FoundationDB** | Property-based oracle (acked-write invariants); strict workload/fault separation | Deterministic in-process simulation — requires the entire system written in their Flow actor language; runs in simulated time, so it can't surface accumulation bugs that need real wall-clock cycles |
+| **Antithesis** | Same property-based oracle philosophy applied to unmodified binaries | Deterministic hypervisor — runs in simulated time on a fake network/disk, so it targets rare-interleaving logic bugs rather than the wall-clock accumulation bugs long-haul exists to catch |
 
 **Universal pattern:** Separate workload from disruptions, run concurrently, verify against an acknowledged-write oracle, use per-operation disruption budgets.
 
--- a/docs/designs/long-haul-test-design.md
+++ b/docs/designs/long-haul-test-design.md
@@ -145,6 +145,6 @@ A Fatal failure does **not** auto-recreate the cluster — the preserved state i
 
 ---
 
-## Open Questions
+## Future Scope
 
-1. Multi-region canary scope — AKS Fleet integration?
+- **Multi-region canary** — extend the Primary/Baseline pattern across regions via AKS Fleet to catch issues that only appear with cross-region replication / failover.
--- a/docs/designs/long-haul-test-design.md
+++ b/docs/designs/long-haul-test-design.md
@@ -120,6 +120,10 @@ Losing an acknowledged write or observing a checksum mismatch is a Tier-1 failur
 
 **Human-in-the-loop alerts.** The hourly monitor posts a summary to the workflow run and, when configured, to a chat channel. A maintainer reviews the evidence and manually creates a GitHub issue. No auto-filed issues — alert fatigue from transient or infrastructure failures would erode trust in the canary.
 
+### Artifact Retention
+
+Two tiers of evidence are kept: a rolling status summary in a `longhaul-report` ConfigMap polled by the monitor workflow, and a forensics bundle (pod logs, events, CR snapshots, metric samples, the journal) uploaded as a GitHub Actions artifact on every Tier-1 / Tier-2 alert and at end of run. Implementation details — which collectors, sanitization rules, bundle layout — live in `test/longhaul/README.md`.
+
 ### Failure Tiers
 
 | Tier | Example | Action |
--- a/docs/designs/long-haul-test-design.md
+++ b/docs/designs/long-haul-test-design.md
@@ -84,7 +84,7 @@ The scheduler picks operations from these categories with weighted randomization
 | **Topology** | scale up, scale down (within CRD bounds for `spec.instancesPerNode`) |
 | **Lifecycle** | DocumentDB version upgrade, operator upgrade |
 | **HA** | controlled failover |
-| **Chaos** | kill primary pod, drain node |
+| **Chaos** | kill primary pod, drain node, kill operator pod |
 | **Data protection** | trigger backup, verify backup |
 
 **Sequencing invariants** (enforced by the scheduler — exact values live in code):
--- a/docs/designs/long-haul-test-design.md
+++ b/docs/designs/long-haul-test-design.md
@@ -67,9 +67,9 @@ The driver supports a **two-cluster** topology so signals become attributable: a
 
 ## Lifecycle
 
-The test runs **continuously** — no cycles, no resets. Workload, metrics, operations, and health monitoring all run as long-lived processes. The system accumulates real state (PVC growth, CR history, operator memory) exactly as it would in production.
+The test runs **continuously** — no cycles, no scheduled resets. Workload, metrics, operations, and health monitoring all run as long-lived processes. The system accumulates real state (PVC growth, CR history, operator memory) exactly as it would in production. The cluster is only re-provisioned manually after a Fatal failure (see Failure Tiers).
 
-**Workload runs through upgrades.** No drain, no quiesce. Draining before upgrade hides exactly the upgrade-under-state bugs we're testing.
+**Workload runs through upgrades.** Both operator and DocumentDB upgrades fire while the workload is live — no drain, no quiesce. Draining before upgrade hides exactly the upgrade-under-state bugs we're testing. Downgrades are not exercised because neither path is a supported product transition.
 
 **Baseline gate before upgrades.** Upgrades are triggered by the operator release workflow, but don't fire immediately. The harness enforces a minimum accumulation period (default 48h) since the last upgrade — ensuring we always test "upgrade after accumulated state". If multiple versions arrive while the gate is closed, only the latest executes.
 
--- a/docs/designs/long-haul-test-design.md
+++ b/docs/designs/long-haul-test-design.md
@@ -1,0 +1,144 @@
+# Long Haul Test Design
+
+**Issue:** [#220](https://github.com/documentdb/documentdb-kubernetes-operator/issues/220)
+
+---
+
+## Terminology
+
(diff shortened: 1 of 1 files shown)


docs/designs/long-haul-test-design.md
```diff
@@ -0,0 +1,144 @@
...
+
+**Per-component attribution.** Metrics are tagged by component (operator pod RSS, DB pod RSS, goroutine count, reconcile rate, API-call rate). Without separate series, a memory climb at hour 30 is undiagnosable.
+
+**Human-in-the-loop alerts.** The hourly monitor posts a summary to the workflow run and, when configured, to a chat channel. A maintainer reviews the evidence and manually creates a GitHub issue. No auto-filed issues — alert fatigue from transient or infrastructure failures would erode trust in the canary.
+
+### Failure Tiers
+
+| Tier | Example | Action |
+|---|---|---|
+| **Fatal** | Acknowledged write lost, checksum mismatch, cluster unrecoverable past budget | Preserve cluster + exit non-zero |
```
hossain-rayhan, line 132: What's the next step after that. It creates an alarm and recreates the cluster automatically?

hossain-rayhan (approved):

docs/designs/long-haul-test-design.md
WentingWu666666: Good question  neither, by design.

- **Cluster is preserved deliberately.** The post-mortem usually depends on the exact state at the moment of failure (operator logs, CR status, replication lag, journal). Auto-recreating would destroy the evidence we wrote the test to capture.
- **Alerting is human-in-the-loop**, per the *Human-in-the-loop alerts* section above. The hourly monitor picks up the non-zero exit / unhealthy state and posts to the workflow summary + chat channel.
- **Recovery is manual** after triage  fix the bug, then re-provision the long-haul cluster as a separate operation.

Added a clarification sentence under the Failure Tiers table in the latest commit so future readers don't have to ask.

docs/designs/long-haul-test-design.md
```diff
@@ -0,0 +1,146 @@
...
+
+| Component | Role | Output |
+|---|---|---|
+| **Writer/Verifier** | Data-plane workload. Connects via `mongodb://` only — no k8s imports. Writers insert monotonic sequences with checksums under majority write concern; verifiers scan for gaps and bad checksums. | Counters (acked, failed, verify passes, gaps, checksum errors); errors to journal. |
+| **Operation Scheduler** | Control plane. Applies weighted-random ops (scale, kill, failover, backup, upgrade) with preconditions and cooldowns. | Operation start/end events to journal. |
+| **Monitor** | Polls pod RSS/CPU and checks readiness of operator + DB pods. | Periodic samples + readiness events to journal. |
+| **Journal** | In-process append-only event log shared by all components. | Reproducible event stream for the report. |
+| **Report** | Aggregates the journal into a markdown summary at a configurable interval; raises alerts on threshold breaches. | Markdown report; alert lines. |
+
+### Cluster Topology
```
xgerman, line 54: we should mention that where possible we want to reuse the code from the e2e tests (e.g. client)

docs/designs/long-haul-test-design.md
```diff
@@ -0,0 +1,146 @@
...
+|---|---|
+| Primary degrades, Baseline stable | Per-cluster bug — caused by operations on Primary |
+| Both degrade | Operator-level bug — leak in the shared operator |
+| Baseline degrades, Primary stable | Infrastructure noise — dismiss |
+
+---
+
+## Lifecycle
+
+The test runs **continuously** — no cycles, no resets. Workload, metrics, operations, and health monitoring all run as long-lived processes. The system accumulates real state (PVC growth, CR history, operator memory) exactly as it would in production.
```
xgerman: Please eleborate how we deal with different versions, e.g. are we always runnign the latest, current, etc. When are we updating? Part of the test?

Is there a point when we start over? What's the criteria?

docs/designs/long-haul-test-design.md
```diff
@@ -0,0 +1,146 @@
...
+
+**Workload runs through upgrades.** No drain, no quiesce. Draining before upgrade hides exactly the upgrade-under-state bugs we're testing.
```
xgerman: Aevwe also downgrading? Or are we starting over at some point so we can test upgrade more than once?

docs/designs/long-haul-test-design.md
```diff
@@ -0,0 +1,146 @@
...
+
+The scheduler picks operations from these categories with weighted randomization:
+
+| Category | Examples |
+|---|---|
+| **Topology** | scale up, scale down (within CRD bounds for `spec.instancesPerNode`) |
+| **Lifecycle** | DocumentDB version upgrade, operator upgrade |
+| **HA** | controlled failover |
+| **Chaos** | kill primary pod, drain node |
+| **Data protection** | trigger backup, verify backup |
```
xgerman, line 88: do we have operator upgrades as well?
Operator chaos?
Remote nodes/mukti-region? (maybe not now but potentially planned in the future)

docs/designs/long-haul-test-design.md
```diff
@@ -0,0 +1,146 @@
...
+
+**Sequencing invariants** (enforced by the scheduler — exact values live in code):
+
+- One disruptive op at a time. Overlapping disruptions are non-diagnosable.
+- Per-category cooldown between ops. Lets the cluster stabilize.
+- Steady-state gate — health check must pass before the next op fires.
+- Backup isolation — no topology changes during backup.
```
xgerman: why not? Backup should block/delay - this should be handled by backup

docs/designs/long-haul-test-design.md
```diff
@@ -0,0 +1,146 @@
...
+E2E tests run for 15–60 minutes from a clean state. They cannot detect bugs whose **accumulation rate is tied to real operations** — memory leaks, lock-table bloat, CR-history drift, upgrade-under-state failures. These bugs surface only after days of continuous operation.
+
+You can't speed up "memory leaked per reconciliation cycle" — you need many real reconciliation cycles. The long-haul infrastructure (persistent cluster, event journal, alerting) exists because these tests can't be attended, can't be reset between runs, and need accumulation that no existing test type provides.
+
+---
+
+## Architecture
+
+### Components
+
```
xgerman, line 27: Not sure where this fits but we need to also specify reasonable anti-affinity policies, pod disription budgets, etc. to ensure what we are testing makes sense...

docs/designs/long-haul-test-design.md
```diff
@@ -0,0 +1,146 @@
...
+Losing an acknowledged write or observing a checksum mismatch is a Tier-1 failure regardless of what else is happening.
+
+---
+
+## Observability
+
+**Per-component attribution.** Metrics are tagged by component (operator pod RSS, DB pod RSS, goroutine count, reconcile rate, API-call rate). Without separate series, a memory climb at hour 30 is undiagnosable.
+
+**Human-in-the-loop alerts.** The hourly monitor posts a summary to the workflow run and, when configured, to a chat channel. A maintainer reviews the evidence and manually creates a GitHub issue. No auto-filed issues — alert fatigue from transient or infrastructure failures would erode trust in the canary.
+
```
xgerman, line 123: we shoudl also record the system dashboard metrics (latency, uptime, etc.) as well as logs of all components for later analysis; where do we keep them?

docs/designs/long-haul-test-design.md
```diff
@@ -0,0 +1,146 @@
...
+## Learnings from Other Projects
+
+| Project | Pattern We Adopt | Pattern We Skip |
+|---|---|---|
+| **Strimzi** | Run-until-failure loops; metrics collection | JUnit (we run a standalone Go binary, not a test framework) |
+| **CloudNative-PG** | Failover via pod delete + SIGSTOP; pod-level resource sampling | Ginkgo framework (we use a long-lived `Deployment` instead) |
+| **CockroachDB** | Chaos runner; separate workload from disruption; roachstress | Custom roachtest framework (too heavy) |
+| **Vitess** | Background stress goroutine; per-query tracking | No fault injection (we need disruptive ops) |
+
```
xgerman, line 150: We are also interested how FoundationDB tests (they turned their approach into Anithesis) - not sure if they cover long haul though

docs/designs/long-haul-test-design.md
```diff
@@ -0,0 +1,146 @@
...
+**Universal pattern:** Separate workload from disruptions, run concurrently, verify against an acknowledged-write oracle, use per-operation disruption budgets.
+
+---
+
+## Open Questions
+
+1. Multi-region canary scope — AKS Fleet integration?
```
xgerman: yes, at a later point

docs/designs/long-haul-test-design.md
WentingWu666666: Good catch — added a "Realistic cluster configuration" note under Cluster Topology in a1048a4: both Primary and Baseline run with `podAntiAffinity` (one replica per node) and a `PodDisruptionBudget` keeping a majority of replicas available during voluntary disruptions.

Without those, chaos and upgrade ops would surface misconfiguration failures (e.g. quorum lost because all replicas were packed onto the drained node) instead of the operator/DB bugs we're trying to catch.

docs/designs/long-haul-test-design.md
WentingWu666666: Added a "Code reuse" paragraph at the end of the Architecture section in ab9a009 (will update SHA in reply if push lands differently):

> Where possible, the driver consumes the same helpers as the e2e suite — the Mongo client, DocumentDB lifecycle operations (create / patch / wait-healthy / delete), and TLS plumbing all live in a shared `test/shared` Go module. This keeps long-haul behavior aligned with what e2e exercises and avoids two diverging mongo-driver wrappers.

That `test/shared` module is what #401 extracts — long-haul will consume it from day one.

docs/designs/long-haul-test-design.md
WentingWu666666: Agreed — renamed `Open Questions` -> `Future Scope` and reworded the multi-region item so it reads as explicitly deferred (still a candidate before GA if scope allows). See 8a79edd.

docs/designs/long-haul-test-design.md
WentingWu666666: Added FoundationDB and Antithesis as separate rows in the Learnings table (309aefc). Short answer to your question: neither covers long-haul — both run in **simulated time** on fake network/disk, so they catch rare-interleaving logic bugs in seconds but can't surface the wall-clock accumulation bugs (mem leaks, lock-table bloat, CR-history drift) that need real reconciliation cycles over real days. We adopt their property-based oracle and workload/fault separation.

docs/designs/long-haul-test-design.md
WentingWu666666: Added an `Artifact Retention` subsection in 314f3dc. Two tiers:

- **Rolling status** — `longhaul-report` ConfigMap polled by the monitor workflow (this part is already used in the existing driver).
- **Forensics bundle** — pod logs, events, CR snapshots, metric samples, journal — uploaded as a GitHub Actions artifact on every Tier-1 / Tier-2 alert and at end of run.

Operational details (which collectors, sanitization rules, bundle layout) are kept in `test/longhaul/README.md` rather than the design doc, so the design stays high-level.

docs/designs/long-haul-test-design.md
WentingWu666666: Agreed — dropped the bullet (d1694f5). Backup-vs-topology is the backup feature's job; isolating it in the harness would hide exactly the bugs we want to catch.

docs/designs/long-haul-test-design.md
WentingWu666666: - **Operator upgrades**: already in the Lifecycle row (`DocumentDB version upgrade, operator upgrade`).
- **Operator chaos**: added `kill operator pod` to the Chaos row in 71dd7b0.
- **Multi-region**: covered under `## Future Scope`.

docs/designs/long-haul-test-design.md
WentingWu666666: No downgrading -- neither the operator nor DocumentDB downgrade as a product path (CRD schemas and Postgres schemas are forward-only). 

If we want to test the upgrade many times, e2e is the right tool.

docs/designs/long-haul-test-design.md
WentingWu666666: Updated Lifecycle in c9aa1ce to make these explicit:

- **Versions**: always **latest** for both operator and DocumentDB — no pinning, no matrix. A workflow polls GHCR hourly: it upgrades the operator directly, and publishes the desired DocumentDB version to a ConfigMap that the driver picks up as a scheduled in-band upgrade (so writers/verifiers see the upgrade window).
- **Downgrades**: not exercised — neither operator CRD schemas nor Postgres major versions support downgrade as a product path.
- **Start-over**: no automatic resets. The cluster runs indefinitely; it's only re-provisioned manually after a Fatal failure (Failure Tiers).
- **Multiple upgrades**: each new release naturally produces another upgrade event over time, so the long-running cluster accumulates upgrades without needing a reset. Testing the upgrade *transition* many times in parallel is what e2e covers.

docs/designs/long-haul-test-design.md
WentingWu666666: Updated Lifecycle in c9aa1ce to make these explicit:

- **Versions**: always **latest** for both operator and DocumentDB — no pinning, no matrix. A workflow polls GHCR hourly: it upgrades the operator directly, and publishes the desired DocumentDB version to a ConfigMap that the driver picks up as a scheduled in-band upgrade (so writers/verifiers see the upgrade window).
- **Downgrades**: not exercised — neither operator CRD schemas nor Postgres major versions support downgrade as a product path.
- **Start-over**: no automatic resets. The cluster runs indefinitely; it's only re-provisioned manually after a Fatal failure (Failure Tiers).
- **Multiple upgrades**: each new release naturally produces another upgrade event over time, so the long-running cluster accumulates upgrades without needing a reset. Testing the upgrade *transition* many times in parallel is what e2e covers.

xgerman (approved):

PR #401, opened 2026-06-10
test: extract test/shared module from e2e helpers for long-haul reuse

WentingWu666666:
Part 5/5 of #348 split.

## Scope
Pure refactor: extracts framework-agnostic test helpers from `test/e2e/pkg/e2eutils` into a new `test/shared/` Go module, then migrates the e2e suite to consume them via re-export shims.

This is a prerequisite for the long-haul driver (PR-2, coming next) to share helpers with the e2e suite without depending on Ginkgo/Gomega.

## What moves
- `test/shared/documentdb/`  framework-agnostic CR lifecycle helpers (Get, PatchInstances, PatchSpec, WaitHealthy, Delete, List, ReadyStatus)
- `test/shared/mongo/`  Mongo client helpers (NewClient, NewFromURI, BuildURI, Ping, Seed, Count, DropDatabase, ClientOptions)
- Unit tests (Ginkgo/Gomega) for each shared package

## What stays in e2e
- Manifest rendering / template assembly (depends on test fixtures)
- Operator/health/namespaces helpers (Ginkgo-coupled)
- Ginkgo-specific assertions

## Verification
- cd test/shared && go build ./... && go test ./...  16 mongo specs + documentdb specs pass
- cd test/e2e && go build ./...  clean
- cd test/e2e && go test ./pkg/...  all 14 packages pass

## Commits (12)
- `78f54a5` extract test/shared module
- `597dff6` test/e2e: re-export CR ops from test/shared
- `6326588` test/shared: convert unit tests to Ginkgo/Gomega
- `40f5503` test/e2e: drop ReadyStatus/DefaultWaitPoll const aliases
- `2ea1cc0` test/e2e: drop documentdb var aliases, call shareddb directly
- `246f63c` test/shared/mongo: add unit tests for NewFromURI
- `b0a4d84` test/shared/documentdb: rename cr.go -> lifecycle.go
- `9e680c0` test/shared/documentdb: rename Describe/suite labels to lifecycle
- `99dad01` test/shared/mongo: merge uri.go into client.go
- `d6cbb29` test/e2e: migrate mongo helpers to test/shared/mongo
- `7686e1c` test/shared: add MIT copyright headers
- `ba80de7` test/shared: go mod tidy

## Related
Splits #348. PR-1 (#400) covers the design doc and can land in parallel. PR-2 (long-haul driver) will consume `test/shared` from the start and depends on this PR.

PR #405, opened 2026-06-16
test(longhaul): add long-haul test driver core

WentingWu666666:
## Part 2 of 5: long-haul test driver core

Part 2 of the [#348](https://github.com/documentdb/documentdb-kubernetes-operator/pull/348) split (design doc landed in #400, shared test module in #401).

### What this PR adds

A new `test/longhaul` Go module — a long-running canary driver that exercises a persistent DocumentDB cluster with continuous workload, scheduled disruptive operations, health monitoring, and reporting. See the design in `docs/designs/long-haul-test-design.md` (merged via #400) for the *why*.

Packages:

| Path | Role |
|---|---|
| `cmd/longhaul` | Driver entrypoint; wires workload + ops + monitor + reporting |
| `config` | Env-driven configuration with validation + unit tests |
| `workload` | Writer / verifier / atomic metrics (acked-write durability oracle) |
| `operations` | Weighted-random scheduler + scale/upgrade ops |
| `monitor` | Cluster client (consumes `test/shared`), health gating, leak detector |
| `journal` | Thread-safe event log + per-operation outage-policy oracle |
| `report` | Markdown report, periodic checkpoint to ConfigMap, GitHub Actions alerts |
| `deploy/setup.yaml` | Bootstrap manifest (namespace + Secret + DocumentDB CR) for canary cluster |
| `Dockerfile`, `README.md` | Container image + operator-facing documentation |

### Code reuse with e2e

Per the design doc (and #401), the driver consumes:

- `test/shared/documentdb` — `Get`, `PatchSpec`, `PatchInstances`, `WaitHealthy`, `Delete`, `ReadyStatus` (no duplicated CR logic; no hardcoded readiness strings)
- `test/shared/mongo` — `NewFromURI`, `Ping`, `PingWithRetry`, TLS plumbing (connection setup only; the workload writer/verifier deliberately use raw `InsertOne`/`Find` because the durability oracle needs per-doc semantics — `Seed`/`Count` cannot express acked-write reconciliation or per-doc checksums)

`go.mod` uses `replace` for both modules so changes to shared helpers flow into the driver immediately.

### Excluded from this PR (follow-ups)

- **PR-3**: CI/CD workflows (`.github/workflows/longhaul-*.yaml`) + `deploy/deployment.yaml` + `deploy/rbac.yaml`
- **PR-4**: auto-upgrade logic for operator + DocumentDB version drift

### Verification

```
cd test/longhaul
go build ./...   # clean
go test ./...    # 6/6 packages pass: config / journal / monitor / operations / report / workload
```

### Reviewer notes

- The whole driver is dormant until #348's CI workflows land in PR-3 — this PR is "the binary exists and its unit tests pass". No CI runs the driver yet.
- `monitor/k8sclient.go` is the typed `controller-runtime` wrapper that addresses xgerman's #348 review feedback (dropped raw `dynamic.Interface` + hardcoded GVR).
- `operations/upgrade.go` reads the desired DocumentDB version from a `longhaul-versions` ConfigMap that PR-4's workflow will populate; today it no-ops if the ConfigMap is missing.


test/longhaul/workload/writer.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+// computeChecksum creates a deterministic hash of the write for verification.
+func computeChecksum(writerID string, seq int64, payload string) string {
+	data := fmt.Sprintf("%s:%d:%s", writerID, seq, payload)
+	hash := sha256.Sum256([]byte(data))
+	return hex.EncodeToString(hash[:8])
+}
+
+// StartWriters launches n writers and returns a cancel function to stop them.
```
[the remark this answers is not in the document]
WentingWu666666: Fixed in 3e9ac70 — doc comment now says `StartWriters` returns the slice of writers and relies on ctx cancellation.

test/longhaul/cmd/longhaul/main.go
```diff
@@ -0,0 +1,278 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// Package main provides a standalone binary entry point for running
+// long haul tests as a Kubernetes Job (without Ginkgo test framework).
+package main
```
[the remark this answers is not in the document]
WentingWu666666: Fixed in 3e9ac70 — header now says `Kubernetes Deployment` and notes that the ConfigMap / GH annotations are the source of truth (since the pod auto-restarts on crash).

test/longhaul/cmd/longhaul/main.go
```diff
@@ -0,0 +1,278 @@
...
+		log.Fatalf("MongoDB ping failed: %v", err)
+	}
+	log.Println("MongoDB connection established")
+
+	db := mongoClient.Database("longhaul")
+
+	// Drop previous test data to avoid duplicate key conflicts.
+	if err := db.Collection(workload.CollectionName).Drop(ctx); err != nil {
+		log.Fatalf("failed to drop collection: %v", err)
+	}
```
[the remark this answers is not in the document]
WentingWu666666: Fixed in 3e9ac70 — drop is now gated behind a new `LONGHAUL_RESET_DATA` env var (default `false`). Also fixed the deeper bug your comment hinted at: writers now seed their seq counter from `FindOne().Sort(seq:-1)` on startup, so a Deployment-driven pod restart resumes past the prior tip instead of colliding with the unique `(writer_id, seq)` index.

test/longhaul/operations/upgrade.go
```diff
@@ -0,0 +1,178 @@
...
+		return fmt.Errorf("read desired version: %w", err)
+	}
+	if desired == "" {
+		return fmt.Errorf("desired version is empty")
+	}
+
+	running, _ := u.client.GetCurrentDocumentDBImageTag(ctx)
+	if err := u.client.UpgradeDocumentDB(ctx, desired); err != nil {
+		return fmt.Errorf("patch CR: %w", err)
+	}
```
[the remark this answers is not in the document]
WentingWu666666: Fixed in 3e9ac70 — the read error is now wrapped and returned (`read current image tag: %w`).

test/longhaul/operations/scheduler.go
```diff
@@ -0,0 +1,176 @@
...
+	j *journal.Journal,
+	cooldown time.Duration,
+) *Scheduler {
+	return &Scheduler{
+		operations:    ops,
+		healthMonitor: health,
+		journal:       j,
+		cooldown:      cooldown,
+	}
+}
```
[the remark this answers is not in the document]
WentingWu666666: Fixed in 3e9ac70 — switched to `math/rand/v2` which is auto-seeded per-process from a random source. (Side note: since Go 1.20 the top-level `math/rand` global source is also auto-seeded, but `rand/v2` is the recommended path going forward and the project is on Go 1.25.)

test/longhaul/report/checkpoint.go
```diff
@@ -0,0 +1,139 @@
...
+
+	ticker := time.NewTicker(r.interval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			// Final report on exit.
+			r.emit(context.Background())
+			return
```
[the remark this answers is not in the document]
WentingWu666666: Fixed in 3e9ac70 — added `EmitFinal()` that uses a 30s bounded context, called synchronously from `main.run()` before exit. The `ctx.Done()` branch in `Run` no longer emits at all (it was racing with `os.Exit` and the verdict never reached the ConfigMap).

test/longhaul/report/alert.go
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package report
+
+import (
+	"fmt"
+	"os"
+	"time"
+)
```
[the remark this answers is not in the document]
WentingWu666666: Fixed in 3e9ac70 — added `strings` import and `escapeAnnotation` helper.

test/longhaul/report/alert.go
```diff
@@ -0,0 +1,44 @@
...
+
+	switch s.Result {
+	case ResultFail:
+		msg := "Long haul test FAILED"
+		if s.FailReason != "" {
+			msg = fmt.Sprintf("Long haul test FAILED: %s", s.FailReason)
+		}
+		// ::error:: annotations show as red in the Actions UI.
+		fmt.Printf("::error title=Long Haul Test Failure::%s\n", msg)
+
```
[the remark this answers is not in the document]
WentingWu666666: Fixed in 3e9ac70 — all three annotation messages (error / notice / warning) now go through `escapeAnnotation` which replaces `%` -> `%25`, `\r` -> `%0D`, `\n` -> `%0A` per the GitHub Actions workflow-command spec.

test/longhaul/deploy/setup.yaml
```diff
@@ -0,0 +1,78 @@
...
+# The operator creates the MongoDB user with these credentials.
+apiVersion: v1
+kind: Secret
+metadata:
+  name: documentdb-credentials
+  namespace: documentdb-test-ns
+type: kubernetes.io/basic-auth
+stringData:
+  username: longhaul-user
+  password: LongHaul-T3st-Passw0rd!
```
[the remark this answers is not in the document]
WentingWu666666: Fixed in 3e9ac70 — replaced with `__LONGHAUL_PASSWORD__` placeholder, plus a comment showing the recommended `sed -i "s/__LONGHAUL_PASSWORD__/$(openssl rand -base64 24)/" deploy/setup.yaml` substitution (or generating the secret out-of-band with `kubectl create secret`).

test/longhaul/README.md
```diff
@@ -0,0 +1,173 @@
...
+| `LONGHAUL_RECOVERY_TIMEOUT` | No | `5m` | Max wait for cluster recovery after an operation. |
+| `LONGHAUL_MIN_INSTANCES` | No | `1` | Minimum `spec.instancesPerNode` for scale-down operations (CRD lower bound: 1). |
+| `LONGHAUL_MAX_INSTANCES` | No | `3` | Maximum `spec.instancesPerNode` for scale-up operations (CRD upper bound: 3). |
+| `LONGHAUL_REPORT_INTERVAL` | No | `1h` | How often to write checkpoint reports to ConfigMap. |
+
+## CI Safety
+
+The long haul test binary is deployed as a Kubernetes Job on a dedicated AKS cluster.
+It does **not** run in any PR-gated CI workflow.
+
```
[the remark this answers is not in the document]
WentingWu666666: Fixed in 3e9ac70 — CI Safety section now says `Kubernetes Deployment` and explicitly calls out that the ConfigMap / GH annotations are the source of truth because the pod auto-restarts.

test/longhaul/README.md
```diff
@@ -0,0 +1,173 @@
...
+| | `test/e2e/` | `test/longhaul/` |
+|---|---|---|
+| Shape | Go test binary (Ginkgo specs) | Standalone long-running daemon |
+| Lifetime | Minutes per spec | Days–weeks per run |
+| Asserts | One behavior per spec, then exits | Continuous invariants over time |
+| Failure mode | `t.Fail` per spec | Journal entry + alert + auto-restart |
+| Cluster | Created + torn down per run | Long-lived dedicated AKS cluster |
+| Operator API | Typed (`previewv1.DocumentDB` via controller-runtime) | Dynamic client (no operator import) |
+
```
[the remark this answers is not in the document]
WentingWu666666: Fixed in 3e9ac70 — table row now reads `Typed (previewv1.DocumentDB via controller-runtime + test/shared/documentdb helpers)` to match the current monitor implementation.

test/longhaul/README.md
```diff
@@ -0,0 +1,173 @@
...
+  `LONGHAUL_MONGO_URI` string; when it moves to per-secret credentials or in-cluster TLS,
+  these helpers become directly applicable.
+- `e2eutils/operatorhealth` — pod-ready / CRD-ready gating used during e2e setup. The
+  monitor's `isPodReady` could delegate to this when the modules are unified.
+- `e2eutils/clusterprobe` — CRD presence checks.
+
+A shared `test/shared/` module is **deliberately not introduced yet**: the modules' Go and
+dependency versions differ today, and the only currently-duplicated surface (raw mongo
+connect + ping) is too small to justify the third-module overhead. Revisit this when the
+long haul driver adopts the same connection model as e2e.
```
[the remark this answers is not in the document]
WentingWu666666: Fixed in 3e9ac70 — removed the `deliberately not introduced yet` paragraph and replaced it with a `Code that is shared today` subsection listing `test/shared/documentdb` and `test/shared/mongo`, plus a `Future opportunities` subsection for the remaining e2e helpers.

test/longhaul/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+**Manual path (one-off / local cluster):**
+
+```bash
+cd test/longhaul
+
+# 1. Build and push the container image (or use the GHCR image from CI).
+docker build -t <your-registry>/longhaul-test:latest -f Dockerfile .
+docker push <your-registry>/longhaul-test:latest
+
```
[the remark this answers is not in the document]
WentingWu666666: Fixed in 3e9ac70 — README now instructs `cd <repo-root>` and `docker build -f test/longhaul/Dockerfile .` (build context = repo root) so the `replace` paths to `../shared` and `../../operator/src` resolve.

test/longhaul/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+# Cache module downloads.
+COPY go.mod go.sum ./
+RUN go mod download
+
+# Copy source.
+COPY . .
+
+# Build the standalone binary.
+RUN CGO_ENABLED=0 GOOS=linux go build -ldflags="-s -w" -o /longhaul ./cmd/longhaul
```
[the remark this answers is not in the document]
WentingWu666666: Fixed in 3e9ac70 — Dockerfile now explicitly copies `operator/src`, `test/shared`, and `test/longhaul`, runs `go mod download` after sources are in place (so replace paths resolve), and bumped the builder image from `golang:1.26-alpine` to `golang:1.25-alpine` to match the repo toolchain. Added a header comment showing the correct `docker build -f test/longhaul/Dockerfile .` invocation.

test/longhaul/deploy/setup.yaml
```diff
@@ -0,0 +1,78 @@
...
+  nodeCount: 1
+  instancesPerNode: 2
+  schemaVersion: "auto"
+  documentDbCredentialSecret: documentdb-credentials
+  resource:
+    storage:
+      pvcSize: 10Gi
+  exposeViaService:
+    serviceType: LoadBalancer
+  sidecarInjectorPluginName: cnpg-i-sidecar-injector.documentdb.io
```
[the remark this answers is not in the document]
WentingWu666666: Fixed in 3e9ac70 — replaced `sidecarInjectorPluginName: ...` with the typed-API path `spec.plugins.sidecarInjectorName: ...` to match the `PluginsSpec` struct in `api/preview/documentdb_types.go`.

test/longhaul/report/checkpoint.go
```diff
@@ -0,0 +1,139 @@
...
+
+func (r *CheckpointReporter) emit(ctx context.Context) {
+	summary := r.summaryFunc()
+
+	// Mark as RUNNING for intermediate checkpoints (unless already FAIL).
+	resultStr := string(summary.Result)
+	if summary.Result == ResultPass {
+		resultStr = "RUNNING"
+	}
+
```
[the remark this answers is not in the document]
WentingWu666666: Fixed in 3e9ac70 — `emit` now takes a `final bool` parameter. On the final shutdown emit (`EmitFinal`) a `PASS` result is persisted as `PASS`; only intermediate ticker-driven emits surface `PASS` as `RUNNING`. `FAIL` is always persisted as `FAIL`.

test/longhaul/operations/scale.go
```diff
@@ -0,0 +1,135 @@
...
+func (s *ScaleUp) OutagePolicy() journal.OutagePolicy {
+	return journal.OutagePolicy{
+		AllowedDowntime:      30 * time.Second,
+		AllowedWriteFailures: 20,
+		MustRecoverWithin:    s.recovery,
+	}
+}
+
+// ScaleDown decreases spec.instancesPerNode by 1 (HA scale dimension; range 1-3).
+type ScaleDown struct {
```
xgerman: `ScaleUp` (above) and `ScaleDown` are ~95% identical: same fields, and `Precondition`/`Execute` differ only by `+1`/`-1`, `>=`/`<=`, the bound, and a couple of constants. Consider collapsing into one parameterized type to halve the file:

```go
type scaleOp struct {
    client    monitor.ClusterClient
    healthMon *monitor.HealthMonitor
    name      string
    weight    int
    delta     int        // +1 up, -1 down
    limit     int        // max (up) or min (down)
    atLimit   func(cur int) bool
    recovery  time.Duration
    policy    journal.OutagePolicy
}
```

Thin `NewScaleUp`/`NewScaleDown` constructors keep call sites unchanged.

(Also: `Name()`/`Weight()` on lines 37-38 and 98-99 fail gofmt — extra spaces.)

test/longhaul/cmd/longhaul/main.go
```diff
@@ -0,0 +1,292 @@
...
+	clientset, err := kubernetes.NewForConfig(restConfig)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to create clientset: %w", err)
+	}
+
+	return client, clientset, nil
+}
+
+func buildRESTConfig() (*rest.Config, error) {
```
xgerman: `buildRESTConfig` here re-implements `monitor.buildRestConfig` (k8sclient.go:86) almost verbatim, and `initK8sClient` then creates a **second** clientset (line 231) even though `K8sClusterClient` already built one (k8sclient.go:59). Expose the existing one instead:

```go
func (k *K8sClusterClient) Clientset() kubernetes.Interface { return k.clientset }
```

Then delete `buildRESTConfig` and the redundant `NewForConfig` — one client, one config path.

test/longhaul/monitor/k8sclient.go
```diff
@@ -0,0 +1,267 @@
...
+	}
+
+	health.ReadyPods = readyCount
+	health.AllPodsReady = readyCount == health.TotalPods && health.TotalPods > 0
+	health.RestartCount = totalRestarts
+
+	// Get the DocumentDB CR status via the shared typed helper. Using
+	// shareddb.IsHealthy keeps the readiness predicate consistent with
+	// the e2e suite (single source of truth for ReadyStatus).
+	dd, err := shareddb.Get(ctx, k.crClient, types.NamespacedName{Namespace: k.namespace, Name: k.clusterName})
```
xgerman: `shareddb.Get(ctx, k.crClient, types.NamespacedName{Namespace: k.namespace, Name: k.clusterName})` is repeated verbatim in 4 methods (`GetClusterHealth`, `GetInstancesPerNode`, `GetCurrentDocumentDBImageTag`, `UpgradeDocumentDB`). Extract a private helper:

```go
func (k *K8sClusterClient) getCR(ctx context.Context) (*previewv1.DocumentDB, error) {
    return shareddb.Get(ctx, k.crClient, types.NamespacedName{Namespace: k.namespace, Name: k.clusterName})
}
```

test/longhaul/monitor/health.go
```diff
@@ -0,0 +1,169 @@
...
+
+// Run starts the health monitoring loop. Blocks until context is cancelled.
+func (h *HealthMonitor) Run(ctx context.Context) {
+	h.journal.Info("health", "health monitor started")
+	defer h.journal.Info("health", "health monitor stopped")
+
+	ticker := time.NewTicker(healthCheckInterval)
+	defer ticker.Stop()
+
+	for {
```
xgerman, line 83: This `ticker` + `select { <-ctx.Done() / <-ticker.C }` loop is duplicated 5× across the module (here, `writer.Run`, `verifier.Run`, `scheduler.Run`, `runMetricsSampling`). Optional, since they span packages, but a tiny shared helper would DRY them:

```go
func tick(ctx context.Context, d time.Duration, fn func(context.Context)) {
    t := time.NewTicker(d); defer t.Stop()
    for { select { case <-ctx.Done(): return; case <-t.C: fn(ctx) } }
}
```

xgerman (commented):
## Review: structure, duplication & terseness

Focused pass on **code duplication, structure, and style** (the existing Copilot threads already covered correctness/security and are addressed in `3e9ac70`). The module is well-documented and `go build`/`go vet`/`go test ./...` all pass locally. 👍

### 🟠 Must fix — gofmt (CI blocker)
Four files are not gofmt-clean (extra-space alignment), so `make fmt`/lint will fail:
`operations/scale.go`, `operations/scheduler.go`, `monitor/health.go`, `report/checkpoint.go`.
Fix: `gofmt -w operations/ monitor/ report/`.

### 🟡 Duplication / structure (highest-value dedup)
1. `ScaleUp` and `ScaleDown` are ~95% identical — collapse into one parameterized type (inline).
2. `main.buildRESTConfig` re-implements `monitor.buildRestConfig` and builds a **second** clientset that `K8sClusterClient` already owns (inline).
3. `shareddb.Get(...)` repeated 4× in `k8sclient.go` — extract a `getCR` helper (inline).
4. The ticker + `select{ctx.Done/ticker.C}` loop recurs 5× (writer/verifier/scheduler/health/metrics) — optional shared helper (inline).

### 🟢 Style / terseness / dead code
- `journal.DefaultOutagePolicy()` is unused (inline).
- `report.go`: `b.WriteString(fmt.Sprintf(...))` → `fmt.Fprintf(&b, ...)` (inline).
- `checkpoint.go`: `map[string]interface{}` → `map[string]any` (inline).
- `metrics.go`: `MetricsSnapshot` doc comment should start with the type name (inline).

No blocking design issues — just the gofmt fix for CI plus the two consolidations (#1, #2) which are the biggest wins. Nice work on the durability oracle and the bounded verifier scan.

test/longhaul/journal/policy.go
```diff
@@ -0,0 +1,79 @@
...
+
+	// AllowedWriteFailures is the maximum number of write failures during the window.
+	AllowedWriteFailures int64
+
+	// MustRecoverWithin is the maximum time from operation start to full recovery.
+	MustRecoverWithin time.Duration
+}
+
+// DefaultOutagePolicy returns a conservative policy suitable for most operations.
+func DefaultOutagePolicy() OutagePolicy {
```
xgerman, line 18: `DefaultOutagePolicy()` is dead code — it's never referenced anywhere (each operation builds its `OutagePolicy` inline). Either use it as the base for the scale/upgrade policies or remove it.

test/longhaul/report/report.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+// GenerateMarkdown produces a human-readable markdown report.
+func GenerateMarkdown(s Summary) string {
+	var b strings.Builder
+
+	b.WriteString("# Long Haul Test Report\n\n")
+
+	// Header
+	b.WriteString(fmt.Sprintf("**Result:** %s\n", s.Result))
```
xgerman: `b.WriteString(fmt.Sprintf(...))` appears ~15× in this function. `fmt.Fprintf(&b, ...)` is the idiomatic, terser form (a `strings.Builder` implements `io.Writer`), e.g.:

```go
fmt.Fprintf(&b, "**Result:** %s\n", s.Result)
```

test/longhaul/report/checkpoint.go
```diff
@@ -0,0 +1,155 @@
...
+			log.Printf("[checkpoint] failed to update ConfigMap: %v", err)
+		} else {
+			log.Println("[checkpoint] ConfigMap updated")
+		}
+	} else {
+		log.Printf("[checkpoint] failed to get ConfigMap: %v", err)
+	}
+
+	// Also log the summary as JSON for structured log consumers.
+	summaryJSON, _ := json.Marshal(map[string]interface{}{
```
xgerman: `map[string]interface{}` → `map[string]any` (Go 1.18+, and go.mod is 1.25). Minor: `fmt.Printf("\n%s\n", "=== CHECKPOINT REPORT ===")` on lines 93/95 can just be `fmt.Println(...)`. Note this file also fails gofmt (the JSON map literal alignment).

test/longhaul/workload/metrics.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+// NewMetrics creates a new Metrics instance with the start time set to now.
+func NewMetrics() *Metrics {
+	return &Metrics{
+		StartTime: time.Now(),
+	}
+}
+
+// Snapshot returns a point-in-time copy of all metric values.
```
xgerman: This doc comment is attached to `type MetricsSnapshot` but reads "Snapshot returns a point-in-time copy...". Per Go convention the comment should start with the identifier it documents, e.g. `// MetricsSnapshot is a point-in-time copy of all metric values.`

test/longhaul/operations/scale.go
WentingWu666666: Fixed in ed498df — collapsed into a shared `scaleOp` struct parameterized by `delta`/`bound`/`boundKind`/`policy`. `ScaleUp` and `ScaleDown` are now thin embedding wrappers so call sites and the existing `NewScaleUp`/`NewScaleDown` constructors are unchanged. Tests use small `maxInstances()`/`minInstances()` accessors instead of field access. Net: scale.go went from 136 -> ~130 lines and has zero duplication. Also gofmt-fixed (the extra-space alignment you flagged on lines 37-38 / 98-99 is gone).

test/longhaul/cmd/longhaul/main.go
WentingWu666666: Fixed in ed498df — added `K8sClusterClient.Clientset() kubernetes.Interface`, deleted `main.buildRESTConfig` and `initK8sClient` entirely, and dropped the unused `k8s.io/client-go/{kubernetes,rest,tools/clientcmd}` imports from main.go. main.go now does the one `monitor.NewK8sClusterClient(...)` call and reuses `clusterClient.Clientset()` for the reporter. main.go shrank by ~40 lines.

test/longhaul/monitor/k8sclient.go
WentingWu666666: Fixed in ed498df — added private `(k *K8sClusterClient) getCR(ctx)` helper. All four call sites (`GetClusterHealth`, `GetInstancesPerNode`, `GetCurrentDocumentDBImageTag`, `UpgradeDocumentDB`) now use it. `types`/`shareddb` are still imported once at the top for the helper itself.

test/longhaul/monitor/health.go
WentingWu666666: Skipped intentionally. Agree the pattern recurs 5 times, but a cross-package utility for a 5-line idiom isn't a clear win — it would mean either a new `test/longhaul/internal` package (overhead) or sticking it in one of the existing packages (awkward import direction). Each call site is also a `Run()` method with package-specific logging, so the saved lines per site are closer to 2-3. Happy to revisit if more components get added in PR-3/PR-4.

test/longhaul/journal/policy.go
WentingWu666666: Keeping `DefaultOutagePolicy()` — it's referenced from 3 test files (`journal/journal_test.go:70-71`, `journal/policy_test.go:80`, `operations/scheduler_test.go:38`) where the tests need a "reasonable defaults" stand-in to construct fake operations / open disruption windows. Removing it would push three literal `OutagePolicy{...}` copies into the tests, which is worse. Open to renaming if you'd prefer something like `BaselineOutagePolicy()` to signal it's a test/baseline default rather than a production default.

test/longhaul/report/report.go
WentingWu666666: Fixed in ed498df — all ~15 occurrences switched to `fmt.Fprintf(&b, ...)`. Pure `b.WriteString("literal")` calls (no formatting needed) kept as-is since `Fprintf` would be unnecessary indirection there.

test/longhaul/report/checkpoint.go
WentingWu666666: Fixed in ed498df — `map[string]any`, `Printf("\n%s\n", "...")` collapsed to `Println("\n...")` / `Print("...\n\n")`, and `gofmt -w` cleaned the JSON map alignment. `go vet` and `gofmt -l` are both clean now.

test/longhaul/workload/metrics.go
WentingWu666666: Fixed in ed498df — comment now reads `// MetricsSnapshot is a point-in-time copy of all metric values.`

test/longhaul/cmd/longhaul/main.go
```diff
@@ -0,0 +1,260 @@
...
+		var timeoutCancel context.CancelFunc
+		ctx, timeoutCancel = context.WithTimeout(ctx, cfg.MaxDuration)
+		defer timeoutCancel()
+	}
+
+	// Initialize components.
+	j := journal.New()
+	metrics := workload.NewMetrics()
+
+	// Connect to MongoDB.
```
xgerman: DocumentDB

test/longhaul/cmd/longhaul/main.go
```diff
@@ -0,0 +1,260 @@
...
+	if cfg.MongoURI == "" {
```
xgerman: same

xgerman (changes requested):

test/longhaul/journal/journal.go
```diff
@@ -0,0 +1,223 @@
...
+		Message:   message,
+	}
+	j.mu.Lock()
+	j.events = append(j.events, e)
+	if len(j.events) > maxEvents+trimHeadroom {
+		trimmed := make([]Event, maxEvents)
+		copy(trimmed, j.events[len(j.events)-maxEvents:])
+		j.events = trimmed
+	}
+	j.mu.Unlock()
```
alaye-ms: Can we use the defer pattern used below here as well to handle any panics and errors

test/longhaul/monitor/k8sclient_test.go
```diff
@@ -0,0 +1,209 @@
...
+
+		var got previewv1.DocumentDB
+		Expect(k.crClient.Get(context.Background(), types.NamespacedName{Namespace: ns, Name: cluster}, &got)).To(Succeed())
+		Expect(got.Spec.DocumentDBVersion).To(Equal("0.110.0"))
+		Expect(got.Spec.SchemaVersion).To(Equal("auto"))
+	})
+
+	Describe("metrics", func() {
+		It("MetricsAvailable mirrors the metricsAvail flag", func() {
+			k := &K8sClusterClient{metricsAvail: true}
```
alaye-ms: It might just be an issue with my environment, but this portion gives an error for me: "cannot use true (untyped bool constant) as atomic.Bool value in struct literal - compiler[IncompatibleAssign]"

test/longhaul/monitor/leakdetect.go
```diff
@@ -0,0 +1,124 @@
...
+		minSamples:     minSamples,
+		samples:        make([]ResourceSample, 0, 256),
+	}
+}
+
+// AddSample records a resource usage observation.
+func (l *LeakDetector) AddSample(s ResourceSample) {
+	l.mu.Lock()
+	l.samples = append(l.samples, s)
+	l.mu.Unlock()
```
alaye-ms: Can we do the defer pattern here as well

alaye-ms (approved):

hossain-rayhan (approved):

xgerman (commented):
## Summary

Reviewed with a focus on workload-generation duplication against the existing E2E code. Net assessment: the **data-plane workload generator is not a real duplicate** (it is a distinct durability oracle), but a few **control-plane helpers are re-implemented rather than shared**, largely because `test/longhaul` is a separate module that can import `test/shared` but **not** `test/e2e`. Nothing blocking; the main ask is to hoist a couple of primitives into `test/shared` and to correct the PR description.

Great control-plane reuse already: `monitor/k8sclient.go` consistently delegates CR access to `test/shared/documentdb` (`Get`, `PatchInstances`, `PatchSpec`, `IsHealthy`) — no duplicated readiness strings or patch logic. 👍

## Why the workload generator is *not* a duplicate (for the record)

There are three "insert docs + count" styles in the repo now, but they are genuinely different concerns:

- `seed/datasets.go` + `shared/mongo.Seed` → one-shot deterministic fixtures (`InsertMany`).
- `performance/perf_insert_test.go` → one-shot throughput tripwire.
- `workload/writer.go` (this PR) → **continuous acked-write durability oracle**: per-tick `InsertOne`, `writeconcern.Majority()`, unique `(writer_id, seq)`, duplicate-key-as-ack reconciliation, resumable seq, checksums.

The writer correctly cannot be built on `Seed` (batched `InsertMany` returning a count cannot express per-doc durability semantics). So the divergence is justified.

## Findings

### 🟡 `[Required]` PR description over-claims data-plane reuse
The description lists `test/shared/mongo — … Seed, Count`, but the driver imports that package only for `NewFromURI`/`Ping`. The hot path bypasses both (`workload/writer.go` uses raw `InsertOne`; `workload/verifier.go` uses raw `Find`). Please reword to "connection only; the workload uses raw collection ops because the durability oracle needs per-doc semantics `Seed`/`Count` cannot express." This avoids future readers assuming the shared seeders are on the workload path.

### 🟠 `[Suggestion]` Pod readiness / restart-count scan is re-implemented
`monitor/k8sclient.go` (`GetClusterHealth` + `isPodReady`, the `cnpg.io/cluster` label scan, restart-count summation) re-derives pod-health inspection that the E2E suite already performs. The CR-readiness predicate *is* shared (`shareddb.IsHealthy`), which makes the un-shared pod scan stand out. Recommend hoisting the pod-health/restart scan into `test/shared` (e.g. `test/shared/k8s` or alongside `shared/documentdb`) so both suites consume one implementation.

### 🟠 `[Suggestion]` Connect+ping re-implemented without the existing retry
`cmd/longhaul/main.go` does `NewFromURI` + a single manual `Ping` (no retry), while E2E already has `pingWithRetry`/`NewFromDocumentDB` in `test/e2e/pkg/e2eutils/mongo/connect.go` that owns the "connection refused" reconnect window. Since the driver **deliberately induces disruption windows** (scale/upgrade ops), a no-retry connect is a latent reconnect gap, not just duplication. Recommend moving connect-with-retry into `test/shared/mongo` and consuming it from both the driver and e2e. (Most actionable item before PR-3 starts running ops.)

### 🟢 `[Nitpick]` Steady-state wait overlaps `WaitHealthy`
`monitor/health.go WaitForSteadyState` overlaps `shareddb.WaitHealthy` on the CR portion but legitimately adds pod-readiness, and it already composes `IsHealthy` underneath — so this is "extend the shared helper" rather than a true duplicate. Fine as-is; just flagging for the eventual consolidation.

## Recommendation
1. Move the genuinely shared primitives down into `test/shared`: pod-health/restart scan, and connect-with-retry (replacing both `main.go`'s manual ping and e2e's `connect.go`).
2. Keep `workload/` as-is — it is a distinct primitive, correctly not reusing the fixture seeders.
3. Fix the PR description's reuse claim, and optionally note in the README's "Relationship to `test/e2e/`" section *why* pod-scan/connect aren't shared yet (module split) and whether PR-3/4 will consolidate them.

`/needs-discussion` on where the shared k8s/mongo primitives should live (`test/shared/k8s`?) — otherwise looks solid and the unit tests are thorough.

xgerman (approved):

PR #406, opened 2026-06-16
feat: add io_uring micro-benchmark playground

xgerman:
## What

Adds `documentdb-playground/io-uring-benchmark/` — a reproducible cloud (AKS/EKS) playground that measures **PostgreSQL 18 `io_method = io_uring` vs `worker` vs `sync`** on DocumentDB, driving the [`documentdb/micro-benchmarks`](https://github.com/documentdb/micro-benchmarks) Locust read-query suite.

## Why / proven feasibility

A spike on a local kind cluster (kernel 6.12, `io_uring_disabled=0`) established the full enable path:

| Finding | Evidence |
|---|---|
| PG18 `18-minimal-trixie` ships io_uring | `initdb` bootstrap postgres attempts io_uring queue setup |
| The only blocker is container seccomp | `RuntimeDefault` → `FATAL: could not setup io_uring queue: Operation not permitted` |
| Relaxing seccomp turns it on | `Unconfined` → PG18.3 starts, `SHOW io_method` = `io_uring`, `pg_stat_io` reads flow |

CNPG natively exposes `spec.seccompProfile`, but the DocumentDB CR has no seccomp field, so the playground relaxes seccomp on the CNPG postgres pods via a **Kyverno mutate policy**.

## Contents

- **CR variants** `manifests/documentdb-{sync,worker,io_uring}.yaml` — differ only in `spec.postgres.parameters.io_method`.
- **seccomp** — Kyverno policies (`Unconfined` quick-start + hardened `Localhost`), the curated CNPG io_uring profile (from [CNPG #10446](https://github.com/cloudnative-pg/cloudnative-pg/pull/10446)), and a node-installer DaemonSet.
- **harness** — an in-cluster Locust runner + an `00…99` script pipeline + `analyze.py` (median + speedup-vs-`sync` table).
- **README** — proven feasibility, the seccomp deep-dive, and an I/O-bound methodology (dataset > RAM, one variable, repeats + median, `pg_stat_io` corroboration).

## How to run

```bash
cd documentdb-playground/io-uring-benchmark
./scripts/00-prereqs.sh && ./scripts/10-deploy-operator.sh && ./scripts/20-deploy-seccomp.sh
STORAGE_CLASS=<fast-ssd> ./scripts/30-deploy-documentdb.sh
DOCUMENT_COUNT=10000000 ./scripts/40-load-data.sh
./scripts/50-run-matrix.sh && ./scripts/60-collect.sh
```

## Notes

[...]

PR #407, opened 2026-06-19
feat: add IOUring opt-in feature gate for PostgreSQL async I/O

xgerman:
## What

Adds an **`IOUring` opt-in feature gate** so customers can enable PostgreSQL 18
`io_method = io_uring` asynchronous I/O on a DocumentDB cluster with a single
flag — **disabled by default** for security.

```yaml
apiVersion: db.microsoft.com/preview
kind: DocumentDB
spec:
  featureGates:
    IOUring: true
```

When the gate is enabled the operator does **both** things io_uring requires,
natively — no external Kyverno policy needed:

1. **Sets `io_method=io_uring`** as a protected PG parameter (mirrors the
   existing `ChangeStreams → wal_level` pattern in `ProtectedParameters`).
2. **Relaxes the postgres container seccomp profile** through CNPG's native
   `Cluster.Spec.SeccompProfile`. The runtime's `RuntimeDefault` profile strips
   `io_uring_setup/enter/register`, which otherwise causes
   `FATAL: could not setup io_uring queue: Operation not permitted` at startup.

## Why

This is the productized outcome of the io_uring benchmark spike (previously
PR #406, now closed). Instead of requiring a Kyverno policy to relax seccomp,
the operator handles it natively behind an opt-in gate.

## Seccomp configuration (operator-level)

Seccomp behavior is a **cluster-wide node decision**, so it is operator-level
config exposed as Helm values (not per-DocumentDB):

| Helm value | Default | Effect |
|---|---|---|
| `operator.ioUring.seccompMode` | `""` → `localhost` | `localhost` (hardened curated profile) or `unconfined` |
| `operator.ioUring.seccompProfile` | `""` → `profiles/documentdb-iouring.json` | Localhost profile path |

Wired through `DOCUMENTDB_IOURING_SECCOMP_MODE` /
`DOCUMENTDB_IOURING_SECCOMP_PROFILE` env vars.

> **Note:** The kernel must also have `io_uring_disabled=0` on the nodes.

## What's included

- **Feature gate**: `IOUring` constant, CEL allowlist entry, and default in
  `featureGateDefaults` (the documented 3-step add-a-gate procedure).
- **Operator behavior**: `io_method` protected param + `applyIOUringSeccomp`
  helper (single clean helper, no duplication).
- **Helm**: first-class `operator.ioUring.*` values + template env vars +
  helm-unittest coverage.
- **Playground**: `documentdb-playground/io-uring-feature/` — modeled on the
  upstream cnpg-playground seccomp approach (kind `extraMount` + a DaemonSet
  installer for real clusters). No Kyverno required.
- **Docs**: `docs/operator-public-documentation/io-uring.md` + mkdocs nav.
- **Tests**: gate default/true/false, `io_method` presence, and the seccomp
  nil/Localhost-default/Localhost-custom/Unconfined branches.

## Validation

- `gofmt` clean · `go build ./...` · `go vet ./...` · `go test ./internal/cnpg/... ./api/preview/...` pass
- `helm unittest .` — **85/85** pass
- `make manifests generate` produces no drift; CRDs in sync (config + helm)

## Notes

- No new CRD struct fields (map-based gate), so `zz_generated.deepcopy.go` is
  unchanged — correct.
- The seccomp profile JSON is intentionally vendored into the playground so it
  is self-contained/standalone.


docs/operator-public-documentation/io-uring.md
```diff
@@ -0,0 +1,174 @@
...
+
+The hands-on [io_uring feature playground](https://github.com/documentdb/documentdb-kubernetes-operator/tree/main/documentdb-playground/io-uring-feature) provides the curated profile plus a kind `extraMount` and a DaemonSet installer that distribute it to every node.
+
+!!! warning "Security trade-off"
+    Relaxing seccomp — even with the hardened Localhost profile that re-allows only the three io_uring syscalls — widens the kernel attack surface. io_uring has been a recurring kernel-exploit vector, so this is a trade-off you accept as the cluster operator. That is why the gate is opt-in and disabled by default.
+
+## Prerequisites
+
+- **PostgreSQL 18 image.** `io_method=io_uring` exists only in PostgreSQL 18 and later. Make sure the cluster runs a PG18 image.
+- **A node kernel with io_uring enabled.** The nodes must run a kernel that exposes io_uring with `io_uring_disabled=0`. Modern AKS, EKS, and GKE node images qualify.
```
hossain-rayhan: The config names are in mixed pattern. One is `io_uring_disabled=0` and our spec conig is `ioUring=1`. Not an issue but can we make them symmetric?

WentingWu666666 (approved):

hossain-rayhan (approved):

WentingWu666666 (approved):

PR #411, opened 2026-06-24
fix: harden injected sidecars for PSA restricted compliance

### LINKED ISSUE #387: [GA blocker] Runtime cluster pods are not Pod Security Admission "restricted" compliant
## Summary

The DocumentDB controller pods shipped by the Helm chart (operator, sidecar-injector, wal-replica) were hardened in #382 to meet Pod Security Admission (PSA) `restricted` requirements. **However**, the *runtime* cluster pods that the operator creates via CNPG are not compliant, because the two sidecar containers we inject via the CNPG-I plugin (`documentdb-gateway` and `otel-collector`) ship with weak / missing `securityContext` fields.

On a namespace labeled `pod-security.kubernetes.io/enforce=restricted`, the Kubernetes API server will reject every DocumentDB cluster pod and the cluster will never come up. The operator log will show successful reconciliation of the `Cluster` CR; the failure surfaces only in CNPG's pod-creation events.

## Reproduction

Install the chart into a namespace with `pod-security.kubernetes.io/enforce=restricted`, then `kubectl apply` a sample `DocumentDB` CR. Cluster pods will fail admission with a message similar to:

```
pods "documentdb-cluster-1" is forbidden: violates PodSecurity "restricted:v1.30":
allowPrivilegeEscalation != false (containers "documentdb-gateway", "otel-collector"),
unrestricted capabilities (containers "documentdb-gateway", "otel-collector" must set securityContext.capabilities.drop=["ALL"]),
seccompProfile (containers "otel-collector" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
```

## Audit (live cluster, `documentdb-cluster-1` pod)

| Container | Source | `runAsNonRoot` | `seccompProfile` | `allowPrivilegeEscalation` | `capabilities.drop` | `readOnlyRootFilesystem` | PSA `restricted`? |
|---|---|---|---|---|---|---|---|
| `postgres` | CNPG built-in | ✅ | ✅ | ✅ false | ✅ ALL | ✅ true | ✅ |
| `bootstrap-controller` (init) | CNPG built-in | ✅ | ✅ | ✅ false | ✅ ALL | ✅ true | ✅ |
| `documentdb-gateway` | DocumentDB sidecar-injector | ⚠️ inherited from pod | ⚠️ inherited from pod | ❌ missing | ❌ missing | ❌ missing | ❌ |
| `otel-collector` | DocumentDB sidecar-injector | ⚠️ inherited from pod | ❌ **not inherited** (no container-level value, sidecar has no SecurityContext at all) | ❌ missing | ❌ missing | ❌ missing | ❌ |

PSA `restricted` requires `allowPrivilegeEscalation: false` and `capabilities.drop: [ALL]` to be set **per container** — pod-level inheritance does not satisfy these checks.

## Root cause

`operator/cnpg-plugins/sidecar-injector/internal/lifecycle/lifecycle.go`:

- Line 176–179: the gateway sidecar's `SecurityContext` is set to only `RunAsUser` / `RunAsGroup`.
- Line 266+: the OTel collector sidecar is constructed with no `SecurityContext` field at all.

## Suggested fix (rough sketch)

Apply a hardened `SecurityContext` to both injected containers:

```go
SecurityContext: &corev1.SecurityContext{
    RunAsUser:                pointer.Int64(1000),
    RunAsGroup:               pointer.Int64(1000),
    RunAsNonRoot:             pointer.Bool(true),
    AllowPrivilegeEscalation: pointer.Bool(false),
    Capabilities:             &corev1.Capabilities{Drop: []corev1.Capability{"ALL"}},
    ReadOnlyRootFilesystem:   pointer.Bool(true),
    SeccompProfile:           &corev1.SeccompProfile{Type: corev1.SeccompProfileTypeRuntimeDefault},
},
```

Notes for the implementer:
- The OTel collector upstream image may need a writable scratch dir; if `readOnlyRootFilesystem: true` breaks it, mount an `emptyDir` at the writable path rather than dropping the flag.
- Consider exposing user-overridable values in the `DocumentDB` CR spec (`spec.gatewaySecurityContext`, `spec.otelCollectorSecurityContext`) so customers can adjust UID/GID for image variants without forking.
- Add a unit test in `lifecycle_test.go` asserting both injected containers carry the required fields, so this can't regress silently.

## Impact

🔴 **GA blocker** for any customer running on a Kubernetes platform that defaults namespaces to PSA `restricted`, including:

- AKS with Azure Policy "Kubernetes cluster pods should only use approved security profiles"
- GKE Autopilot
- OpenShift (which is even stricter, via SCC)
- Any cluster following CIS Benchmark recommendations

These customers can install the operator successfully (PR #382 covers the controller pods) but cannot create a working DocumentDB cluster.

## Out of scope

- Adding PSA enforcement labels to namespaces in the Helm chart (separate decision; see #381).
- Hardening controller pods themselves — done in #382.

## References

- Follow-up to #381 (umbrella GA chart audit)
- Related to #382 (controller pod hardening)
- [Pod Security Standards — restricted profile](https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted)

WentingWu666666:
## Summary

Fixes #387 — the CNPG-I sidecar-injector injects two containers into every DocumentDB cluster pod (`documentdb-gateway` and `otel-collector`), but neither carried a `SecurityContext` that satisfies the Kubernetes **Pod Security Admission (PSA) `restricted`** profile:

- `documentdb-gateway` set only `RunAsUser` / `RunAsGroup`.
- `otel-collector` had **no** `SecurityContext` at all.

PSA requires `allowPrivilegeEscalation: false`, `capabilities.drop: [ALL]`, `seccompProfile`, and `runAsNonRoot` to be set **per container** (pod-level inheritance does not satisfy the checks). On namespaces labeled `pod-security.kubernetes.io/enforce=restricted` (GKE Autopilot, OpenShift, AKS Azure Policy baseline, CIS-benchmark clusters) the API server rejects every cluster pod, so the DocumentDB cluster never comes up — even though the operator reports successful reconciliation.

## Changes

- Add a shared `hardenedSecurityContext()` helper and apply it to **both** injected sidecars. It sets, per container:
  - `runAsNonRoot: true`, `runAsUser/runAsGroup: 1000`
  - `privileged: false`, `allowPrivilegeEscalation: false`
  - `capabilities.drop: [ALL]`
  - `seccompProfile.type: RuntimeDefault`
- Add a unit test asserting the helper carries every PSA-`restricted` field, so this can't regress silently.

This mirrors how CloudNativePG hardens its own built-in containers (`pkg/specs` `GetSecurityContext`) and how the CNPG barman-cloud plugin hardens its injected sidecar.

### Design note: `readOnlyRootFilesystem`

Intentionally **omitted**. It is *not* required by the PSA `restricted` profile (confirmed by the admission error in #387, which flags only `allowPrivilegeEscalation`, `capabilities`, and `seccompProfile`), and the upstream gateway / OTel collector images haven't been verified to run on a read-only root filesystem. It can be added later with an `emptyDir` scratch mount once validated, without affecting admission compliance.

## Testing

- `go build ./...`, `go vet ./...`, `gofmt -l` — clean
- `go test ./...` (sidecar-injector module) — passing, including the new `TestHardenedSecurityContext_PSARestricted`

## Out of scope

- Adding PSA enforcement labels to chart namespaces (#381)
- Controller pod hardening (already done in #382)
- Exposing user-overridable `spec.gatewaySecurityContext` / `spec.otelCollectorSecurityContext` (can follow up if desired)


xgerman (approved):

xgerman:
Questions

 • 🟡 [Question]  enforce-version: latest  in fixtures.go means the  restricted  ruleset tracks the cluster's k8s version and can shift    across upgrades (forward-coverage vs. reproducibility). Intentional, or pin a version?
 • 🟢 The otel-collector runtime hardening (relying on the image's baked-in UID 10001 +  runAsNonRoot ) is only unit-tested — the deploy    e2e runs with monitoring off, so it's never admission-validated end-to-end. PR acknowledges this; a monitoring-on e2e would close the    loop.

I think WE Can live with both but posting for a wider discussion.

WentingWu666666:
Thanks for the review @xgerman! On the two points:

1. **`enforce-version: latest`** — intentional, for *forward-coverage*: the e2e suite validates against the newest `restricted` ruleset the API server knows, so a future k8s minor that tightens `restricted` surfaces against our injected sidecars immediately. Since this runs on a CI-pinned kind cluster, it's effectively deterministic per CI config and only shifts when we deliberately bump k8s — which is exactly when we'd want the signal. Happy to pin if the wider discussion prefers reproducibility over forward-coverage.

2. **otel-collector only unit-tested** — agreed, and it's a pre-existing gap (the suite runs with monitoring off, so otel is never injected/admission-validated e2e). Filed as a follow-up: #412. The existing `AssertInjectedSidecarsPSARestricted` helper already checks `otel-collector` when present, so that issue just needs a monitoring-on spec to light it up.

alaye-ms (approved):

PR #415, opened 2026-07-01
Add Mongoose + DocumentDB playground

hossain-rayhan:
## Add Mongoose + DocumentDB playground

Adds a self-contained playground under `documentdb-playground/mongoose/` that demonstrates and validates using [Mongoose](https://mongoosejs.com/) (the popular Node.js MongoDB ODM) against DocumentDB.

### What's included
- **Express + Mongoose REST API** (`app/server.js`, `app/models/book.js`, `app/db.js`) — health, CRUD, and a genre-aggregation endpoint.
- **Standalone Mongoose CRUD/compatibility suite** (`app/mongoose-crud-test.js`) — connect, index sync, insertOne/insertMany, filtered find + sort + limit, countDocuments, updateOne, findOneAndUpdate, aggregation, unique-index enforcement, deleteOne, and drop.
- **Helper scripts** (`scripts/`):
  - `run-app.sh` / `run-test.sh` — **primary path**: run the app/tests locally against an in-cluster DocumentDB via a temporary `kubectl port-forward`.
  - `deploy.sh` / `cleanup.sh` — **optional path**: build the image, load into kind, and run the app in-cluster.
  - `lib.sh` — shared helpers (resolve `MONGO_URI` from the DocumentDB status, manage the port-forward).
- **Manifests** — `documentdb.yaml` (sample DocumentDB instance sized for the demo) and `k8s/mongoose-app.yaml` (in-cluster Deployment + Service).
- **README** with architecture, connection requirements, a Mongoose↔DocumentDB compatibility matrix, and troubleshooting.

### How Mongoose connects
The DocumentDB gateway speaks the MongoDB wire protocol as a **standalone** server over **TLS**, so the app/test require `directConnection: true`, `tls: true`, `tlsAllowInvalidCertificates: true` (for the default self-signed install), and strip `replicaSet=rs0` from the connection string. The scripts handle this automatically.

### Validation (local kind, all default images

### DIFF
--- a/documentdb-playground/mongoose/README.md
+++ b/documentdb-playground/mongoose/README.md
@@ -1,0 +1,312 @@
+# Mongoose with DocumentDB
+
+This playground shows how to use [Mongoose](https://mongoosejs.com/), the most
+popular MongoDB ODM for Node.js, against DocumentDB. It includes:
+
+- a small **Express + Mongoose REST API** (`app/`), and
+- a standalone **Mongoose CRUD/compatibility test suite**
+  (`app/mongoose-crud-test.js`) that exercises connect, schema/index creation,
+  insert, query, update, aggregation, unique-index enforcement, and delete.
+
+**The primary path is to run the app and test suite locally on your machine**,
+connecting to a DocumentDB instance running in a cluster (local kind or remote
+AKS) through a `kubectl port-forward`. No image build or in-cluster deployment
+is required. Deploying the app *into* the cluster is an optional path documented
+later.
+
+Use it to validate that your application's Mongoose models behave correctly
+against the DocumentDB gateway.
+
+> **What is Mongoose?** Mongoose is a **Node.js library** (an ODM, Object Data
+> Modeling layer), not a CLI tool or a server. Your application imports it to
+> define schemas/models and talk to a MongoDB-compatible database. In this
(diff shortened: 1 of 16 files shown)


documentdb-playground/mongoose/README.md
```diff
@@ -0,0 +1,312 @@
...
+
+```bash
+# 1. Deploy a DocumentDB instance into your cluster (skip if you already have one).
+#    Works the same whether the cluster is local kind or remote AKS.
+kubectl create namespace documentdb-test
+kubectl apply -f documentdb.yaml
+kubectl wait --for=jsonpath='{.status.status}'="Cluster in healthy state" \
+    documentdb/documentdb-cluster -n documentdb-test --timeout=300s
+
+# 2. Run the Mongoose app LOCALLY (port-forwards DocumentDB for you).
```
alaye-ms, line 68: To me, this block implies that you run the app with `run-app.sh` then simultaneously run the tests with `run-test.sh` but I see that they run separately, and actually you can't run both at once due to the port-forwarding clobbering each other. Would we be able to make it clearer that these are separate operations?

alaye-ms (approved):
Ran the demo on a multi-region cluster successfully, both using the curl commands and the test suite

PR #419, opened 2026-07-08
test(longhaul): add data-protection verifier (ScheduledBackup + retention)

WentingWu666666:
## Long-haul data-protection verifier (ScheduledBackup + retention)

Adds the **Data protection** category from the long-haul design (`docs/designs/long-haul-test-design.md` → *Operations*) to the driver introduced in #405. Scope: exercise a **ScheduledBackup** on the canary cluster and verify the properties only a **multi-day** run can establish — that backups keep being produced and completing, and that expired backups are actually garbage-collected so the backup population stays bounded.

### What this PR adds

A new `test/longhaul/backup` package: a long-lived verifier that ensures a `ScheduledBackup` against the canary cluster and runs **concurrently** with the operation scheduler. Per the design, backup is deliberately *not* isolated from topology/chaos so that backup-vs-topology serialization bugs surface here.

Each verification cycle checks only what a multi-day canary can uniquely prove (unit/e2e tests can't):

| Check | Signal |
|---|---|
| **Scheduling liveness** | `status.lastScheduledTime` keeps advancing; a stalled scheduler (past `nextScheduledTime` + grace) warns |
| **Completion** | child `Backup` CRs keep reaching `completed`; terminal `failed` counted (a `skipped` backup is an intentional standby no-op, **not** a failure) |
| **Retention leak** | no completed backup outlives its retention window (`stoppedAt + spec.retentionDays*24h` + grace) -> a lingering backup is a **FAIL** |

A retention leak flips the run verdict to `FAIL` and is surfaced in the report's new **Data Protection** section.

### Design notes

- **Black-box oracle.** The leak deadline is derived from each backup's own declared `spec.retentionDays` (stamped by the ScheduledBackup at creation), **not** the operator-populated `status.expiredAt`. This verifies the *outcome* (expired backups disappear) without re-deriving the operator's expiration arithmetic — which is a pure function already covered by the operator's own unit tests (`backup_funcs_test.go`) and needs no accumulation. Keying off the per-backup value also keeps the check correct if a later run uses a different retention.
- **Bounded state.** The dedup maps are pruned to the live backup set each cycle (backup names are timestamp-unique and never reused), so the verifier's own memory stays bounded over a multi-day run.
- Because the minimum meaningful retention is 1 day, the leak check only fires on multi-day runs — exactly the accumulation window long-haul exists to cover.

### Changes

- `test/longhaul/backup/{metrics,verifier}.go` (+ unit tests, suite): the verifier, atomic metrics, and the retention-leak oracle.
- `monitor/k8sclient.go`: `EnsureScheduledBackup` / `GetScheduledBackup` / `ListScheduledChildBackups`. `EnsureScheduledBackup` **reconciles** an existing CR in place when a run's schedule/retention differ (history preserved, never recreated). Child backups are selected by the `scheduledbackup=<name>` label.
- `config.go`: `LONGHAUL_BACKUP_ENABLED` / `LONGHAUL_BACKUP_SCHEDULE` / `LONGHAUL_BACKUP_RETENTION_DAYS` + validation. (The verifier poll cadence is an internal package const, not an env knob.)
- `report.go`: Data Protection report section; `main.go`: bootstrap + goroutine wiring + FAIL gating.
- `deploy/setup.yaml`: corrected AKS prerequisite to k8s **>= 1.35** (operator requires ImageVolume, GA in 1.35).
- `.gitignore`: re-include `test/longhaul/backup/` (the generic Visual Studio `Backup*/` rule swallowed it on case-insensitive filesystems, same as the e2e backup dirs).

### Follow-up dependency

The driver ServiceAccount needs `create`/`get`/`list`/`update` on `scheduledbackups.documentdb.io` and `list` on `backups.documentdb.io`. Those verbs belong in `deploy/rbac.yaml` (added by the CI/CD PR); without them the verifier logs an error and the rest of the run continues. Documented in `test/longhaul/README.md`.

### Verification

\`\`\`
cd test/longhaul
go build ./...   # clean
go vet ./...     # clean
go test ./...    # all packages pass
\`\`\`


test/longhaul/config/config_test.go
```diff
@@ -160,6 +160,29 @@ var _ = Describe("Config", func() {
 			cfg.MaxInstances = 4
 			Expect(cfg.Validate()).To(MatchError(ContainSubstring("must not exceed 3")))
 		})
+
+		It("fails when backups enabled but schedule is empty", func() {
+			cfg := DefaultConfig()
+			cfg.ClusterName = "test"
+			cfg.BackupSchedule = ""
```
xgerman, line 219: [Required] Please also clear the new backup env vars (`LONGHAUL_BACKUP_ENABLED`, `LONGHAUL_BACKUP_SCHEDULE`, `LONGHAUL_BACKUP_RETENTION_DAYS`) in the `LoadFromEnv` `BeforeEach`. Right now tests are not hermetic: with `LONGHAUL_BACKUP_RETENTION_DAYS=abc` in the shell, multiple `LoadFromEnv` specs fail before asserting their target behavior.

test/longhaul/README.md
```diff
@@ -132,8 +132,47 @@ All configuration is via environment variables.
...
+concurrently with the operation scheduler (backup is deliberately **not**
+isolated from topology/chaos, per the design).
+
+The verifier only checks the properties a **multi-day** run can establish —
+things unit and e2e tests cannot:
+
+- **Scheduling liveness** — `status.lastScheduledTime` keeps advancing; a stalled
+  scheduler (past `status.nextScheduledTime` + grace) raises a warning.
+- **Completion** — child `Backup` CRs keep reaching `completed`; terminal
+  failures (`failed` / `skipped`) are counted.
```
xgerman: [Required] This sentence conflicts with the implementation in `backup/verifier.go`: `BackupPhaseSkipped` is explicitly treated as a non-failure (expected no-op on standby) and is not counted in `Failed`. Please align the README wording so operators dont interpret skipped backups as failures.

xgerman (changes requested):
## Summary
The backup verifier design and structure are good overall (clear separation into `backup` package, bounded dedup state, and focused metrics). I found two required fixes before merge, both around correctness/clarity rather than style.

## Critical Issues
1. `[Required]` README behavior mismatch: docs currently say `failed/skipped` are counted as failures, but implementation intentionally excludes `skipped` from failure counting.
2. `[Required]` `config` tests are no longer hermetic for the new backup env vars. Repro: `LONGHAUL_BACKUP_RETENTION_DAYS=abc go test ./test/longhaul/config` causes multiple unrelated `LoadFromEnv` specs to fail.

## Suggestions
- `[Suggestion]` Once the above is fixed, add a small `LoadFromEnv` coverage case for the new backup env knobs (`enabled`, `schedule`, `retentionDays`) to keep parser behavior explicit.

## Questions
None.

## Local kind validation
- Ran unit/module tests: `cd test/longhaul && go test ./...` ✅
- Attempted local kind runtime validation:
  - `go run ./cmd/longhaul` against existing `pgcosmos-emulator` service via port-forward failed during Mongo ping (`length of read message too large`), so it was not a valid longhaul target for this driver run.
  - Tried provisioning a temporary DocumentDB CR for longhaul testing, but this cluster currently has a cert-manager webhook certificate expiry (`x509: certificate has expired or is not yet valid`), so reconciliation could not complete.

Given the current kind environment state, runtime E2E verification of this PR could not be completed here, but the unit/module coverage for the new code is passing.

xgerman (commented):
## Summary

Reviewed against the actual operator CRDs and controllers (not just the diff). This is a high-quality PR — builds/vets clean and all package tests pass. The retention oracle in particular is very well designed.

**Verified correct:**
- **Label selector** — `ListScheduledChildBackups` filters on `scheduledbackup=<name>`, exactly what the operator stamps (`api/preview/scheduledbackup_funcs.go:25`). ✅
- **Retention-leak oracle alignment** — the verifier's deadline `stoppedAt + spec.retentionDays*24h + gcGrace` sits exactly `gcGrace` after the operator's own delete deadline: `Backup.CalculateExpirationTime` anchors completed backups on `StoppedAt + retentionDays*24h` (`api/preview/backup_funcs.go`). So the 10-min grace cleanly absorbs requeue latency with zero false positives. ✅
- Backup/ScheduledBackup status fields and CNPG phase constants all match the types. ✅
- Dedup + per-cycle prune keeps the maps bounded; the "timestamp-unique names are never reused" argument makes the prune safe against double-counting. ✅
- Bootstrap failure is non-fatal; `EnsureScheduledBackup` reconciles in place and never recreates, preserving history across restarts. ✅

Two findings below (one worth a design discussion), plus a couple of nits. Nice work.

test/longhaul/backup/metrics.go
```diff
@@ -0,0 +1,96 @@
...
+		Failed:         m.Failed.Load(),
+		RetentionLeaks: m.RetentionLeaks.Load(),
+		LastChildCount: m.LastChildCount.Load(),
+		LastScheduled:  lastScheduled,
+	}
+}
+
+// HasRetentionLeak returns true if any completed backup has outlived its
+// retention window. A true result flips the overall run verdict to FAIL.
+func (s MetricsSnapshot) HasRetentionLeak() bool {
```
xgerman, line 136: 🟠 **Major / Question — a wholly-broken backup subsystem can still report PASS.**

Only `RetentionLeaks > 0` flips the run verdict (confirmed in `cmd/longhaul/main.go` `buildSummary`, where only `backupSnap.HasRetentionLeak()` calls `appendReason`). `Failed` and scheduling-stall are observational.

So if *every* child backup terminally fails — `Completed=0`, `Failed` climbing — while `status.lastScheduledTime` keeps advancing and nothing lingers to leak, the long-haul run still reports **PASS**. For a suite whose whole purpose is multi-day data-protection regression detection, a completely broken backup path would go unflagged.

Is the non-fatal `Failed` intentional (tolerating transient chaos-induced failures)? If so, consider a *sustained*-failure gate that keeps that tolerance but still catches the real regression — e.g. "no `completed` backup within N schedule intervals" or a failure-ratio ceiling — rather than only the leak oracle.

test/longhaul/backup/verifier.go
```diff
@@ -0,0 +1,253 @@
...
+		}
+	}
+
+	// Liveness: if the next scheduled time is well in the past and no new
+	// backup has been scheduled since, surface a warning (once per overdue
+	// deadline). This is a Degraded signal, not a Fatal one — scheduling may
+	// be transiently delayed by apiserver load.
+	if next := sb.Status.NextScheduledTime; next != nil {
+		deadline := next.Time.Add(livenessGrace)
+		if now.After(deadline) && !v.lastScheduled.After(next.Time) && !v.stallWarnedFor.Equal(next.Time) {
```
xgerman, line 172: 🟡 **Minor — scheduling-stall path is untested.** The `stallWarnedFor` warn-once liveness logic here is the one branch without a unit test, while every other path (leak-once, gc-grace, own-retention-vs-config, prune, transient errors) is well covered. A small `checkScheduling` test driving `now` past `nextScheduledTime + livenessGrace` (and asserting warn-once) would close the gap.

alaye-ms (approved):

test/longhaul/backup/metrics.go
WentingWu666666: Addressed in 7c4405a. Added a completion-liveness oracle: the verifier now tracks the high-water mark of consecutive backups **scheduled with no intervening completion** (`MaxScheduledWithoutCompletion`), and the run FAILs once it reaches `completionStallThreshold` (3). A single completed backup resets the running gap, so transient chaos-induced failures stay under the ceiling — only a *sustained* dead completion path (every backup failing/hanging, ~18h at the default 6h schedule) trips it. This keeps the tolerance you wanted while catching the "wholly-broken subsystem" regression. Surfaced in the report + README; unit tests cover the stall, keep-pace, and recovery cases.

test/longhaul/backup/verifier.go
WentingWu666666: Addressed in 7c4405a. Added `Verifier.checkScheduling liveness` tests covering the `stallWarnedFor` warn-once path (driving `now` past `nextScheduledTime + livenessGrace`), plus within-grace and recovered-scheduling negatives.

.github/workflows/longhaul-smoke.yml
```diff
@@ -61,9 +74,9 @@ on:
   workflow_dispatch:
     inputs:
       max_duration:
-        description: "Bounded driver run length (Go duration, e.g. 3m)"
+        description: "Bounded driver run length (Go duration). Keep >= 6m so the 5m backup verify tick fires."
         required: false
-        default: "3m"
+        default: "6m"
```
xgerman, line 79: [Required] This window is too tight for a reliable end-to-end backup assertion. The driver only runs `checkOnce` on the first 5-minute ticker, while the first `*/1` schedule can occur nearly a minute after bootstrap and snapshot creation/completion can take additional time. If the backup is not completed by that single tick, the final report has `Backups Completed = 0` and this otherwise healthy smoke run fails intermittently. Please either make the smoke run materially longer (for example, at least 10–15 minutes) or make the verifier perform an immediate/short-cadence initial check before relying on the 5-minute interval.

test/longhaul/backup/verifier.go
```diff
@@ -0,0 +1,274 @@
...
+	if sb.Status.LastScheduledTime != nil {
+		last := sb.Status.LastScheduledTime.Time
+		if last.After(v.lastScheduled) {
+			v.lastScheduled = last
+			v.metrics.Scheduled.Add(1)
+			v.metrics.LastScheduledUnix.Store(last.Unix())
+			// A new backup was scheduled; widen the completion-stall gap. A
+			// completion in checkChildBackups resets it, and checkOnce publishes
+			// the high-water mark after both run.
+			v.scheduledSinceCompleted++
```
xgerman, line 161: [Required] `BackupPhaseSkipped` is documented as an expected no-op when the cluster is not primary, but this counter is incremented for every scheduled child before the verifier knows whether it will complete or be skipped. Three consecutive standby/skip schedules therefore trigger `HasCompletionStall()` and fail the run, even though skipped backups are intentionally non-failures. Please exclude skipped children from this stall gap (or reset/resolve the gap when the corresponding child reaches `skipped`) so a normal failover or standby interval does not produce a false FAIL.

xgerman (commented):
[Required] Documentation: `test/longhaul/README.md` still says the long-haul binary does not run in any PR-gated CI workflow. This PR adds `.github/workflows/longhaul-smoke.yml` on `pull_request` and runs the driver in kind. Please update the CI Safety section to distinguish the AKS long-running deployment from the new PR-gated smoke run.

xgerman (approved):

test/longhaul/config/config_test.go
WentingWu666666: Done — \`LoadFromEnv\`'s \`BeforeEach\` now clears \`LONGHAUL_BACKUP_ENABLED\`, \`LONGHAUL_BACKUP_SCHEDULE\`, \`LONGHAUL_BACKUP_RETENTION_DAYS\` (and \`LONGHAUL_BACKUP_VERIFY_INTERVAL\`) so the specs are hermetic regardless of the shell environment.

test/longhaul/README.md
WentingWu666666: Fixed — README wording now states skipped is an intentional no-op (non-primary/standby), is **not** counted as a failure, and resets the completion-stall gap like a completion, so operators won't read skipped as a failure.

test/longhaul/backup/metrics.go
WentingWu666666: Good catch, and agreed on the sustained-failure gate. \`HasCompletionStall()\` ("no completed backup within N schedule intervals", N=3) is now wired into the run verdict in \`cmd/longhaul/main.go\` \`buildSummary\` via \`appendReason\`, alongside the retention-leak oracle. This keeps tolerance of transient/chaos-induced \`Failed\` (still observational) but flips the run to FAIL when the backup path is genuinely broken (\`Completed=0\` while schedules keep advancing). Skipped children reset the gap so standby/failover doesn't false-trip it.

test/longhaul/backup/verifier.go
WentingWu666666: Added — \`Verifier.checkScheduling liveness\` specs now drive \`now\` past \`nextScheduledTime + grace\` and cover the warn-once behavior, re-arm on a new overdue interval, and recovery clearing \`stallWarnedFor\`.

.github/workflows/longhaul-smoke.yml
WentingWu666666: Addressed by making the verify cadence configurable rather than draining at shutdown. The smoke run sets \`LONGHAUL_BACKUP_VERIFY_INTERVAL: 30s\` (vs the 5m default) so the verifier's real periodic loop ticks ~12x within the 6m window and reliably observes a scheduled+completed backup. We chose the short-cadence loop over a one-off immediate check so the gate genuinely proves the periodic verifier works, not just that a shutdown snapshot happened.

test/longhaul/backup/verifier.go
WentingWu666666: Fixed — \`recordPhase\` now treats a \`skipped\` child like a completion for the stall gap: it resets \`scheduledSinceCompleted\` to 0 (and dedups + counts it via a \`Skipped\` metric). Consecutive standby/failover skips therefore no longer trip \`HasCompletionStall()\`. Covered by two new tests (standby skips every backup → no stall; a skip resets a climbing gap).

guanzhousongmicrosoft (approved):

guanzhousongmicrosoft (approved):

Issue #344, opened 2026-04-17
feat(otel): use dedicated least-privilege monitoring user for OTel Collector sidecar


WentingWu666666:
## Summary

The OTel Collector sidecar (added in #342) currently uses `PGUSER`/`PGPASSWORD` from the `<cluster>-app` secret (the CNPG application user) to connect to PostgreSQL. Following the principle of least privilege, the sidecar should use a dedicated monitoring user with only `pg_monitor` role access.

## Motivation

- **Least privilege**: The monitoring sidecar only needs to run health-check queries and read `pg_stat_*` views — it should not have application-level read/write access.
- **Audit clarity**: Monitoring queries would be attributable to a separate role in `pg_stat_activity`.
- **Future-proof**: `pg_monitor` already grants read access to all stats views, supporting richer metrics without needing superuser.
- **Blast radius**: If the OTel Collector is compromised, the attacker only gets read-only stats access.

## Proposed Implementation

1. Operator creates a dedicated `otel_monitor` PostgreSQL role with `pg_monitor` grant
2. Store credentials in a dedicated secret (e.g., `<cluster>-otel-monitor`)
3. OTel sidecar reads from the dedicated secret instead of `<cluster>-app`

## Context

Raised by @xgerman in PR #342 review: https://github.com/documentdb/documentdb-kubernetes-operator/pull/342#issuecomment-4269273931

PR #420, opened 2026-07-09
feat(otel): use dedicated monitoring identity for OTel Collector sidecar

udsmicrosoft:
## Summary

Addresses part of #344.

The OTel Collector sidecar no longer uses the CNPG application user's credentials. It now connects as a dedicated `otel_monitor` PostgreSQL identity.

PostgreSQL host authentication currently uses `trust`, so this PR intentionally does not generate or inject a password: it would not be checked and would require unnecessary cluster-wide Secret write permissions. This change improves query attribution and prepares the monitoring path for future authentication hardening, but it is not an isolation boundary for a compromised sidecar until `trust` is replaced.

## What changed

1. **Dedicated identity** — while monitoring is enabled, CNPG manages `otel_monitor` with `LOGIN`, `DisablePassword`, and no inherited monitoring roles. The current SQL receiver query is only `SELECT 1`, so broader `pg_monitor` access is not required.
2. **No monitoring Secret** — the sidecar sets `PGUSER=otel_monitor` directly, does not set `PGPASSWORD`, and the SQL datasource contains no password parameter.
3. **Correct disable lifecycle** — while monitoring is disabled, the desired CNPG role remains declared with `ensure: absent`, allowing CNPG to drop an existing or previously orphaned role. Reconciliation preserves unrelated managed roles and services.
4. **Read-only Secret RBAC** — the Helm and generated ClusterRoles retain only `get/list/watch` on Secrets.
5. **Plugin validation** — all OTel parameters now require the collector image and ConfigMap together, and mutation round-trips the image, ConfigMap, config hash, resource settings, and Prometheus port.

## Security scope

- The dedicated identity avoids normal monitoring queries running as the application user.
- `trust` authentication still allows any process that can reach PostgreSQL to claim another database role. Replacing it with SCRAM or certificate authentication remains necessary before claiming compromised-sidecar blast-radius isolation.
- Statistics privileges should be granted only when concrete collector queries require them.

## Testing

- `go test ./internal/otel ./internal/cnpg ./internal/controller`
- `go vet ./internal/otel ./internal/cnpg ./internal/controller`
- `go test ./internal/config ./internal/lifecycle ./internal/operator`
- `go vet ./internal/config ./internal/lifecycle ./internal/operator`
- `helm unittest -f 'tests/05_clusterrole_test.yaml' .`
- `make manifests`


operator/src/config/rbac/role.yaml
```diff
@@ -8,10 +8,20 @@ rules:
   - ""
   resources:
   - persistentvolumeclaims
+  verbs:
+  - get
+  - list
+  - watch
```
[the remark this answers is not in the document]
udsmicrosoft: Fixed in 130bcbc. The kustomize ClusterRole now grants `create` and `delete` on PVCs, matching `reconcilePVRecovery`, the kubebuilder marker, and the Helm ClusterRole.

operator/src/internal/cnpg/cnpg_sync.go
```diff
@@ -215,6 +215,29 @@ func SyncCnpgCluster(
...
+				Path:  PatchPathManaged,
+				Value: desired.Spec.Managed,
+			})
+		} else {
+			patchOps = append(patchOps, JSONPatch{
+				Op:    PatchOpAdd,
+				Path:  PatchPathManagedRoles,
+				Value: managedRoles(desired),
+			})
+		}
```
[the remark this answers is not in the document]
udsmicrosoft: Fixed in 130bcbc. Managed-role reconciliation now emits an RFC 6902 `remove` for `/spec/managed/roles` when monitoring is disabled instead of assigning a nil slice (`null`). It also treats nil and empty role lists equivalently; regression coverage asserts the remove operation and omitted value.

operator/cnpg-plugins/sidecar-injector/internal/config/config.go
```diff
@@ -89,6 +91,7 @@ func FromParameters(
 		DocumentDbCredentialSecret: credentialSecret,
 		OtelCollectorImage:         helper.Parameters[otelCollectorImageParameter],
 		OtelConfigMapName:          helper.Parameters[otelConfigMapNameParameter],
+		OtelMonitorSecret:          helper.Parameters[otelMonitorSecretParameter],
 		PrometheusPort:             prometheusPort,
```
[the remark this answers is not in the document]
udsmicrosoft: Fixed in 130bcbc. `FromParameters` now enforces `otelCollectorImage`, `otelConfigMapName`, and `otelMonitorSecret` as an all-or-nothing set whenever any OTel parameter (including port/hash) is configured. I also fixed `ValidateClusterChange` to append rather than overwrite those parameter errors, with config and operator-level regression tests.

operator/src/internal/controller/documentdb_controller.go
```diff
@@ -188,6 +195,10 @@ func (r *DocumentDBReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 			logger.Error(err, "Failed to clean up OTel ConfigMap")
 			return ctrl.Result{RequeueAfter: RequeueAfterShort}, nil
 		}
+		if err := r.deleteOtelMonitorSecret(ctx, documentdb.Name, req.Namespace); err != nil {
+			logger.Error(err, "Failed to clean up OTel monitoring secret")
+			return ctrl.Result{RequeueAfter: RequeueAfterShort}, nil
+		}
```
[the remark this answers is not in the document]
udsmicrosoft: Fixed in 130bcbc. Cleanup of both the OTel ConfigMap and monitoring Secret now happens only after `SyncCnpgCluster` succeeds and removes their references. A controller regression injects a CNPG patch failure and verifies both resources remain available for a safe retry.

operator/documentdb-helm-chart/templates/05_clusterrole.yaml
```diff
@@ -31,10 +31,11 @@ rules:
   resources: ["serviceexports", "multiclusterservices", "serviceimports", "internalserviceexports"]
   verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
 # Secrets: certificate_controller reads cert-manager-issued TLS secrets to
-# stamp into Cluster spec; no controller writes Secrets. Read-only.
+# stamp into Cluster spec; the DocumentDB controller also generates and manages
+# the per-cluster OTel monitoring credential secret (<cluster>-otel-monitor).
 - apiGroups: [""]
   resources: ["secrets"]
-  verbs: ["get", "list", "watch"]
+  verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
```
xgerman: [Suggestion] Since this adds required Secret write/delete permissions, please call out upgrade ordering in release notes/changelog (apply updated RBAC before or with controller rollout). In local kind testing, running the new controller against old RBAC repeatedly failed with `cannot create resource "secrets"` until ClusterRole was updated.

xgerman (approved):
## Summary
[Required] No blocking issues found in this PR.

I validated the key behavior changes locally (including kind):
- dedicated `<cluster>-otel-monitor` secret lifecycle (create/preserve/delete)
- `spec.managed.roles` reconciliation for `otel_monitor`
- sidecar credential wiring (`PGUSER`/`PGPASSWORD`) to `otelMonitorSecret`
- monitoring disable/enable cleanup and re-create path

Targeted tests run clean:
- `go test ./...` (sidecar-injector module)
- `go test ./internal/otel ./internal/cnpg ./internal/controller` (operator module)

I left one non-blocking inline suggestion about documenting RBAC upgrade ordering, based on local rollout behavior with stale ClusterRole.

alaye-ms (approved):

operator/src/internal/cnpg/cnpg_cluster.go
```diff
@@ -372,7 +376,40 @@ func applyIOUringSeccomp(spec *cnpgv1.ClusterSpec, documentdb *dbpreview.Documen
...
+	if documentdb == nil || documentdb.Spec.Monitoring == nil || !documentdb.Spec.Monitoring.Enabled {
+		return
+	}
+	if spec.Managed == nil {
+		spec.Managed = &cnpgv1.ManagedConfiguration{}
+	}
+	spec.Managed.Roles = append(spec.Managed.Roles, cnpgv1.RoleConfiguration{
+		Name:    otelcfg.MonitorRoleName,
+		Comment: "Least-privilege role for the OTel Collector monitoring sidecar",
+		Ensure:  cnpgv1.EnsurePresent,
```
xgerman: 🟢 **Observation (non-blocking) — the Postgres role outlives the secret on disable.** With `Ensure: present` (and no `absent` on teardown), removing the role from `spec.managed.roles` makes CNPG *stop managing* `otel_monitor` but does not DROP it. So after monitoring is disabled the `<cluster>-otel-monitor` secret is deleted while the DB role lingers. It's benign (pg_monitor-only, no application access) and re-enabling monitoring cleanly re-manages it with a fresh password — but worth a one-line comment here, or `Ensure: absent` on the removal path if a clean teardown is desired. Given `pg_hba` is currently `trust`, the lingering role can still connect without a password, which is another reason to note it (scope is still just monitoring views).

operator/documentdb-helm-chart/templates/05_clusterrole.yaml
xgerman: 🟢 **Nitpick (non-blocking) — broad secrets grant.** This widens the operator ClusterRole from read-only to `create/update/patch/delete` on **all** Secrets cluster-wide. It's needed (the operator already had cluster-wide get/list/watch, and it now owns `<cluster>-otel-monitor`), and a cluster-scoped operator can't easily namespace this — so acceptable. Just calling it out since a compromised operator can now write/delete any secret in the cluster; worth keeping in mind if the operator ever moves toward a namespaced-Role model. Kept correctly in sync with the generated `config/rbac/role.yaml`.

xgerman (approved):
## Summary

Reviewed against the actual CNPG managed-role/secret semantics and the collector's queries (both modules build, vet, and test green in an isolated worktree). This is a clean, security-positive change and the reasoning in the code comments is excellent. Verified the parts most likely to be subtly wrong:

**Verified correct:**
- **Password generation** — `crypto/rand`, 24 bytes (192-bit) via `base64.RawURLEncoding` → URL-safe, no chars that could break a PG connection string; generated once and preserved across reconciles (test: *preserves the existing password*). Secret carries an owner reference so it's GC'd with the CR. ✅
- **No perpetual re-patch** — `managedRolesPatch` uses `reflect.DeepEqual` and `applyOtelMonitorRole` sets the CNPG/CRD-defaulted fields (`ConnectionLimit=-1`, `Inherit=true`, `Ensure=present`) explicitly so the desired role matches the API-server-defaulted live form (test: *does not patch when managed roles are unchanged*). ✅
- **No managed-block clobber** — roles patch targets `/spec/managed/roles`, the replication flow owns `/spec/managed/services/additional`; disjoint paths, and `addAzureFleetManagedServices` now preserves the existing managed block instead of overwriting it (test: *preserves managed.services when patching only roles*). nil-vs-empty roles handled. ✅
- **Disable ordering is safe** — on disable, `applyOtelMonitorRole` is a no-op so the sync removes the role reference *before* `deleteOtelMonitorSecret` runs, so CNPG never dangles a missing `passwordSecret`; and the delete-after-sync guard retains OTel resources if sync fails (test: *retains OTel resources when cluster sync fails*). ✅
- **Sidecar gating** — injection now requires all three OTel params, and credentials are sourced from `<cluster>-otel-monitor` instead of `<cluster>-app`. ✅
- **`pg_monitor` is sufficient** — the only static query is `SELECT 1` against `dbname=postgres` (LOGIN + PUBLIC CONNECT), and the dynamic `pg_stat_*` metrics are exactly what `pg_monitor` grants — no functional regression. ✅

**Nice catch:** the `validation.go` change fixes a real latent bug — `ValidateChanges` previously **overwrote** the `FromParameters` errors (so the new required-OTel-param validation would have been silently dropped); appending is correct.

Approving. Two non-blocking observations inline.

operator/documentdb-helm-chart/templates/05_clusterrole.yaml
xgerman: 🔴 **Blast radius: cluster-wide secrets write.** This ClusterRole already grants `get;list;watch` on secrets in *every* namespace (needed for TLS reads). Adding `create;update;patch;delete` cluster-wide means a compromised operator (or a reconcile bug / hostile CR) can not only harvest but *forge/overwrite* any secret cluster-wide — overwriting a ServiceAccount token or a secret a privileged workload consumes is a classic k8s privilege-escalation path. Read-all + write-all on secrets is close to cluster-admin-equivalent.

`create` can't be constrained with `resourceNames`, so a broad ClusterRole verb is the wrong tool here. Prefer a **namespaced Role** (+ per-namespace RoleBinding) for the write verbs, keeping only `get;list;watch` cluster-wide for the cert reads. That caps write blast-radius to the namespaces where DocumentDB actually runs.

operator/src/config/rbac/role.yaml
```diff
@@ -8,10 +8,23 @@ rules:
...
+  - delete
+  - get
+  - list
+  - watch
+- apiGroups:
+  - ""
+  resources:
   - secrets
   verbs:
+  - create
```
xgerman: 🔴 Same concern as the Helm ClusterRole: cluster-wide `create/delete/patch/update` on secrets is a large grant. If the operator only ever writes `<cluster>-otel-monitor` in DocumentDB namespaces, scope these write verbs to a namespaced Role rather than the cluster-wide role. (Regenerate via the `+kubebuilder:rbac` marker in `documentdb_controller.go` once the marker is narrowed.)

operator/src/internal/cnpg/cnpg_sync.go
```diff
@@ -306,3 +314,47 @@ func getParam(params map[string]string, key string) string {
...
+	// Treat nil and empty role lists as equivalent.
+	if len(currentRoles) == 0 && len(desiredRoles) == 0 {
+		return nil
+	}
+	if reflect.DeepEqual(currentRoles, desiredRoles) {
+		return nil
+	}
+	if len(desiredRoles) == 0 {
+		return &JSONPatch{
+			Op:   PatchOpRemove,
```
xgerman: 🟠 **Lingering role.** Removing the entry from `spec.managed.roles` makes CNPG *stop reconciling* the role — it does **not** drop it. CNPG only drops a managed role when it's declared with `ensure: absent`. So after monitoring is disabled the `otel_monitor` LOGIN role stays in Postgres with its last-synced password, while `deleteOtelMonitorSecret` removes the secret — leaving an **orphaned credential** the operator can no longer rotate, and that silently diverges if monitoring is re-enabled.

Suggestion: instead of emitting a `remove` here, keep the role for one reconcile with `Ensure: EnsureAbsent` so CNPG actually `DROP`s it, then stop declaring it. (Or at minimum, document that the DB role intentionally lingers.)

operator/src/internal/cnpg/cnpg_cluster.go
```diff
@@ -372,7 +376,40 @@ func applyIOUringSeccomp(spec *cnpgv1.ClusterSpec, documentdb *dbpreview.Documen
...
+		Login:   true,
+		InRoles: []string{"pg_monitor"},
+		PasswordSecret: &cnpgv1.LocalObjectReference{
```
xgerman: 🟠 **This `PasswordSecret` (and its generated password) is unused today, which blunts the PR's security value.** `pg_hba` in this same file is `host all all 0.0.0.0/0 trust`, so the password is never checked — anyone who can reach 5432 can already connect as *any* role, including the superuser. Under `trust`, a dedicated `pg_monitor`-only role + random-password secret deliver no real access control (only audit-attribution).

Two things to decide:
- Until auth is tightened, you could **skip the secret entirely**: set `PGUSER=otel_monitor` as a literal env on the sidecar and declare the role with `DisablePassword`/no `passwordSecret`. That removes the whole generate-secret path *and* the cluster-wide secrets-write RBAC.
- The real fix is moving `pg_hba` to SCRAM (or cert auth via CNPG) — that's what makes this least-privilege role actually enforce anything.

operator/src/internal/cnpg/cnpg_cluster.go
```diff
@@ -372,7 +376,40 @@ func applyIOUringSeccomp(spec *cnpgv1.ClusterSpec, documentdb *dbpreview.Documen
...
+	}
+	if spec.Managed == nil {
+		spec.Managed = &cnpgv1.ManagedConfiguration{}
+	}
+	spec.Managed.Roles = append(spec.Managed.Roles, cnpgv1.RoleConfiguration{
+		Name:    otelcfg.MonitorRoleName,
+		Comment: "Least-privilege role for the OTel Collector monitoring sidecar",
+		Ensure:  cnpgv1.EnsurePresent,
+		Login:   true,
+		InRoles: []string{"pg_monitor"},
```
xgerman: 🟡 `pg_monitor` is a broad bundle (`pg_read_all_stats` + `pg_read_all_settings` + `pg_stat_scan_tables`). It's not an escalation risk, but it does expose **query text** via `pg_stat_activity`/`pg_stat_statements` (literals/PII embedded in SQL) and full server config. If the OTel `sqlquery` receiver only reads a known set of `pg_stat_*` views, consider a custom role granting just those views instead of the whole `pg_monitor` role — tighter least-privilege and no query-text leakage.

xgerman (commented):
Security review of the least-privilege monitoring role. The direction (dedicated `pg_monitor`-only role instead of the app user) is the right call and the role itself is **not** a Postgres privilege-escalation vector — `otel_monitor` has no `SUPERUSER`/`CREATEDB`/`CREATEROLE` (tests confirm), and `pg_monitor` is read-only observability. But three things blunt the security value and one increases the operator's blast radius. Details inline; summary:

1. 🔴 **Cluster-wide secrets write** — the operator already reads every secret cluster-wide; adding `create/update/patch/delete` cluster-wide is a strong k8s escalation primitive (forge SA tokens / overwrite secrets privileged workloads consume). Prefer a namespaced Role.
2. 🟠 **Lingering role on disable** — removing the entry from `spec.managed.roles` makes CNPG *stop managing* the role, not drop it. The `otel_monitor` LOGIN role persists with its last password while its secret is deleted → orphaned, un-rotatable credential. Use `ensure: absent` for one reconcile before removing.
3. 🟠 **`pg_hba` is `trust`** — the generated password is unused today; anyone who can reach 5432 connects as *any* role incl. superuser. So the whole secret + RBAC-widening buys nothing until auth is tightened. Consider skipping the secret entirely for now.
4. 🟡 **`pg_monitor` is broader than needed** — leaks query text (`pg_stat_activity`/`pg_stat_statements`) and config. Consider granting only the specific views the collector queries.

Not blocking the concept — flagging (2) and (3) as the items that determine whether this actually delivers least-privilege.

operator/src/internal/controller/documentdb_controller.go
```diff
@@ -1105,3 +1123,78 @@ func (r *DocumentDBReconciler) deleteOtelConfigMap(ctx context.Context, clusterN
...
+		secret.Type = corev1.SecretTypeBasicAuth
+		if secret.Data == nil {
+			secret.Data = map[string][]byte{}
+		}
+		// CNPG requires the secret username to match the managed role name exactly.
+		secret.Data[corev1.BasicAuthUsernameKey] = []byte(otelcfg.MonitorRoleName)
+		// Generate the password only once; preserve it on subsequent reconciles so
+		// the managed role's password and the sidecar's cached env stay in sync.
+		if len(secret.Data[corev1.BasicAuthPasswordKey]) == 0 {
+			password, genErr := generateRandomPassword(rand.Reader)
```
xgerman: 🟡 Tie-in with the `pg_hba trust` and RBAC comments: this password generation (and the owning secret) is the reason the operator needs cluster-wide secrets-write. Since the password is unused under `trust`, the lowest-risk option is to not generate/store it at all for now — source `PGUSER=otel_monitor` as a literal sidecar env and declare the managed role with `DisablePassword`. Reintroduce the secret when moving to SCRAM/cert auth. If you keep it, at least narrow the `+kubebuilder:rbac` marker (line 80) so the write verbs land in a namespaced Role.

udsmicrosoft: Addressed in d1ccd5a. The monitoring Secret, PGPASSWORD wiring, password datasource parameter, and Secret write RBAC were all removed while trust authentication remains.

operator/src/config/rbac/role.yaml
udsmicrosoft: Addressed in d1ccd5a. The DocumentDB Secret write marker was removed and make manifests now generates read-only Secret permissions while preserving the existing PVC lifecycle permissions.

operator/documentdb-helm-chart/templates/05_clusterrole.yaml
udsmicrosoft: Addressed in d1ccd5a. The monitoring Secret path was removed entirely, so the Helm ClusterRole is back to read-only get/list/watch access on Secrets.

operator/src/internal/cnpg/cnpg_sync.go
udsmicrosoft: Addressed in d1ccd5a. Monitoring-disabled desired state now always declares otel_monitor with ensure: absent, including cleanup of previously orphaned roles. Sync merges the operator-owned role by name and preserves unrelated managed roles and services.

operator/src/internal/cnpg/cnpg_cluster.go
udsmicrosoft: Addressed in d1ccd5a using the suggested interim design. The generated Secret and password were removed, the sidecar sets PGUSER=otel_monitor directly, the role uses DisablePassword, and the PR description now explicitly states that trust authentication is not a compromised-sidecar isolation boundary.

operator/src/internal/cnpg/cnpg_cluster.go
udsmicrosoft: Addressed in d1ccd5a. pg_monitor membership was removed; the current SELECT 1 query needs only LOGIN. Statistics privileges can be added narrowly when concrete collector queries require them.

operator/src/internal/cnpg/cnpg_cluster.go
udsmicrosoft: Addressed in d1ccd5a: when monitoring is disabled, the operator now declares `otel_monitor` with `ensure: absent`, so CNPG drops the role instead of leaving it unmanaged.

operator/documentdb-helm-chart/templates/05_clusterrole.yaml
udsmicrosoft: Addressed in d1ccd5a: the monitoring Secret path was removed, and the ClusterRole now keeps Secret access read-only with `get/list/watch`.

hossain-rayhan (approved):

PR #421, opened 2026-07-10
test(longhaul): add kill-operator-pod and kill-primary-pod chaos operations

WentingWu666666:
## Long-haul chaos operations + sequence-mode smoke gate

Adds the first two **Chaos / HA** operations from the long-haul design (`docs/designs/long-haul-test-design.md` → *Operations*) to the driver introduced in #405, reworks the outage-budget model they depend on, and adds a deterministic **sequence** operation mode plus a **PR smoke gate** that runs the real driver end-to-end against kind. Scope: fault-inject the control plane and the database primary on the canary cluster, and assert — over a multi-day run — that the data plane keeps serving and the cluster reconverges to its declared topology; and gate every PR by exercising each operation once against a real cluster.

### Chaos operations

Two new operations registered with the scheduler:

| Operation | Fault | Asserts |
|---|---|---|
| `kill-operator-pod` | Deletes the operator pod | The data plane keeps serving throughout (control-plane restart must not drop writes); the operator `Deployment` returns to `Available`. |
| `kill-primary-pod` | Deletes the CNPG primary pod | CNPG promotes a standby (automatic failover); writes resume and the cluster returns to full topology. Requires HA (`instancesPerNode>=2`). |

The continuous workload verifier independently catches any data loss caused by a failover.

### Outage policy: duration-based, writer-count-independent

The chaos ops needed a principled disruption budget, so `OutagePolicy` moved from a raw write-failure **count** to a wall-clock **write-outage duration**:

- `OutagePolicy.MaxWriteOutage time.Duration` replaces `AllowedWriteFailures int64`. The journal converts the observed failure count into an estimated outage using the workload's aggregate write rate (`WriteFailures / WritesPerSecond`), so budgets no longer scale with `LONGHAUL_NUM_WRITERS`. The rate is computed once at startup from `NumWriters` and the fixed writer cadence (`workload.AggregateWriteRate`) and stamped onto each disruption window.
- **Two orthogonal guardrails, both fail independently** (`ExceededPolicy` trips if either is exceeded):
  - `MaxWriteOutage` — client-visible **write availability** (data plane).
  - `MustRecoverWithin` — the managed cluster's return to **full declared topology** (all pods Ready, CR Ready). Because a failed op is only logged (not scored), this is the sole mechanism that fails the run when the cluster never reconverges — e.g. a failover where writes resume fast but a replacement standby never rejoins.
- **Shared budgets to prevent drift** (all in `journal/policy.go`):
  - `NoOutagePolicy` (~300ms cushion) for operations that keep the write path up throughout: `kill-operator-pod` and the scale ops (which only add/remove a standby). One fully-failed write tick maps to ~100ms of estimated outage regardless of writer count, so the cushion absorbs unrelated noise without tolerating a real outage.
  - `PrimaryHandoverPolicy` (30s) for `kill-primary-pod`: an ungraceful failover interrupts writes for exactly one primary handover.
  - `UpgradeOutagePolicy` (90s) for `upgrade-documentdb`: a cross-version rolling upgrade's graceful switchover coincides with the extension migration under live write load, so it gets its own, larger budget (calibrated from a real smoke run measuring ~33s, with headroom). The longer whole-topology restart is bounded by `MustRecoverWithin`, not the write-outage budget.

### Operation runner modes + Long-Haul Smoke Gate

The operation runner now supports three modes (`LONGHAUL_OPERATION_MODE`): `random` (production long-haul — weighted, one op every 10s, global cooldown), `sequence` (an explicit ordered `LONGHAUL_OPERATION_SEQUENCE`, each op once, exits when the sequence completes/fails), and `disabled`. All modes keep the continuous writer/verifier workload active.

Sequence mode powers a new PR smoke gate — `.github/workflows/longhaul-smoke.yml` — that builds the operator/sidecar from the PR, stands up a real DocumentDB on kind, and runs the actual driver through `scale-up → scale-down → upgrade-documentdb → kill-operator-pod → kill-primary-pod` once each, asserting every op reaches `PASSED`, the run reaches `COMPLETE`, and the backup verifier scheduled+completed a snapshot. The `upgrade-documentdb` step is a **real cross-version rolling upgrade** between two published releases (`0.110.0 → 0.113.0`, both hardcoded and pulled from the public registry — the smoke builds no database image), exercising the operator's rolling-update path with clean semvers the image-rollback webhook accepts.

### Design notes

- **Operator discovery.** The operator `Deployment` name is a chart-pinned literal (`documentdb-operator`, singleton install), so it is a safe constant; the namespace is install-overridable, exposed as `LONGHAUL_OPERATOR_NAMESPACE` (default `documentdb-operator`). There is no back-reference from the DocumentDB CR to the operator, so cluster-derived discovery isn't viable. `kill-operator-pod` builds its pod selector from the Deployment's own `spec.selector.matchLabels` rather than a hardcoded label (the operator label is release-name dependent), and refuses to proceed on an empty selector so it can never list/delete every pod in the namespace.
- **Primary discovery.** `kill-primary-pod` reads the CNPG `Cluster.Status.CurrentPrimary` (CNPG cluster name == DocumentDB CR name) to target the primary pod, then waits for the promoted primary to change and steady state to return.
- **HA guard.** `kill-primary-pod` and `upgrade-documentdb` skip (do not fail) when `instancesPerNode<2` — the disruption would be guaranteed downtime with no standby. Skips don't consume the scheduler cooldown and are re-evaluated on the next tick.

### Changes

- `operations/kill_operator.go`, `operations/kill_primary.go` (+ unit tests): the two chaos ops.
- `operations/{registry,runner,sequence}.go` (+ tests) and `operations/scheduler.go`: the operation runner with `random` / `sequence` / `disabled` modes.
- `journal/policy.go`, `journal/journal.go`: duration-based `OutagePolicy`, `WritesPerSecond` per window + `EstimatedWriteOutage`, `SetWriteRate`, and the shared `NoOutagePolicy` / `PrimaryHandoverPolicy` / `UpgradeOutagePolicy` helpers.
- `workload/writer.go`: `AggregateWriteRate(numWriters)`.
- `monitor/{health,k8sclient}.go`: `GetPrimaryInstance` / `DeletePod` on `ClusterClient`; CNPG scheme registration.
- `operations/scale.go`, `operations/upgrade.go`: adopt the shared policies (scale ops were previously over-lenient count budgets; upgrade gets its own 90s budget).
- `config/config.go`: `LONGHAUL_OPERATOR_NAMESPACE`, `LONGHAUL_OPERATION_MODE`, `LONGHAUL_OPERATION_SEQUENCE` + validation.
- `report/{report,checkpoint}.go`: disruption-window table gains an **Est. Write Outage** column; report exposes `operation-status` / `operation-results` / `operation-aggregates`.
- `cmd/longhaul/main.go`: supply the write rate to the journal; wire up the operation mode; register both ops.
- `deploy/rbac.yaml`: chaos-op RBAC (see below).
- `.github/workflows/longhaul-smoke.yml`: the sequence-mode PR smoke gate.
- `docs/designs/long-haul-test-design.md` / `test/longhaul/README.md`: README slimmed to a runner's guide; design doc gains the outage-budget, backup-verification, HA-precondition, and e2e-relationship rationale.
- `go.mod`: promote cloudnative-pg to a direct dependency.

### RBAC

Included in `deploy/rbac.yaml`: the driver ServiceAccount gets `get`/`list` on `clusters.postgresql.cnpg.io` + `delete` on `pods` (kill-primary), and, via a namespace-scoped Role/RoleBinding on `LONGHAUL_OPERATOR_NAMESPACE`, `get` on `deployments` + `get`/`list`/`delete` on `pods` in the operator namespace (kill-operator). Documented in `test/longhaul/README.md`.

### Verification

```
cd test/longhaul
go build ./...   # clean
go vet ./...     # clean
go test ./...    # all packages pass
```

The Long-Haul Smoke Gate runs the full sequence (including the real `0.110.0 → 0.113.0` upgrade) green on kind.

### Follow-up tasks
- Improve write-outage measurement to handle concurrent writers correctly. Do not end an outage after a single writer succeeds; require a stronger recovery signal, such as consecutive successful writes from all active writers or a continuous healthy-write interval. Add unit tests covering overlapping failures, partial recovery, and recovery after all writers resume.


test/longhaul/journal/journal.go
```diff
@@ -146,14 +148,57 @@ func (j *Journal) CloseDisruptionWindow() {
...
+		// Writers are concurrent and each captures attemptStart before its
+		// (possibly blocking) driver call, but outcomes arrive here in
+		// completion order. A success whose attempt began at or before the
+		// current outage start carries no information that the outage has
+		// ended — it was already in flight when the outage opened — so it must
+		// neither shrink the measured outage nor clear the marker. Only a
+		// success that started strictly after the outage began proves the write
+		// path recovered.
+		if attemptStart.After(w.WriteOutageStart) {
+			if gap := attemptStart.Sub(w.WriteOutageStart); gap > w.MaxWriteOutageObserved {
```
xgerman: **Major:** This measures recovery at the successful attempt’s start, but recovery is only proven when that call completes. For example, a write can start 100 ms after the first failure, block in server selection for 25 seconds, and then succeed; this records a 100 ms outage rather than 25.1 seconds, potentially allowing an over-budget failover to pass. Please capture the attempt completion time as well: retain `attemptStart` for rejecting stale in-flight successes, but calculate the outage using `attemptEnd.Sub(w.WriteOutageStart)`. A test should cover a success that starts early and completes much later.

test/longhaul/operations/scheduler.go
```diff
@@ -76,48 +106,58 @@ func (s *Scheduler) Run(ctx context.Context) {
...
-	s.executeOp(ctx, op)
+	err := s.executeOp(ctx, op)
 
 	s.mu.Lock()
 	s.inProgress = false
 	s.lastOpTime = time.Now()
 	s.opsExecuted++
 	s.mu.Unlock()
+
+	s.recordExecution(op.Name(), err)
```
xgerman, line 178: **Major:** If the configured random-mode duration expires while `Execute` is waiting for recovery, the parent context cancellation propagates as an error here and is recorded as an operation failure. A normal bounded run can therefore produce `FAIL` solely because its deadline landed during an operation. Please either stop scheduling and let the in-flight operation drain under its own bounded recovery context, or classify normal parent-context cancellation separately and recheck steady state before reporting. Add a lifecycle test that cancels the context inside `Execute`.

test/longhaul/report/checkpoint.go
```diff
@@ -99,13 +121,19 @@ func (r *CheckpointReporter) emit(ctx context.Context, final bool) {
...
+		return summary
 	}
 
 	data := map[string]string{
-		"latest-report": markdown,
-		"last-updated":  time.Now().UTC().Format(time.RFC3339),
-		"result":        resultStr,
+		"latest-report":     markdown,
+		"last-updated":      time.Now().UTC().Format(time.RFC3339),
+		"result":            resultStr,
```
xgerman, line 193: **Major:** Replacing the complete ConfigMap data allows the automatic Deployment restart to erase an operation failure. The failed process writes `result=FAIL` and exits; the restarted process has fresh in-memory state and its next checkpoint overwrites that with `RUNNING`. A delayed monitor can then miss the incident entirely. Please persist durable incident state across restarts—for example, a latched failure or monotonic incident count plus the last failure reason/time—and have the monitor alert from that state. A reporter test should start from an existing failed ConfigMap and verify that a fresh intermediate checkpoint preserves the incident.

xgerman (changes requested):
Requesting changes for three correctness issues that can hide real failures in the long-haul test: outage duration is under-counted, normal random-mode expiry can be reported as an operation failure, and a pod restart can erase the durable failure verdict.

xgerman (approved):
Re-reviewed the latest head. The three previously blocking issues are fixed with regression coverage: write outages use successful completion time, expected random-mode cancellation is not recorded as failure, and durable incident state survives driver restarts and is consumed by the monitor. The long-haul race tests, vet, smoke workflow, and required CI checks pass.

guanzhousongmicrosoft (approved):

PR #422, opened 2026-07-14
test(longhaul): add retention pruner to bound workload collection disk usage

WentingWu666666:
## What

Adds a **count-based retention pruner** to the long-haul test driver so the endurance write load no longer grows the DocumentDB collection (and its PVC) without bound.

## Why

An infinite-duration write test on a fixed disk always eventually fills it. The westus2 long-haul cluster exhausted its 10Gi volume after ~24.9M writes, which put Postgres into a `CrashLoopBackOff` ("Not enough disk space") and tripped the scheduled Long Haul Monitor. This bounds steady-state disk use so it can't recur.

## How

- **Pruner** (`test/longhaul/workload/pruner.go`): every 5m, per writer, deletes documents older than the most-recent `LONGHAUL_RETAIN_PER_WRITER` (default **2,000,000**; `0` disables). Uses an index-backed range `DeleteMany` on the existing unique `(writer_id, seq)` index.
- **No false data-loss (key invariant):** the pruner deletes *strictly below the verifier's confirmed floor*, so it never removes a document the durability oracle hasn't already accounted for. The floor is published via atomics so the pruner reads it race-free.
- **Restart-safe verifier:** on startup the verifier seeds each writer's expected seq from the collection's current **minimum** seq, so a post-restart scan doesn't misread the pruned prefix as one enormous gap (which would be a false FAIL).
- Wiring: `main.go` starts the pruner only when `RetainPerWriter > 0`; each prune (and any freeze-on-failure) is recorded in the run journal; default set in the deployment ConfigMap.

## Testing

- `go build ./...`, `go vet ./...`, `gofmt -l` clean
- `go test ./...` (all packages) + `go test -race ./workload/...` pass
- New unit tests: pruner logic incl. the "never delete at/above the floor" safety invariant, verifier `ConfirmedFloor`, config load/validate for the new env var

## Notes

The already-running westus2 cluster was recovered separately by expanding its PVC to 64Gi; this PR prevents the refill from happening again. Opening as **draft** for review.

### DIFF
--- a/test/longhaul/workload/pruner.go
+++ b/test/longhaul/workload/pruner.go
@@ -1,0 +1,137 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package workload
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/documentdb/documentdb-operator/test/longhaul/journal"
+
+	"go.mongodb.org/mongo-driver/v2/bson"
+	"go.mongodb.org/mongo-driver/v2/mongo"
+)
+
+const (
+	// pruneInterval is how often the pruner trims old documents. A long-haul
+	// run writes ~10 docs/sec/writer, so a few thousand rows accumulate per
+	// writer between cycles — a small, index-backed DeleteMany each time.
+	pruneInterval = 5 * time.Minute
+)
+
+// floorProvider reports the highest fully-verified seq per writer. *Verifier
+// satisfies this; the pruner only ever deletes strictly below the floor, so it
+// can never remove a document the verifier has not already accounted for.
+type floorProvider interface {
+	ConfirmedFloor(writerID string) int64
+}
+
+// pruneBackend abstracts the delete so Pruner can be unit-tested without a
+// live collection. Production uses docdbPruneBackend.
+type pruneBackend interface {
+	// deleteThrough removes all documents for writerID with seq <= throughSeq
+	// and returns the number deleted.
+	deleteThrough(ctx context.Context, writerID string, throughSeq int64) (int64, error)
+}
+
+// docdbPruneBackend adapts *mongo.Collection to pruneBackend. The delete filter
+// rides the existing unique (writer_id, seq) index, so it is a bounded range
+// delete rather than a collection scan.
+type docdbPruneBackend struct {
+	coll *mongo.Collection
+}
+
+func (m docdbPruneBackend) deleteThrough(ctx context.Context, writerID string, throughSeq int64) (int64, error) {
+	res, err := m.coll.DeleteMany(ctx, bson.D{
+		{Key: "writer_id", Value: writerID},
+		{Key: "seq", Value: bson.D{{Key: "$lte", Value: throughSeq}}},
+	})
+	if err != nil {
+		return 0, err
+	}
+	return res.DeletedCount, nil
+}
+
+// Pruner bounds the workload collection by keeping only the most recent
+// retainPerWriter documents per writer. It deletes strictly below the
+// verifier's confirmed floor, so pruning never affects the durability verdict:
+// every removed document was already scanned, and the verifier's startup
+// DB-min seed keeps a post-restart scan from misreading the pruned prefix as a
+// gap.
+type Pruner struct {
+	writers         []*Writer
+	floor           floorProvider
+	backend         pruneBackend
+	retainPerWriter int64
+	metrics         *Metrics
+	journal         *journal.Journal
+}
+
+// NewPruner constructs a Pruner. retainPerWriter must be > 0; callers gate on
+// that (0 disables pruning entirely) before constructing.
+func NewPruner(coll *mongo.Collection, writers []*Writer, floor floorProvider, retainPerWriter int64, metrics *Metrics, j *journal.Journal) *Pruner {
+	return &Pruner{
+		writers:         writers,
+		floor:           floor,
+		backend:         docdbPruneBackend{coll: coll},
+		retainPerWriter: retainPerWriter,
+		metrics:         metrics,
+		journal:         j,
+	}
+}
+
+// Run starts the prune loop. It blocks until the context is cancelled.
+func (p *Pruner) Run(ctx context.Context) {
+	p.journal.Info("pruner", fmt.Sprintf("pruner started (retain %d docs/writer)", p.retainPerWriter))
+	defer p.journal.Info("pruner", "pruner stopped")
+
+	ticker := time.NewTicker(pruneInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			p.pruneAll(ctx)
+		}
+	}
+}
+
+func (p *Pruner) pruneAll(ctx context.Context) {
+	for _, w := range p.writers {
+		p.pruneWriter(ctx, w.id)
+	}
+}
+
+// pruneWriter deletes the writer's documents older than the retention window.
+// The delete boundary is (confirmed floor - retainPerWriter): everything at or
+// below it is both fully verified and outside the retained tail, so removing it
+// is safe and bounds the writer's footprint to ~retainPerWriter documents.
+func (p *Pruner) pruneWriter(ctx context.Context, writerID string) {
+	floor := p.floor.ConfirmedFloor(writerID)
+	throughSeq := floor - p.retainPerWriter
+	if throughSeq < 1 {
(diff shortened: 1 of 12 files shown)


test/longhaul/workload/verifier_test.go
```diff
@@ -45,6 +47,24 @@ var _ = Describe("Verifier", func() {
 		// what the verifier actually does. Document the boundary.
 		Skip("verifyAll requires a *mongo.Database; covered by long-haul integration runs")
 	})
+
+	Describe("ConfirmedFloor", func() {
+		It("returns 0 for a writer registered but not yet verified", func() {
+			v := &Verifier{floor: map[string]*atomic.Int64{"w000": {}}}
```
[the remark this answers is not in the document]
WentingWu666666: Not a compile error — this is valid Go. Inside a composite literal whose element type is a pointer (`*atomic.Int64`), the `&T` is elided, so `{}` constructs `&atomic.Int64{}` ([Go spec: Composite literals](https://go.dev/ref/spec#Composite_literals)). `go vet ./workload/...` passes and the test binary compiles/runs green. Leaving as-is.

test/longhaul/workload/verifier_test.go
```diff
@@ -45,6 +47,24 @@ var _ = Describe("Verifier", func() {
...
+			Expect(v.ConfirmedFloor("w000")).To(BeZero())
+		})
+
+		It("reflects the published floor and is isolated per writer", func() {
+			v := &Verifier{floor: map[string]*atomic.Int64{
+				"w000": {}, "w001": {},
+			}}
```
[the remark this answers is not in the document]
WentingWu666666: Same as above — `{}` in a `map[string]*atomic.Int64` literal elides `&atomic.Int64{}` and compiles cleanly (per the Go spec on composite literals). Verified with `go vet` + `go test ./workload/...`. No change needed.

test/longhaul/workload/pruner.go
```diff
@@ -0,0 +1,158 @@
...
+	if throughSeq < 1 {
+		// Not enough verified history yet to prune anything.
+		return
+	}
+
+	deleted, err := p.backend.deleteThrough(ctx, writerID, throughSeq)
+	if err != nil {
+		p.journal.Warn("pruner", fmt.Sprintf("prune failed for writer %s: %v", writerID, err))
+		return
+	}
```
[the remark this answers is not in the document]
WentingWu666666: Good catch on the discrepancy — but the resolution is to fix the description, not the code. `DocsPruned` was intentionally removed during review (report row → snapshot field → counter) because it had no production reader once the report row was dropped; the pruner already journals every prune and any freeze-on-failure, which is the real observable. Updated the PR description to drop the stale `DocsPruned` claim.

test/longhaul/workload/verifier.go
```diff
@@ -205,8 +236,27 @@ func (v *Verifier) verifyWriter(ctx context.Context, w *Writer) {
...
+	// One-time startup seed: retention pruning deletes already-verified
+	// documents below the confirmed floor, so after a pod restart (which
+	// resets the in-memory nextSeq to 1) a naive scan from seq 1 would see the
+	// pruned prefix as one enormous gap — a false data-loss verdict. Advance
+	// expectedSeq to the collection's current minimum seq so the verifier only
+	// audits documents that still exist. Real gaps above the surviving floor
+	// are still detected. Done once per writer; steady-state cycles never
+	// re-read below nextSeq.
+	if !v.seeded[writerID] {
+		v.seeded[writerID] = true
```
xgerman: 🟠 **Major — the one-time seed is marked done *before* it succeeds, so a transient error produces a false data-loss FAIL on a pruned+restarted collection.**

`v.seeded[writerID] = true` is set unconditionally, before the `minSeq` query. If that query errors transiently at startup (network blip, majority-read timeout), the warning is logged but `seeded` stays `true`, so the seed is **never retried**. The very next scan then runs from `expectedSeq = 1` against a collection whose pruned prefix no longer exists → the surviving `minSeq` document registers as one enormous internal gap → `HasDataLoss()` flips to FAIL for the rest of the run, and (ironically) that freezes the pruner.

This is exactly the restart-after-pruning path this PR introduces — before pruning, a scan-from-1 after restart was always correct, so the seed *has* to be robust to be safe.

**Fix:** only mark seeded on success, so a transient error retries next cycle (`minSeq` returning `(0, nil)` for an empty collection counts as success):

```go
if !v.seeded[writerID] {
    if minSeq, err := v.minSeq(ctx, writerID); err != nil {
        v.journal.Warn("verifier", fmt.Sprintf("min-seq seed failed for writer %s (will retry): %v", writerID, err))
    } else {
        v.seeded[writerID] = true
        if minSeq > expectedSeq {
            expectedSeq = minSeq
        }
    }
}
```

test/longhaul/workload/verifier_test.go
```diff
@@ -45,6 +47,24 @@ var _ = Describe("Verifier", func() {
 		// what the verifier actually does. Document the boundary.
 		Skip("verifyAll requires a *mongo.Database; covered by long-haul integration runs")
 	})
+
+	Describe("ConfirmedFloor", func() {
```
xgerman, line 52: 🟡 **Minor — the startup min-seq seed has no test**, which is precisely where the Major bug above lives. The suite covers `ConfirmedFloor`/`publishFloor` and the auditor, but not the restart seed. A test that pre-populates the collection starting at seq N, constructs a fresh verifier (empty `nextSeq`), runs one cycle, and asserts **no gap** is counted would exercise the happy path — plus an error-injection variant asserting the seed is retried (not permanently marked done) would lock in the fix.

xgerman (changes requested):
## Summary

Reviewed against the actual verifier/pruner semantics (built, vetted, all package tests pass). The core design is genuinely good — floor-based deletion + freeze-on-data-loss is a safe, well-reasoned way to bound disk without ever touching the durability verdict:

**Verified correct:**
- Pruner only ever deletes `seq <= confirmedFloor - retainPerWriter`, i.e. strictly below what the verifier has already accounted for; the delete rides the existing `(writer_id, seq)` unique index (bounded range delete, not a scan). ✅
- Any gap/checksum failure freezes pruning permanently to preserve evidence; freeze is checked before pruning and `HasDataLoss()` is monotonic, and the verifier increments the loss counters *before* it publishes the advanced floor — so the pruner can never observe a floor past a gap without also seeing the freeze. ✅
- Concurrency is sound: `floor` is a read-only `map[string]*atomic.Int64` built once at construction; only the atomics are Load/Store-d across goroutines. ✅
- The streaming `auditor` refactor is O(1) per cycle and now surfaces `cursor.Err()` — a latent truncated-read → false-tail-loss bug that the old slice path silently swallowed. ✅
- Config gating (`RetainPerWriter==0` disables), default (2M docs/writer), and validation all look right. Pruner tests are thorough — boundaries, the never-delete-at-or-above-floor safety invariant, and all three freeze cases.

One blocking bug in the new startup seed, plus a test gap and a nit.

test/longhaul/workload/verifier.go
WentingWu666666: Fixed in 11a91a9. The seed decision is now a pure `seedExpectedSeq` helper returning `(newExpectedSeq, ok, err)`; the call site marks the writer seeded **only** when `err == nil`, so a transient `minSeq` error logs a "will retry" warning and re-attempts next cycle instead of being skipped forever. An empty collection (`minSeq` returns `(0, nil)`) still counts as success. Thanks for catching this — it's exactly the restart-after-pruning path this feature introduces.

test/longhaul/workload/verifier_test.go
WentingWu666666: Added in 11a91a9. Since the suite has no mongo seam (the `verifyAll` DB path is `Skip`ped, "covered by long-haul integration runs"), I extracted the seed decision into the pure `seedExpectedSeq` helper and unit-tested it directly under `Describe("startup min-seq seed")`:

- **restart happy path** — fresh verifier (`expectedSeq=1`) against a pruned collection whose surviving minimum is 5000 → `expectedSeq` advances to 5000, so the pruned prefix is not read as a false gap;
- **never rewinds** — a minimum at/below the resume point leaves `expectedSeq` unchanged;
- **empty collection** — `(0, nil)` counts as a successful seed;
- **regression guard** — a transient error does **not** mark the writer seeded (`expectedSeq` unmoved) and the next cycle retries and succeeds.

alaye-ms (approved):

hossain-rayhan (approved):

xgerman (approved):

PR #426, opened 2026-07-16
test: add e2e schema-upgrade test for two-phase extension migration

WentingWu666666:
## What

Adds a new e2e upgrade spec — `test/e2e/tests/upgrade/upgrade_schema_test.go` — that covers the **two-phase extension schema upgrade** (the irreversible `ALTER EXTENSION documentdb UPDATE` migration) that the existing image-only upgrade spec (`upgrade_images_test.go`) deliberately does **not** exercise.

## Why

Today we cannot gate a bump of the default DocumentDB version on schema-migration safety — there is no e2e coverage of the extension schema upgrade path. This showed up while preparing the operator 0.3.0 release: the database version bump was deferred because we had no way to prove the schema migration is safe. This test closes that gap.

## Approach

Drives the upgrade through the **user-facing knobs** — `spec.documentDBVersion` (which sets the extension + gateway images together) and `spec.schemaVersion` — rather than raw `spec.image.*` overrides, mirroring how a real user bumps a version. A new mixin `manifests/mixins/documentdb_version.yaml.template` pins the version at create time.

Flow (per the two-phase contract documented on `spec.schemaVersion`):

1. Create at the **old** version with `schemaVersion` unset (two-phase) and seed data → `status.schemaVersion` settles on the old semver.
2. Bump `spec.documentDBVersion` to the **new** version → pods roll (`status.documentDBImage` advances), but schema must **stay** at the old version (no auto `ALTER EXTENSION`), seeded data retained. Held with `Consistently` so a stray auto-migration fails.
3. Set `spec.schemaVersion` to the new version → operator runs `ALTER EXTENSION UPDATE`, `status.schemaVersion` advances, seeded data still retained.

## Runs on every e2e PR

- Lives in the existing `upgrade` matrix row, which already sets `E2E_UPGRADE=1` on PRs.
- Gated at High level (`SkipUnlessLevel(High)`), which runs at the PR default depth (Medium).
- Old/new versions **default to `0.109.0 → 0.110.0`** (public GHCR images) so the spec runs without depending on repo CI vars. Overridable via `E2E_UPGRADE_OLD_DOCUMENTDB_VERSION` / `E2E_UPGRADE_NEW_DOCUMENTDB_VERSION`.

## Status

Draft — not yet run end-to-end against a live cluster (compiles/vets clean). Opening for early review of the approach and assertions.

### DIFF
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -191,6 +191,12 @@ jobs:
       E2E_UPGRADE_CURRENT_CHART: ${{ vars.E2E_UPGRADE_CURRENT_CHART || '' }}
       E2E_UPGRADE_OLD_DOCUMENTDB_IMAGE: ${{ vars.E2E_UPGRADE_OLD_DOCUMENTDB_IMAGE || '' }}
       E2E_UPGRADE_NEW_DOCUMENTDB_IMAGE: ${{ vars.E2E_UPGRADE_NEW_DOCUMENTDB_IMAGE || '' }}
+      # Schema-upgrade spec knobs (spec.documentDBVersion + spec.schemaVersion).
+      # Set explicitly to the released pair so the two-phase migration spec
+      # runs on every e2e PR without depending on repo `vars`. Bump these
+      # when validating a newer old->new pair.
+      E2E_UPGRADE_OLD_DOCUMENTDB_VERSION: '0.109.0'
+      E2E_UPGRADE_NEW_DOCUMENTDB_VERSION: '0.110.0'
 
     steps:
       - name: Checkout repository
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -46,6 +46,16 @@ on:
         required: false
         type: boolean
         default: false
+      old_documentdb_version:
+        description: 'Schema-upgrade spec: OLD DocumentDB version override (empty = built-in default)'
+        required: false
+        type: string
+        default: ''
+      new_documentdb_version:
+        description: 'Schema-upgrade spec: NEW DocumentDB version override (empty = built-in default)'
+        required: false
+        type: string
+        default: ''
 
 permissions:
   contents: read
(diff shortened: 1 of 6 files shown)


alaye-ms (approved):

.github/workflows/test-e2e.yml
```diff
@@ -191,6 +191,12 @@ jobs:
       E2E_UPGRADE_CURRENT_CHART: ${{ vars.E2E_UPGRADE_CURRENT_CHART || '' }}
       E2E_UPGRADE_OLD_DOCUMENTDB_IMAGE: ${{ vars.E2E_UPGRADE_OLD_DOCUMENTDB_IMAGE || '' }}
       E2E_UPGRADE_NEW_DOCUMENTDB_IMAGE: ${{ vars.E2E_UPGRADE_NEW_DOCUMENTDB_IMAGE || '' }}
+      # Schema-upgrade spec knobs (spec.documentDBVersion + spec.schemaVersion).
+      # Set explicitly to the released pair so the two-phase migration spec
+      # runs on every e2e PR without depending on repo `vars`. Bump these
+      # when validating a newer old->new pair.
+      E2E_UPGRADE_OLD_DOCUMENTDB_VERSION: '0.109.0'
```
xgerman: how will we updatw this? Please eitehr automate or add as a step n the release agent.

test/e2e/tests/upgrade/helpers_test.go
```diff
@@ -29,6 +29,25 @@ const (
...
+	envOldDocumentDBVersion = "E2E_UPGRADE_OLD_DOCUMENTDB_VERSION"
+	envNewDocumentDBVersion = "E2E_UPGRADE_NEW_DOCUMENTDB_VERSION"
+
+	// Default old/new versions for the schema-upgrade spec, applied when
+	// the env vars above are unset. Chosen as the last released pair so
+	// the spec exercises a real two-phase migration on every e2e PR
+	// without depending on repo-level CI vars. Both images are published
+	// on the public GHCR documentdb/gateway repos, so documentDBVersion
+	// resolves to pullable tags in kind.
+	defaultOldDocumentDBVersion = "0.109.0"
```
xgerman, line 54: same

xgerman (changes requested):

xgerman:
Also we should file tickets for future work:
Not covered (explicit gaps in this PR)

•  spec.schemaVersion: "auto"  single-step path (documented but untested).
• Rollback within the safe window (binary downgrade before finalize).
• Negative/failure paths (migration failure, invalid/unpullable version, skipping >1 minor).
• Multi-instance/HA rollout ordering — the spec pins  INSTANCES=1 .
• Only one old→new hop; no multi-version chain.

.github/workflows/test-e2e.yml
WentingWu666666: Good call — automated it rather than leaving it as a manual step.

- The old→new pair is now a single source of truth: the `defaultOldDocumentDBVersion` / `defaultNewDocumentDBVersion` constants in `test/e2e/tests/upgrade/helpers_test.go`.
- `release_documentdb_images.yml` now bumps that pair automatically on each DocumentDB release (step 9: old ← previous default, new ← the released version), in the same auto-PR that already bumps `values.yaml` / `constants.go`. So it advances with every DB release, no manual step.
- Here in `test-e2e.yml` I removed the duplicated hardcoded `0.109.0`/`0.110.0` env and replaced it with an optional `workflow_dispatch` override (`old_documentdb_version` / `new_documentdb_version`); empty falls through to the built-in default.

test/e2e/tests/upgrade/helpers_test.go
WentingWu666666: Same fix as the workflow thread: these two constants are now the single source of truth for the pair and are auto-bumped by `release_documentdb_images.yml` on each DB release (old ← previous default, new ← released version). The duplicate hardcode in `test-e2e.yml` is gone; it only passes an optional `workflow_dispatch` override now.

WentingWu666666:
@xgerman filed #439 to track those coverage gaps as future work — `schemaVersion: "auto"` single-step, rollback within the safe window, negative/failure paths (migration failure, invalid/unpullable version, skipping >1 minor), multi-instance/HA rollout ordering, and the multi-version chain. Kept this PR scoped to the happy-path single old→new hop.

xgerman (approved):

Issue #439, opened 2026-07-31
e2e: expand schema-upgrade coverage — gaps not covered by #426


WentingWu666666:
## Context

Follow-up to #426, which adds the first two-phase extension schema-upgrade e2e spec (`test/e2e/tests/upgrade/upgrade_schema_test.go`). In review, @xgerman noted the PR intentionally covers only the happy-path single old→new hop and asked to file tickets for the explicit gaps. This issue tracks that future work.

## Not covered by #426 (explicit gaps)

- [x] **`spec.schemaVersion: "auto"` single-step path** — documented behavior but currently untested. Verify that `auto` migrates the schema in a single step (no separate finalize) and retains data. _(#443)_
- [x] **Rollback within the safe window** — binary downgrade *before* finalize. In two-phase mode the schema stays at the old version, so a `spec.documentDBVersion` downgrade before setting `schemaVersion` should be safe; assert data + schema integrity across the down-hop. _(#443)_
- [x] **Multi-instance / HA rollout ordering** — the current spec pins `INSTANCES=1`. Cover a multi-replica cluster and assert rollout ordering and data availability during the schema upgrade. Verifies replica schema convergence directly via `pg_extension.extversion`. _(#443)_
- **Negative / failure paths:**
  - [x] Invalid / unpullable `documentDBVersion` — schema does not advance, data stays readable, cluster recovers after rollback. _(#443)_
  - [x] Migration failure (e.g. `ALTER EXTENSION UPDATE` fails) — operator surfaces the error, does not corrupt data. _(deferred → #448; realistically triggered by a broken update-path, so it needs the preflight to be reachable deterministically)_
  - [x] Skipping more than one minor (jump >1 version) — expected handling. _(deferred → #448; depends on the fail-fast update-path preflight)_
- [x] **Multi-version chain** — the current spec does a single old→new hop. Cover a chained upgrade across more than two versions. _(deferred → #448)_

## Acceptance

Each gap above is either covered by an e2e spec (with appropriate labels/level) or explicitly deferred with a documented rationale. ✅ Met — see the closing comment.

Follow-up to #426. Deferred work tracked in #448.

WentingWu666666:
First two gaps addressed in #443 (draft):
- `spec.schemaVersion: "auto"` single-step path
- Rollback within the safe window (binary downgrade before finalize)

Remaining: negative/failure paths, multi-instance/HA rollout ordering, and multi-version chain.

[...]

PR #444, opened 2026-08-13
feat: fail-fast preflight for schema upgrade path (multi-minor jumps)

WentingWu666666:
## Summary

Adds an operator-side **fail-fast update-path preflight** before running `ALTER EXTENSION documentdb UPDATE`, so a schema upgrade that PostgreSQL cannot resolve is detected up front instead of failing every reconcile.

Follow-up to #426 / part of #439 (the "skip more than one minor" / multi-version-chain gap).

## Motivation

The `documentdb` extension ships minors frequently (roughly every couple of weeks), so users may bump **more than one minor** in a single change. `ALTER EXTENSION ... UPDATE` can chain intermediate `documentdb--A--B.sql` scripts, but only when the extension actually packages that continuous chain. When the chain is broken, `ALTER EXTENSION UPDATE` fails at execution time with a raw `no update path` error on **every** reconcile, with no actionable signal.

## What this does

- Before the `ALTER EXTENSION`, query `pg_extension_update_paths('documentdb')` for a non-NULL path from the installed schema to the target (same version graph `ALTER EXTENSION UPDATE` walks internally).
- If a path exists → proceed as before.
- If not → **skip** the ALTER, leave schema/data untouched, and emit a `SchemaUpgradePathMissing` Warning event (mirroring the existing `ExtensionRollback` precedent).
- A no-op (`source == target`) short-circuits to "path exists" without querying.
- The preflight only runs in single-step modes (`auto` / explicit version); two-phase mode returns early before it, so those paths are unaffected.

## Tests

- New unit test: no-path case skips ALTER, emits the Warning event, leaves the status schema version un-bumped.
- New `extensionUpdatePathExists` unit tests: `source==target` short-circuit, has-path, no-path, and SQL error propagation.
- Existing ALTER-path mocks converted from order-based to content-based branching to account for the extra preflight query.
- `go test ./internal/controller/...` green (163 specs).

## Design note / open question

The status API has no `Conditions` field today, so this surfaces the failure via a Warning event only (consistent with `ExtensionRollback`). Adding a status condition would need a CRD/deepcopy regen — flagging as a possible follow-up. See the design comment on #439; **pending maintainer sign-off** on the fail-fast contract and webhook-vs-controller placement.

## Follow-ups (not in this PR)

- e2e multi-minor **jump** test (success path + broken-chain), which needs 3+ versions / a crafted gap in CI.

Refs: #439

### DIFF
--- a/operator/src/internal/controller/documentdb_controller.go
+++ b/operator/src/internal/controller/documentdb_controller.go
@@ -769,6 +769,46 @@ func parseExtensionVersionsFromOutput(output string) (defaultVersion, installedV
 	return defaultVersion, installedVersion, true
 }
 
+// updatePathSentinelYes / updatePathSentinelNo are the tokens the preflight
+// query emits so the result can be parsed out of psql's aligned table output
+// without depending on column layout.
+const (
+	updatePathSentinelYes = "HAS_UPDATE_PATH"
+	updatePathSentinelNo  = "NO_UPDATE_PATH"
+)
+
+// extensionUpdatePathExists reports whether PostgreSQL can resolve an
+// ALTER EXTENSION update path for the documentdb extension from source to
+// target. Both versions are in pg_available_extensions format (e.g.
+// "0.109-0"). It queries pg_extension_update_paths, which exposes the same
+// version graph ALTER EXTENSION UPDATE walks internally: a row exists for a
+// resolvable (source, target) pair with a non-NULL path, and the path is NULL
+// when no chain of update scripts connects them.
+//
+// The query returns a unique sentinel token rather than the raw path so the
+// answer survives psql's aligned output formatting.
+func (r *DocumentDBReconciler) extensionUpdatePathExists(
+	ctx context.Context,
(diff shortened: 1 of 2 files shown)


WentingWu666666:
Heads-up: #450 supersedes this PR. It carries the same `pg_extension_update_paths` preflight plus the two items called out here as follow-ups — the `status.conditions` entry that this PR deferred ("would need a CRD/deepcopy regen"), and the e2e coverage for multi-minor jumps, multi-version chains, and a deterministic migration failure. It's opened against #448, which was filed for exactly those follow-ups.

Two behavioral differences from this PR, detailed in [a comment on #450](https://github.com/documentdb/documentdb-kubernetes-operator/pull/450): it **fails open** on a preflight error rather than propagating it, and it wraps the lookup in a `COALESCE` sentinel so "no rows" and "NULL path" parse identically.

Proposing we close this in favour of #450 to avoid two PRs editing `handleExtensionUpgrade`. Leaving it open for now in case reviewers would rather land this first.

PR #446, opened 2026-08-22
refactor(operator): introduce product-neutral ClusterIntent seam (non-functional)

hossain-rayhan:
### Summary
Purely non-functional refactor. It introduces an internal, product-neutral seam so the operator's reconciliation core can be reused by future operator variants without carrying product-specific branding logic in the reconciler. There is **no change to runtime behavior, the rendered CNPG `Cluster`, the CRD/API, Helm values, env vars, or image defaults.**

### Motivation
The reconciler previously read product-specific values (image names, credential secret, plugin names, spec fields) directly off the `DocumentDB` type throughout the CNPG cluster builder. This change extracts a small `internal/product` seam (`ProductProfile` + `Adapter` + `ClusterIntent`) with `DocumentDB` as the first adapter, so the shared rendering path consumes a neutral model. This is groundwork only; it does not add any new product.

### What changed
- **New `internal/product` package**
  - `ProductProfile` — product identity plus image repos, default images, credential secret, and CNPG plugin names (sourced from the existing operator constants).
  - `Adapter` interface and `DocumentDBAdapter` (first adapter).
  - `ClusterIntent` — product-neutral desired state: `Images{PostgresExtension, Gateway, Postgres, PullSecrets}`, `Topology{Instances, Affinity}`, `Storage{PvcSize}`, `Identity{Name, UID, APIVersion, Kind}`, `Postgres{UID, GID, PostInitSQL}`, `FeatureGates{IOUring}`, `Bootstrap{Recovery}`, `CredentialSecret`, plugin names, and `Product`.
  - `DocumentDBAdapter.ToClusterIntent(*DocumentDB)` resolves these via the profile.
- **`util.ResolveComponentImage`** made product-neutral: the environment-resolved version is passed in as a parameter, so the shared resolver contains no product-specific env var or default names. The `DocumentDB` wrappers keep the same behavior.
- **CNPG builder** renders from `ClusterIntent` via `GetCnpgClusterSpecFromIntent(...)`. The original `GetCnpgClusterSpec(...)` is retained as a thin backward-compatible wrapper, so existing call sites and tests are untouched.
- **Reconciler** builds the intent through the adapter and calls the intent-based entry point.
- Bootstrap helpers gain intent-based implementations with `*DocumentDB` wrappers kept for existing direct-call tests.

### Non-breaking guarantee
- Image resolution is proven identical to the previous `util.Get*` functions by drift-guard unit tests (adapter output == util output across all cases).
- Topology, storage, images, identity, process identity (UID/GID), the IOUring gate, and bootstrap (backup / persistent-volume / `PostInitSQL`) map 1:1 into the rendered `Cluster`.
- The backup reference reconstruction is lossless (`cnpgv1.LocalObjectReference` has only `Name`).
- No public API, CRD, Helm value, env var, or image-default names were renamed. Product-neutral naming is confined to the new internal seam.

### Testing
- `go build`, `go vet`, and `gofmt` are clean.
- Full non-e2e suite is green (uncached) across all packages, including the `internal/controller` envtest integration tests that render and assert the actual CNPG cluster spec.
- The `test/e2e` module does not import any changed internal symbols; it drives the operator through the Kubernetes API against the (byte-identical) rendered `Cluster`. Recommend running the e2e `smoke` label in CI as the final gate.

### Follow-ups (out of scope, not in this PR)
- Route the remaining builder inputs through the intent: PostgreSQL parameter/GUC merge, TLS (spec + status), monitoring/OTel (needs name + namespace).
- Split the Helm chart into a shared library layer plus a thin product chart (separate chart PR; verified byte-equivalent via helm-unittest).


operator/src/internal/product/documentdb.go
```diff
@@ -0,0 +1,150 @@
...
+func DocumentDBProfile() ProductProfile {
+	return ProductProfile{
+		Name:                    "DocumentDB",
+		ExtensionImageRepo:      util.DOCUMENTDB_EXTENSION_IMAGE_REPO,
+		GatewayImageRepo:        util.GATEWAY_IMAGE_REPO,
+		DefaultExtensionImage:   util.DEFAULT_DOCUMENTDB_IMAGE,
+		DefaultGatewayImage:     util.DEFAULT_GATEWAY_IMAGE,
+		DefaultCredentialSecret: util.DEFAULT_DOCUMENTDB_CREDENTIALS_SECRET,
+		SidecarInjectorPlugin:   util.DEFAULT_SIDECAR_INJECTOR_PLUGIN,
+		WALReplicaPlugin:        util.DEFAULT_WAL_REPLICA_PLUGIN,
```
alaye-ms, line 25: Getting to have all the defaults lined up nicely like this is really great

operator/src/internal/utils/util.go
```diff
@@ -421,57 +421,63 @@ func GenerateConnectionString(documentdb *dbpreview.DocumentDB, serviceIp string
...
+		return changeStreamImage
 	}
+	return defaultImage
+}
 
-	// Fall back to default
-	return DEFAULT_GATEWAY_IMAGE
+// GetGatewayImageForDocumentDB returns the gateway image for a DocumentDB instance.
+// Priority: spec.image.gateway > spec.documentDBVersion > env.DOCUMENTDB_VERSION > default
+func GetGatewayImageForDocumentDB(documentdb *dbpreview.DocumentDB) string {
```
alaye-ms: What is this function used for if we have the adapter.GatewayImage() function?

alaye-ms (approved):

operator/src/internal/utils/util.go
hossain-rayhan: This is a good catch. Will keep only the adapter path, no parallel implementation.

xgerman (approved):

PR #450, opened 2026-08-27
feat: fail-fast preflight for extension update paths

### LINKED ISSUE #448: feat: fail-fast preflight for extension update paths (+ e2e for >1-minor jumps and migration failure)
## Context

Follow-up to #439, which tracked the schema-upgrade e2e gaps left open by #426.

#443 landed the test-only gaps that needed no operator change (`schemaVersion: "auto"`, rollback within the safe window, multi-instance/HA rollout ordering, unpullable `documentDBVersion`). The remaining gaps from #439 were explicitly deferred because they all depend on one unmade decision: **the operator has no update-path validation at all today.** It runs a blanket `ALTER EXTENSION documentdb UPDATE` and relies entirely on PostgreSQL to resolve the path. Writing e2e tests for the failure modes first would just pin in whatever behavior falls out of that, so the contract needs to be ratified before the tests are meaningful.

This issue carries the design decision plus the work and tests that follow from it. Full write-up in [this comment on #439](https://github.com/documentdb/documentdb-kubernetes-operator/issues/439#issuecomment-3603237963).

## Decision needed from maintainers

1. **Confirm multi-minor jumps are a supported path, not rejected.** The `documentdb` extension ships minors roughly monthly, at times bi-weekly. A user upgrading quarterly is already 3+ minors behind; twice-a-year is 6+. Adjacent-only (N → N+1) upgrades would mean 3–12 sequential rolling restarts to catch up, which in practice pushes users to never upgrade.
2. **Confirm the fail-fast contract:** block + surface a status condition when no update path exists, rather than firing `ALTER EXTENSION` and surfacing the raw PG error.
3. **Confirm placement:** controller only, or also a static check in the validating webhook. The path check needs DB connectivity, so the controller (which already holds a SQL executor and cluster handle) is the pragmatic home; only cheap static parts could move to the webhook.

## The risk being managed

A jump only works if the extension ships a *continuous chain* of `documentdb--A--B.sql` update scripts. If any release omits its script, `ALTER EXTENSION UPDATE` fails at execution time with a raw PG error (`extension has no update path from X to Y`), re-fires every reconcile, and the user gets a cryptic failure with no actionable status.

## Proposed implementation

PostgreSQL exposes the update-path graph read-only:

```sql
SELECT path FROM pg_extension_update_paths('documentdb')
WHERE source = '0.109-0' AND target = '0.113-0';
```

- Path resolvable → non-NULL arrow-joined sequence.
- No path → `path IS NULL` → fail-fast signal.

Hook this into `documentdb_controller.go` immediately **before** the `SQLExecutor(... updateSQL)` call. If no path exists: skip `ALTER EXTENSION`, emit a `Warning` event plus a `SchemaUpgradeBlocked` status condition (reason `NoUpdatePath`) naming the exact gap, and return cleanly without a crash-loop.

| | Today | With preflight |
|---|---|---|
| When failure is learned | After firing `ALTER EXTENSION` | Before touching schema |
| Error quality | Raw PG error in logs | Actionable status condition + event |
| Visibility | Operator logs only | `status.conditions` on the CR |
| Loop behavior | Errors every reconcile | Clean stop |

## Scope

- [ ] Ratify the contract above with maintainers.
- [ ] Implement the `pg_extension_update_paths` preflight in `documentdb_controller.go`, with the `SchemaUpgradeBlocked` / `NoUpdatePath` condition and warning event.
- [ ] Unit coverage for the preflight (path present, path absent, same-version no-op).
- [ ] **e2e: skipping more than one minor (jump > 1)** — deferred from #439. Assert a resolvable jump succeeds and data survives.
- [ ] **e2e: multi-version chain** — deferred from #439. Chained upgrade across more than two versions (the #426 spec covers only a single old→new hop).
- [ ] **e2e: migration failure (`ALTER EXTENSION UPDATE` fails)** — deferred from #439. With the preflight in place this is reachable deterministically via a broken/absent update path; assert the operator surfaces the condition, does not corrupt data, and does not crash-loop.

## Acceptance

The preflight is implemented and the three deferred e2e gaps from #439 are covered with appropriate labels/levels, or re-deferred with a documented rationale.

Follow-up to #439. Original coverage work in #426 and #443.

WentingWu666666:
Closes #448. Follows up on #439 (deferred e2e gaps) and relates to #426 / #443.

## Problem

`handleExtensionUpgrade` runs a blanket `ALTER EXTENSION documentdb UPDATE` with no check that PostgreSQL can resolve an update path. If the extension image is missing a `documentdb--<from>--<to>.sql` script anywhere in the requested range, the `ALTER` fails at execution time with a raw PostgreSQL error that re-fires every reconcile. The user gets a cryptic failure, no status signal, and no indication of what to do.

## What this does

Adds a read-only preflight against `pg_extension_update_paths` immediately before the `ALTER`:

```sql
SELECT COALESCE(
  (SELECT path FROM pg_extension_update_paths('documentdb')
    WHERE source = '<installed>' AND target = '<target>'),
  'NO_UPDATE_PATH') AS update_path
```

When the absence of a path is **positively proven**, the operator:

- skips the `ALTER` entirely (no partial migration, `status.schemaVersion` does not move),
- emits a `Warning` event, and
- sets `status.conditions[SchemaUpgradeBlocked] = True` with reason `NoUpdatePath` and a message naming both versions plus the remediation,
- and returns cleanly — retrying cannot help until the spec changes, so there is no retry storm.

**It fails open on anything inconclusive.** A SQL error, unparseable psql output, or an unexpected version format all proceed with the `ALTER` exactly as before. The preflight is an advisory improvement; an inconclusive check must never wedge an upgrade that would otherwise succeed (e.g. on a PostgreSQL build lacking the function).

### Why the `COALESCE` sentinel

`pg_extension_update_paths` expresses "no path" two different ways: a `NULL` path when the pair is known-but-unreachable, and **zero rows** when either version isn't advertised at all. Those produce different psql output shapes (an empty field vs. a `(0 rows)` footer), which is fragile to parse. The `COALESCE` collapses both into a single always-present, never-empty row, so the parse is unambiguous and can't misread a gap as a valid path.

### New status conditions

`DocumentDBStatus` gains a standard `conditions` list. `SchemaUpgradeBlocked` is reconciled on **every** migration-planning path — `SchemaUpToDate`, `UpdatePathAvailable`, `NoMigrationPlanned` (rollback / two-phase mode), `NoUpdatePath` — so a stale `True` cannot survive a user correcting their spec. `meta.SetStatusCondition`'s return value suppresses redundant status writes, and the write is `RetryOnConflict`-wrapped because the informer read can lag a status update made earlier in the same reconcile.

```console
$ kubectl get documentdb my-db -o jsonpath='{.status.conditions[?(@.type=="SchemaUpgradeBlocked")]}'
```

## Test coverage

**Unit (18 new specs, run in the PR gate):** SQL shape, parser edge cases, both fail-open paths, the version-format guard, blocked → no `ALTER` fires, preflight targets the *pinned* version rather than the binary version, and both recovery transitions (revert to two-phase, retarget to a reachable version).

**e2e — closes the three gaps deferred from #439:**

| Gap | Spec | Level |
|---|---|---|
| Jump of >1 minor | `upgrade_schema_multiversion_test.go` | Lowest |
| Multi-version chain | `upgrade_schema_multiversion_test.go` | Lowest |
| `ALTER EXTENSION UPDATE` fails | `upgrade_schema_preflight_test.go` | Low |

The migration-failure spec was deferred in #439 because reproducing it needed a doctored extension image. It's now reachable with **stock images**: request a patch version that was never released (e.g. `0.109.999` while installed is `0.109.0` and the binary is `0.110.0`). That is `<=` the binary version so the validating webhook admits it, and `>` the installed version so `determineSchemaTarget` plans a migration — but no update script exists. The spec skips itself if the configured old/new versions share a major.minor, where the trick wouldn't hold.

A new `E2E_UPGRADE_DOCUMENTDB_VERSION_CHAIN` env var (documented in `test/e2e/README.md`) drives the multi-version specs, defaulting to the published tags `0.109.0,0.110.0,0.113.0,0.114.0`.

## Open: the three contract questions in #448 are still unratified

#448 asks maintainers to ratify three points, and there's been no answer. **This implementation deliberately doesn't foreclose any of them:**

1. *"Are multi-minor jumps supported?"* — treated as **yes**, which is what the code already did. The preflight adds no new restriction; it only reports what PostgreSQL itself would have done. If the answer turns out to be "no", that's a separate `spec` validation, not a change here.
2. *"Should the check live in the webhook too?"* — **controller-only**, as #448 itself recommends as the pragmatic option. The webhook can't query a live database, so it can't do this check without a round trip at admission time.
3. *"What's the recovery contract?"* — implemented as "recoverable, not terminal": correcting the spec clears the condition and the migration proceeds. The e2e spec pins that.

Happy to adjust if maintainers land somewhere different.

## Notes for reviewers

- `docs/.../api-reference.md` carries some unrelated drift (`ComponentResources`, the `postgres` image pin, `disableTLS`). That's pre-existing staleness that `make api-docs` corrected — not part of this change.
- `make lint` reports 8 `SA1019` staticcheck findings. All 8 are present on `main` unchanged; this PR adds none.
- The e2e specs are dry-run verified only; they need a kind cluster with the operator installed and `E2E_UPGRADE=1` to actually execute.

## Validation

- `make test` — all packages pass (`internal/controller` coverage 56.3% → 56.5%)
- `make manifests generate fmt vet api-docs` — no drift
- `helm unittest` — 88/88 pass
- `go vet ./...` + `gofmt` in `test/e2e` — clean
- `TEST_DEPTH=4 ginkgo --dry-run ./tests/upgrade/...` — 11/12 specs construct (1 pre-existing pending), all new specs present with correct labels


WentingWu666666:
### Relationship to #444

Flagging an overlap I found after opening this: #444 (open since 2026-08-13) already implements the core preflight against `pg_extension_update_paths`. This PR is a **superset** of it, built for #448:

| | #444 | this PR |
|---|---|---|
| `pg_extension_update_paths` preflight | ✅ | ✅ |
| Warning event on no-path | ✅ (`SchemaUpgradePathMissing`) | ✅ (`SchemaUpgradeBlocked`) |
| `status.conditions` | ❌ — explicitly deferred as "needs a CRD/deepcopy regen" | ✅ `SchemaUpgradeBlocked` reconciled on every planning path |
| e2e coverage | ❌ — listed as a follow-up (">1 minor jump", broken chain) | ✅ all three gaps from #439 |
| Unit specs | 4 | 21 |

Two deliberate behavioral differences worth a maintainer opinion:

1. **Error handling.** #444 propagates a preflight SQL error (fail closed). This PR **fails open** — an error, unparseable output, or an unexpected version string proceeds with the `ALTER` exactly as before. The preflight is advisory, so an inconclusive check shouldn't wedge an upgrade that would otherwise succeed (e.g. on a PostgreSQL build without the function).
2. **Query shape.** #444 checks for a non-NULL `path` directly; this PR wraps it in `COALESCE(..., 'NO_UPDATE_PATH')` so that "zero rows" (version not advertised) and "NULL path" (known but unreachable) collapse into one always-present row. Both mean "no path" but produce different psql output shapes, and the sentinel removes the parsing ambiguity.

The `source == target` short-circuit in #444 isn't needed here: `determineSchemaTarget` only returns a target strictly greater than the installed version, and the `defaultVersion == installedVersion` case returns earlier, so the preflight is never reached for a no-op.

**Suggestion:** land this one and close #444, rather than merging #444 first and rebasing this on top — they touch the same function and the merge would be mostly conflict resolution. Happy to do it the other way round if reviewers prefer; say the word.

WentingWu666666:
### Note on the second commit (`fix(ci):`)

The E2E matrix failed on all 18 shards at **"Setup test environment"**, before any test ran. Root cause is a pre-existing race in `.github/actions/setup-test-environment/action.yml`, not anything in this change:

```bash
if ! docker images --format "table {{.Repository}}:{{.Tag}}" | grep -q "$OPERATOR_IMAGE"; then
```

The step runs under `set -o pipefail`. `grep -q` exits the moment it matches, docker takes SIGPIPE and returns 141, and pipefail propagates that — so **finding** the image fails the check. The logs show it plainly: all eight images load successfully, the arm64 operator image is listed, and the very next line is `❌ Required operator image not found` for that same image.

Minimal repro:

```console
$ set -o pipefail
$ seq 1 10000 | grep -q 1; echo $?
141
```

Whether it trips depends on how early the match lands in docker's output and how much output is left to write, which is why this has been intermittent (there's at least one earlier failed run on `main` and one on another PR) rather than constant.

Fix: capture the list into a variable and match against that, removing the pipe. Also switched to `grep -qxF` so it's an exact literal line match instead of a substring regex — `.` in the image name was being treated as a wildcard. The `helm list | grep -q cert-manager` check a few steps later had the same shape and is fixed the same way.

I'd normally keep a CI change out of a feature PR, but this one gates the E2E suite this PR is adding specs to. Happy to split it into its own PR if reviewers prefer — it's an independent commit.

WentingWu666666:
### Note on the third commit (`ci:`)

While confirming the new e2e specs actually execute, I found they can't be reached from CI at all.

`test/e2e/levels.go` accepts five depth tiers (`Highest|High|Medium|Low|Lowest`), but `test-e2e.yml`'s `workflow_dispatch` `depth` choice only listed three:

```yaml
options:
  - Low
  - Medium
  - High
```

So a spec declared at `level:lowest` compiles, is label-selected, and is then skipped by the runtime depth gate at *every* tier a dispatch can request. The multi-version specs in this PR sit at that tier, as do any future full-sweep specs. I listed all five options; the default stays `Medium`, so PR and push runs are untouched — this only widens what a manual dispatch can ask for.

### E2E status

Green — 18/18 shards, 64/64 checks. Confirmed from the `E2E upgrade (amd64)` log that the new status condition is being written on a live cluster:

```yaml
conditions:
  - type: SchemaUpgradeBlocked
    status: "False"
    reason: SchemaUpToDate
    message: Extension schema is at 0.113.0; no migration pending.
```

One caveat worth stating plainly: the four new specs are **dry-run verified only** (`TEST_DEPTH=4 ginkgo --dry-run` — all four construct with the right labels). CI runs `TEST_DEPTH: Medium`, so they were among the 6 skipped, and I don't have Docker/kind available locally to run them for real. With the third commit in place a maintainer can now exercise them via **Actions → TEST - E2E → Run workflow → depth: `Lowest`, label: `upgrade`**. I'd suggest doing that before merge rather than taking the dry-run on trust.

The one E2E failure on the previous attempt was an unrelated CSI flake (`tar: ./pgdata/base/5: file changed as we read it` while snapshotting a live PGDATA in `cluster-replication backup`) and passed on re-run.

PR #459, opened 2026-09-15
feat: bump default DocumentDB version to 0.117.0 and unpin postgres image

WentingWu666666:
## What

- Bumps the **database-track runtime default** `documentDbVersion` from `0.113.0` -> `0.117.0` (the version users deploy).
- Reverts the temporary PostgreSQL image pin introduced in #445 from `18.4-minimal-trixie` back to the floating `18-minimal-trixie` tag.

## Why

#445 pinned CNPG's postgres image to `18.4-minimal-trixie` as a stopgap: the shipped DocumentDB `0.113.0` extension segfaulted on PG 18.6 (an unguarded `list_nth` on `fdwPrivLists` in `ExecInitModifyTable`). That was **fixed upstream** in [`documentdb/documentdb@3555c7d`](https://github.com/documentdb/documentdb/commit/3555c7d0d4997058b9d28609fd0296bda285846e), which first shipped in **`0.116.0`** (verified present at tag `v0.116-0`, absent at `v0.114-0`).

This executes the "un-pin gated on an upstream release" follow-up noted in #445: from `0.116.0` on, the insert path is ABI-correct on PG 18.6, so the default can float within PG major 18 again and automatically pick up new minors and Debian/PGDG security rebuilds. The default is set to the latest published release, `0.117.0`.

## Scope

Intentionally limited to the **runtime default users consume** plus **testing on the latest version**. The database-image build/release track (build/release workflows, gateway build Dockerfile) and cosmetic fixtures are left as-is — they are driven independently by the release automation.

## Changes

- **Runtime default (0.113.0 -> 0.117.0):** `values.yaml`, `constants.go` (`DEFAULT_DOCUMENTDB_IMAGE` / `DEFAULT_GATEWAY_IMAGE`), sidecar-injector `config.go`
- **Un-pin postgres:** `documentdb_types.go` `+kubebuilder:default` -> `ghcr.io/cloudnative-pg/postgresql:18-minimal-trixie`; comment rewritten with a `>=0.116.0` floor note + short historical note; CRDs regenerated (`make manifests generate`)
- **Literal-asserting tests:** sidecar `config_test.go`, helm `02_sidecar_injector_test.yaml`, `09_operator_deployment_test.yaml`
- **Test on latest:** e2e upgrade pair -> `0.116.0` -> `0.117.0` (`test/e2e/tests/upgrade/helpers_test.go`; both endpoints carry the PG-18.6 fix)
- **Docs:** runtime-default rows in `docs/designs/image-management.md` and `AGENTS.md`

## Validation

- `make fmt vet` + `go test ./...` in `operator/src` — pass
- `go test ./...` in `operator/cnpg-plugins/sidecar-injector` — pass
- `helm template` renders `DOCUMENTDB_VERSION: "0.117.0"` (matches updated helm unittest assertions)
- CI E2E is the real proof: floating `18-minimal-trixie` (now 18.6+) with a `>=0.116.0` upgrade pair must pass the insert path that previously segfaulted.


xgerman (approved):

operator/src/api/preview/documentdb_types.go
```diff
@@ -168,13 +168,10 @@ type ImageSpec struct {
...
-	// "18-minimal-trixie" tag, which rolled 18.4 -> 18.6 on 2026-08-13 and
-	// crashed the DocumentDB 0.113.0 extension on insert. Staying on 18.4
-	// avoids that regression while still receiving CNPG's Debian/PGDG
-	// security rebuilds; revert to the floating "18-minimal-trixie" tag once
-	// a DocumentDB release carrying the PG 18.6 fix ships.
-	// +kubebuilder:default="ghcr.io/cloudnative-pg/postgresql:18.4-minimal-trixie"
+	// The default floats within PostgreSQL major 18 via CNPG's
+	// "18-minimal-trixie" tag, so deployments pick up new 18.x minors and
+	// Debian/PGDG security rebuilds automatically.
+	// +kubebuilder:default="ghcr.io/cloudnative-pg/postgresql:18-minimal-trixie"
```
guanzhousongmicrosoft, line 180: This un-pin needs the e2e version pair bumped in the same PR, or it reintroduces the #445 crash.

`test/e2e/tests/upgrade/helpers_test.go:57-58` is still `0.109.0` / `0.110.0` — both predate the fix. Those specs do get this new default: the base manifest declares `spec.image` as a present object, so defaulting descends and stamps `18-minimal-trixie` (PG 18.6) onto the CR. `upgrade_schema_test.go:154` then inserts on the 0.109.0 extension — the segfaulting path.

CI won't catch it: those specs need `SkipUnlessLevel(High)` and `TEST_DEPTH` defaults to `Medium`, so this PR goes green regardless.

Please bump the pair to >=0.116.0, and note the >=0.116.0 floor in the comment above — nothing else records that the floating tag is only safe for fixed extensions.

(`release_documentdb_images.yml:141-145` normally advances this pair automatically; a hand-authored bump skips it.)

operator/documentdb-helm-chart/tests/02_sidecar_injector_test.yaml
```diff
@@ -82,7 +82,7 @@ tests:
           path: spec.template.spec.containers[0].env
           content:
             name: DOCUMENTDB_VERSION
-            value: "0.113.0"
+            value: "0.116.0"
```
guanzhousongmicrosoft: Hand-editing this points at an automation gap. `release_documentdb_images.yml:100-145` seds eight files but not this one or `09_operator_deployment_test.yaml:74` — both hardcode the version and assert it matches `values.yaml`.

So the next auto-generated release PR will bump `values.yaml` and fail `helm unittest`. Worth adding a sed step for these two.

docs/designs/image-management.md
```diff
@@ -374,16 +374,16 @@ When bumping database image versions, the following locations must be updated (a
...
-| `operator/src/internal/utils/constants.go` | `DEFAULT_GATEWAY_IMAGE` | `...gateway:0.113.0` |
-| `operator/cnpg-plugins/sidecar-injector/internal/config/config.go` | Default gateway image | `...gateway:0.113.0` |
-| `operator/cnpg-plugins/sidecar-injector/internal/config/config_test.go` | Expected gateway image | `...gateway:0.113.0` |
-| `operator/documentdb-helm-chart/values.yaml` | `documentDbVersion` | `"0.113.0"` |
+| `operator/src/internal/utils/constants.go` | `DEFAULT_DOCUMENTDB_IMAGE` | `...documentdb:0.116.0` |
+| `operator/src/internal/utils/constants.go` | `DEFAULT_GATEWAY_IMAGE` | `...gateway:0.116.0` |
+| `operator/cnpg-plugins/sidecar-injector/internal/config/config.go` | Default gateway image | `...gateway:0.116.0` |
+| `operator/cnpg-plugins/sidecar-injector/internal/config/config_test.go` | Expected gateway image | `...gateway:0.116.0` |
+| `operator/documentdb-helm-chart/values.yaml` | `documentDbVersion` | `"0.116.0"` |
 | `.github/workflows/test-backup-and-restore.yml` | `DOCUMENTDB_IMAGE`, `GATEWAY_IMAGE` env | `...documentdb:0.113.0` |
```
guanzhousongmicrosoft: These two rows stayed at `0.113.0` while the other nine moved, and they point at workflows that no longer exist (consolidated into `test-e2e.yml`, per `release_documentdb_images.yml:113-118`).

The table also omits three files this PR did edit: the two Helm unittest fixtures and `helpers_test.go`. As the "what to update when bumping" checklist, it currently sends you to dead files and misses live ones — plausibly how the stale e2e pair happened.

operator/documentdb-helm-chart/tests/02_sidecar_injector_test.yaml
WentingWu666666: Good catch — you're right, hand-editing these fixtures pointed straight at the sed gap. Rather than defer it, I've closed it in this PR: `release_documentdb_images.yml` now also seds `02_sidecar_injector_test.yaml` and `09_operator_deployment_test.yaml` (new step "3b", right after the `values.yaml` bump), so a future auto-release PR advances `values.yaml` and these two fixtures together and won't break `helm unittest`. The PR-body checklist was updated to match.

operator/src/api/preview/documentdb_types.go
WentingWu666666: Fixed. The e2e upgrade pair in `test/e2e/tests/upgrade/helpers_test.go` now goes `0.116.0` -> `0.117.0` — both endpoints carry the `fdwPrivLists` fix (first shipped in `0.116.0`), so the insert path no longer hits the segfault under the floating `18-minimal-trixie` (PG 18.6) default. I also added the `>=0.116.0` floor note to the `documentdb_types.go` comment (and regenerated CRDs) so the version constraint on the floating tag is recorded where the default lives.

docs/designs/image-management.md
WentingWu666666: Fixed. Dropped the two dead rows (they pointed at `test-backup-and-restore.yml` / `test-upgrade-and-rollback.yml`, now consolidated into `test-e2e.yml`) and added the three live files the PR actually edits — the two Helm unittest fixtures and `helpers_test.go`. So the "what to update when bumping" table now points only at live files.

Note on scope: per the discussion, this PR bumps only the runtime default (values.yaml / constants.go / config.go + their tests) and the e2e pair to `0.117.0`; the database-image build/release-track rows (build_wf, release_wf, gateway Dockerfile) are intentionally left at `0.113.0` since they're driven independently by the release automation.

guanzhousongmicrosoft (approved):

PR #461, opened 2026-09-16
refactor(operator): route render inputs through ClusterIntent seam

WentingWu666666:
## Summary

Routes the CNPG cluster render path through the product-neutral `ClusterIntent` seam introduced in #446. `GetCnpgClusterSpecFromIntent(intent, storageClass, isPrimaryRegion, log)` now renders the CNPG `Cluster` from a resolved, product-neutral model instead of reading `*DocumentDB` directly. **No runtime behavior change — the rendered CNPG `Cluster` is byte-identical, guarded by a drift test.**

## What changed

- **`internal/product`** — `ClusterIntent` is now a self-contained, resolved desired-state model. Groups mirror the DocumentDB CRD where the mapping is a direct passthrough:
  - `Images`, `Topology{NodeCount, InstancesPerNode, Affinity}`, `Storage{PvcSize}`, `Identity`, `Postgres`, `Resource`, `TLS{GatewaySecretName, PostgresCertificates}`, `Monitoring` (`otel.MonitoringConfig`), `LogLevel`, `Timeouts{StopDelay}`, `FeatureGates{IOUring}`, `Bootstrap`, `CredentialSecret`, `Plugins{SidecarInjectorName, WalReplicaName}`, `Product`.
  - `DocumentDBAdapter.ToClusterIntent` resolves images, credential secret, and plugin names (honoring the spec, else profile defaults), and contributes DocumentDB change-streams as a `Postgres.Parameters` entry (`wal_level=logical`) rather than leaking a product-specific feature gate into the neutral model.
- **`internal/cnpg`** — the renderer reads every input from the intent. Neutralized `MergeParameters` / `ComputeResourceSplit` via `MergeParametersResolved` / `ComputeResourceSplitFromResource` to avoid a `product`↔`cnpg` import cycle (the `*DocumentDB` functions remain thin wrappers). Params, resource split, log level, stop delay, Postgres certs, gateway TLS secret, and OTel/monitoring config all flow through the intent.
- **Reconcile-time inputs** (`storageClass`, `isPrimaryRegion`) are passed as render params, not carried on the intent, keeping the intent a pure function of the CR.

## Not in scope (owned by other reconcile paths)

`ClusterReplication`, `ExposeViaService`, `Backup`, and `SchemaVersion` are not render inputs and remain owned by their existing controllers. `NodeCount` is carried on the intent for completeness but is not yet rendered (no active reader today).

## Testing

- `cnpg_intent_drift_test.go`: byte-identical guard asserting the intent render path matches the retained `*DocumentDB` computations across a spec matrix (TLS spec+status, user params, resource envelope/overrides, IOUring, change-streams, custom logLevel/stopDelay, process identity, monitoring).
- `make test` (envtest) green across all packages; `go build` / `go vet` / `gofmt` clean. No CRD/API/Helm changes; no generated-file drift.


operator/src/internal/cnpg/cnpg_cluster.go
```diff
@@ -24,27 +24,30 @@ import (
...
+// accepted for call-site compatibility but no longer consumed: object
+// coordinates now come from the intent's Identity, and the service account is
+// not read by the renderer.
 func GetCnpgClusterSpec(req ctrl.Request, documentdb *dbpreview.DocumentDB, documentdbImage, serviceAccountName, storageClass string, isPrimaryRegion bool, log logr.Logger) *cnpgv1.Cluster {
 	intent := product.DocumentDBAdapter{}.ToClusterIntent(documentdb)
 	if documentdbImage != "" {
 		intent.Images.PostgresExtension = documentdbImage
 	}
-	return GetCnpgClusterSpecFromIntent(req, documentdb, intent, serviceAccountName, storageClass, isPrimaryRegion, log)
+	return GetCnpgClusterSpecFromIntent(intent, storageClass, isPrimaryRegion, log)
```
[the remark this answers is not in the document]
WentingWu666666: Thanks — this isn't a production issue, and I'd like to keep the current behavior. Rationale:

- `GetCnpgClusterSpec(req, …)` is now a **test-only** compatibility shim. The reconciler calls `GetCnpgClusterSpecFromIntent` directly, and it builds identity from the **fetched CR** (`documentdb.ObjectMeta`) via `intent.Identity`. In `Reconcile`, `documentdb` is loaded with `r.Get(ctx, req.NamespacedName, …)`, so its coordinates always equal `req` — production identity (and the OTel namespace hash) is correct.
- Sourcing identity from the CR itself is deliberate and *more* correct than `req`: the CNPG `Cluster` is owned by that CR, so its name/namespace/owner-refs/labels must track the CR. The pre-refactor wrapper was actually inconsistent here (ObjectMeta from `req`, but OTel/labels/owner-refs from `db.Name`); this unifies on the CR.
- Copying `req` into `Identity` would also break `TestRenderIntentSeamNoDrift`: its cases carry no `ObjectMeta`, and the expected OTel config hash is computed from `db.Namespace`, so forcing `req.Namespace` would diverge.

The only real smell is that `req`/`serviceAccountName` are now vestigial on this shim (documented as accepted-for-compatibility). Happy to drop them and migrate the ~40 test call-sites in a follow-up if preferred.

operator/src/internal/product/documentdb.go
```diff
@@ -122,28 +154,77 @@ func (a DocumentDBAdapter) ToClusterIntent(db *dbpreview.DocumentDB) ClusterInte
...
+			Namespace:  db.Namespace,
 			UID:        db.UID,
 			APIVersion: db.APIVersion,
 			Kind:       db.Kind,
 		},
-		Postgres: pg,
+		Postgres:   pg,
+		Resource:   ResourceFromSpec(db.Spec.Resource),
+		TLS:        tls,
+		Monitoring: MonitoringConfigFromSpec(db.Spec.Monitoring),
```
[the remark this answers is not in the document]
WentingWu666666: Resolved by aligning the PR description with the implementation. The "Phase 0b deferred" wording was stale from an earlier revision; the description has been updated to state that monitoring/OTel inputs are routed through the intent (`Monitoring` is a resolved `otel.MonitoringConfig`, and the renderer consumes config-map name/hash + Prometheus port from it). Mapping the full exporter config into `ClusterIntent` here is intentional and remains byte-identical — `TestRenderIntentSeamNoDrift` cross-checks the OTel config hash against a direct `otelcfg.GenerateConfigMapData(db.Name, db.Namespace, mon)` computation.

WentingWu666666:
### Re: Copilot review — addressing the suppressed comments

Thanks for the review. Summarizing how the three suppressed comments are addressed, since they don't have inline threads:

**1. `FeatureGates.ChangeStreams` "missing" (intent.go:79).** This is intentional, and the description has been corrected. Change-streams is a DocumentDB-specific gate whose only cluster effect is a PostgreSQL GUC, so the adapter contributes it as a resolved `Postgres.Parameters["wal_level"]=logical` rather than leaking a product concept into the product-neutral `FeatureGates`. Only genuinely cross-product (infrastructure) gates — currently `IOUring` — live on `FeatureGates`. The earlier description advertising a `ChangeStreams` field was stale and has been updated.

**2. Monitoring routed through the intent (controller.go).** Same root cause — the "Phase 0b deferred" note was stale. Monitoring/OTel now flows through the intent by design; the description reflects this.

**3. Drift test independence (cnpg_intent_drift_test.go).** Fair point that the *parameters* and *resource-split* comparisons delegate through the same neutral helpers on both sides. However, byte-identical-ness is established by the suite as a whole, not the drift test alone:
- The neutral helpers have independent literal-value tests: `resource_split_test.go` (e.g. `want 3Gi`, `want 13184Mi`, exact byte counts) and `pg_defaults_test.go` (e.g. `shared_buffers == "512MB"`, exact param maps).
- `cnpg_cluster_test.go` asserts the full rendered `Cluster` against literals (name/namespace, certs, storage class, images, bootstrap).
- The drift test's remaining checks — OTel config hash (via a direct `otelcfg` call), logLevel, stopDelay, Postgres certs, gateway-TLS secret — are independent of the intent path.

So the drift test is a supplementary consistency guard on top of a literal-baseline suite. The only fully-independent alternative for the circular parts would be golden fixtures captured from `main`, which would be redundant given the existing literal tests.

operator/src/internal/cnpg/cnpg_intent_drift_test.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+
+	for name, db := range cases {
+		t.Run(name, func(t *testing.T) {
+			spec := GetCnpgClusterSpec(req, db, "", "test-sa", "", true, log).Spec
+
+			// Parameters: intent path must equal the direct MergeParameters over the
+			// same memory-aware split.
+			wantMem := ComputeResourceSplit(db, DefaultSplitConfig()).PostgresMemoryBytes
+			wantParams := MergeParameters(db, wantMem)
```
guanzhousongmicrosoft, line 151: This guard can't detect drift introduced by this PR — both sides run the same code.

`MergeParameters` (`pg_defaults.go:115-118`) and `ComputeResourceSplit` (`resource_split.go:106-109`) were rewritten *in this PR* to delegate through `ToClusterIntent`, which is exactly what the renderer calls. So `wantParams` and the rendered parameters are the same expression.

I verified it: commenting out `Parameters: db.Spec.Postgres.Parameters` in `ToClusterIntent` — silently dropping every user-supplied GUC from the rendered cluster — leaves `TestRenderIntentSeamNoDrift` green. The existing `pg_defaults_test.go` catches that regression; this test does not. Same structure for the resource-split assertions at L150/L191.

The `logLevel` / `maxStopDelay` / `certificates` / `gatewayTLSSecret` assertions *are* independent and worth keeping. For the parameter and split cases, either pin concrete expected values (e.g. assert `wal_level=logical` literally in the change-streams case), or drop the `*DocumentDB` wrappers so the test is forced to state its expectations directly.

This is the PR's central correctness claim, so I'd want it resolved before merge. It also matches Copilot's suppressed comment on this line.

guanzhousongmicrosoft (commented):
Three things before merge, one of which I'd call blocking.

operator/src/internal/product/documentdb.go
```diff
@@ -122,28 +154,77 @@ func (a DocumentDBAdapter) ToClusterIntent(db *dbpreview.DocumentDB) ClusterInte
...
+		Gateway:  componentFromSpec(res.Gateway),
+		OTel:     componentFromSpec(res.OTel),
+	}
+}
+
+// MonitoringConfigFromSpec maps the DocumentDB monitoring spec onto the
+// product-neutral OTel collector config carried on the intent. It is the single
+// CRD->neutral mapping shared by the builder and the controller's ConfigMap
+// reconciliation, so generated collector config (and its hash) stays consistent.
+func MonitoringConfigFromSpec(spec *dbpreview.MonitoringSpec) otelcfg.MonitoringConfig {
```
guanzhousongmicrosoft, line 206: This mapping has no test anywhere.

The `otel/config_test.go` cases that used to cover CRD → collector config were converted to build `MonitoringConfig{...}` literals, so the downstream half kept its coverage and the CRD half lost all of it. Now untested: `spec == nil`, `Exporter == nil`, `Exporter.Prometheus != nil → Prometheus = true` (the pointer-presence → bool conversion, the easiest line here to get wrong), `OTLP.Endpoint == ""`, and the `Enabled` passthrough.

The drift test uses this same function as its own oracle, so it won't catch a mismapping either. A small table test in `internal/product` covering those five cases would close it.

operator/src/internal/cnpg/cnpg_cluster.go
```diff
@@ -24,27 +24,30 @@ import (
 
 // GetCnpgClusterSpec renders the Cluster for a DocumentDB instance. The
 // documentdbImage argument overrides the resolved extension image; pass "" to
-// use the image resolved from the instance. Retained for callers that supply an
-// explicit extension image.
+// use the image resolved from the instance. req and serviceAccountName are
+// accepted for call-site compatibility but no longer consumed: object
+// coordinates now come from the intent's Identity, and the service account is
+// not read by the renderer.
 func GetCnpgClusterSpec(req ctrl.Request, documentdb *dbpreview.DocumentDB, documentdbImage, serviceAccountName, storageClass string, isPrimaryRegion bool, log logr.Logger) *cnpgv1.Cluster {
```
guanzhousongmicrosoft, line 31: `GetCnpgClusterSpec`, `MergeParameters`, `ComputeResourceSplit` and `ProtectedParameters` all have zero non-test callers at head — the controller now calls `GetCnpgClusterSpecFromIntent` directly. The doc comment says "retained for call-site compatibility," but there are no production call sites left, and it concedes two of the six parameters are ignored (`serviceAccountName` was already dead before this PR).

Worth deleting here rather than deferring: it also resolves the drift-test issue above, since the tests would then have to state their expectations directly instead of calling back through the new path.

Same vein, both unread at head: `Topology.NodeCount` (no reader anywhere, as the commit message notes — it's also what motivated the `Instances` → `InstancesPerNode` rename), and the new `Plugins.WalReplicaName` override in `product/documentdb.go:94-97` (the only other resolution site, `physical_replication.go:90-102`, is inside a commented-out `TODO re-enable` block).

